PR #21, opened 2016-03-04
Add document parser with simple HTML parser

henryre:
Abstract base class so that people can write their own


ajratner:
Awesome!  I'll look at closely in the am, but looks good

henryre:
Whoops added an extra commit with the initial calibration plots. I'll close this and redo it

PR #25, opened 2016-03-04
Add document parser with simple HTML parser (fixed)

henryre:
Abstract base class so that people can write their own

### DIFF
--- a/DeepDiveLite.ipynb
+++ b/DeepDiveLite.ipynb
@@ -9,11 +9,20 @@
   },
   {
    "cell_type": "code",
-   "execution_count": 1,
+   "execution_count": 4,
    "metadata": {
     "collapsed": false
    },
-   "outputs": [],
+   "outputs": [
+    {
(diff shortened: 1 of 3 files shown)


henryre:
@ajratner are we making any memory usage assumptions for ddlite? If we think people might still be working with large data sets, I should switch the doc parser to a generator

ajratner:
Hey @henryre,

This all looks great!  And I even learned about a new type of soup.  I guess only two minor comments:

1. I agree with your point above, we should switch to generators.  I don't think we should over-optimize for industrial-scale use-cases or anything, as we may never want / need to get there... but this is an easy fix, in keeping with our python code elsewhere too.  (I do however think that between (i) keeping everything as a generator, (ii) supporting sparse matrices, and (iii) potentially having a db connection of some sort, we should be able to handle decent-enough sized sets... let's see!)

2. Minor: should we maybe call `NullParser` `TextParser` to be clearer?

PR #32, opened 2016-03-10
Add entity class and simple feature extractor

henryre:
QC/improvement from Mr. treedlib himself for the feature extractor is much needed


ddlite.py
```diff
@@ -227,6 +196,101 @@ def get_rule_priority_vote_accuracy(self, ground_truth, holdout_only=False):
...
+
+class Entities(Extractions):
+  def __init__(self, sents):
+    super(Entities, self).__init__(sents)
+    self.entities = self.extractions
+
+  def _apply(self, sent):
+    xt = corenlp_to_xmltree(sent)
+    # Search for consecutive named entity blocks
+    nr_pos = ['NN', 'NNS', 'NNP', 'NNPS']
```
ajratner: @henryre I think I quickly responded over email agreeing with this choice, but I actually don't think it should be hardcoded- examples have already come up where we want other POS types, or more flexibility in general

entity_features.py
```diff
@@ -0,0 +1,30 @@
...
+from templates import *
+import lxml.etree as et
+
+def compile_entity_feature_generator():
+  """
+  Given optional arguments, returns a generator function which accepts an xml root
+  and a list of indexes for a mention, and will generate relation features for this entity
+  """
+  
+  BASIC_ATTRIBS_REL = ['lemma', 'dep_label']
```
ajratner: May want more here e.g. if POS not hard-coded to be NN*?

ajratner:
Hey @henryre Looks great overall

Two main things:

1. Shouldn't the `Entities` extractor be able to extract matches over a dictionary, in the exact same way that the `Relations` one does for two dictionaries?  In general, I think the method of entity extraction should be less hard-coded.  I was originally thinking it would work in much the same way as `Relations`.  So e.g. `Entities(DictionaryMatch('G', genes))` would get dictionary matches, and maybe something slightly more customizable like `Entities(RegexMatch('pos', 'NN*'))` would implement the functionality you have in now.  Maybe the above exposes too much to the user, but we could always wrap it... thoughts?

2. Features look good for now- I have no perfect answers about what should go there... one thought is that we should make sure that the words immediately to the left and right are captured.  This may be a deficiency of treedlib... if this isn't easy to add in, we can add to this.  See the [ddlib feature list](http://deepdive.stanford.edu/gen_feats#list-of-generic-features) for inspiration / to check if anything obvious missing?

We also may want to implement a very simple 'basic' feature extractor- just word to the left / right, etc- for sanity checking?  Should be easier to do for entities than relations.  If it would be helpful for testing

henryre:
Re 1: definitely correct. I was tuning too hard for building a tagger. Will generalize a bit now for the hackathon tomorrow, then work on it again later.

Re 2: Sounds good. I checked ddlib but didn't have a great intuition for how to integrate the window features, etc. in treedlib. Thoughts here? Would be nice to do without cobbling the treedlib-based extractor together with a second one

ajratner:
Well if treedlib is too bulky right now, simple enough to make a very basic
feature generator temporarily... or I don't actually think unioning some
treedlib features with some other ones would be that bad?

henryre:
On it. We can revisit it soon too

henryre:
@ajratner updated

ajratner:
Looks great!

PR #34, opened 2016-03-11
Simplifies and semi-automates the installation with virtualenv

netj:
including Jupyter notebook :)

I think we should abstract the whole installation steps away and provide a simple launcher script for these DDlite notebooks.  We could even provide a oneliner command people can type into their terminal and just wait to get the demo DDlite notebook pop up and start playing.


henryre:
Awesome!

ajratner:
This is great!!

PR #37, opened 2016-03-11
Adds Mindtagger integration for error analysis

### LINKED ISSUE #16: Can we get a MindTagger into DDL / Jupyter notebook?
@netj Chris mentioned doing this (I think you were there?)

netj:
which downloads Mindtagger (in latest DeepDive release), then creates a
task, and launches it in a user-specific port, and displays the GUI in
an iframe.

- `rels.open_mindtagger(num_sample=20, width='100%', height=600)`
- `rels.launch_mindtagger(task_name, item_generator, ...)`
- `rels.get_mindtagger_tags()`

As long as the GUI is running in the background, users can freely
add/update the tags, and get the tags to postprocess them from the rest
of the notebook cells.

Also updates early notebook cells to skip NLP markup steps by default if
there's cached result, and show some examples to give a sense of what
the data looks like.

Known issue: Mindtagger UI is not responsive enough for small screens,
so needs some touches to fit nicely into typical notebook widths and
heights.  However, this can be done independently from DDlite.

Here's a screenshot:

Fixes #16


chrismre:
SO AWESOME!!!

henryre:
This rocks. I'll play with it in the new example

henryre:
@ajratner @netj 

Looks like the mindtagger task for evaluating Relations is set up in the Entities class. Give me a few to make it play nice with the Extractions class

henryre:
@netj Cleaned up the integration and the notebook a bit in the latest push to this branch, so everything should work smoothly (no more "it works on my machine"). I think there's an issue interfacing between ddlite's tagged indexes and the ones in MindTagger. See `Relations.get_mindtagger_items()`. For the example you gave, ddlite gives the index of "BMPR-II" with `e1_idxs` and the indices of "pulmonary hypertension" with `e2_idxs`. But when `e2_idxs` are loaded into MindTagger, they get translated to "BMPR-II" and "not". Any guess as to what's going on?

netj:
@henryre Thanks for restructuring this!  Looks so much better :)

There apparently was a bug in the MT template I was generating and made it explicit that we're laying down the indexes as JSON arrays, which now fixes the highlights as below.

henryre:
@netj Great! @ajratner I'll get `Entities` to work nicely then this looks good to go

henryre:
LGTM now. @ajratner all you. We should add utilities for analyzing precision, recall, etc. once we set up a good framework for doing that with ddlite.

ajratner:
Working (on my Mac)!  Thanks @netj!!

One very minor thing: I had issues for a second because I was loading the pickled dump of the parsed sentences from a different branch... this issue should quickly go away, but just noting in case anyone else runs across this

Merging now, excited to start using!

henryre:
I had the same issue, but was hoping it wouldn't corrupt systems worldwide based on the ... exclusivity of our user base. Easy fix is to give unique names for the sentence dumps we're doing in the examples (should have already been doing this)

ajratner:
Nbd either way.  Maintaining brand exclusivity trumps all

PR #38, opened 2016-03-11
New gene tagging example notebook

henryre:
Not completed yet, but gives us a common way to evaluate new features without the current hacky example


ajratner:
LGTM!

PR #42, opened 2016-03-13
Speed up learning with sparse operations and friends

### LINKED ISSUE #13: Extend logistic regression functions to handle scipy.csr_matrix
Support sparse matrices as optional input

henryre:
Vectorization and a few other tweaks give a 2000x speed up on a 100 observation, 4K feature problem (full gene example couldn't be run with the old methods)
Using the sparse methods gives an *additional* 50x speed up on the full gene example notebook
Sampling methods for SGD sped up similarly

Fixes #13

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,10 @@
 *.pyc
 parser/
 /stanford-corenlp-*
+
+/saved_sents.pkl
+
+/deepdive
+/deepdive-*.tar.gz
+/mindtagger.pid
+/mindtagger
--- a/DeepDiveLite.ipynb
+++ b/DeepDiveLite.ipynb
@@ -59,7 +59,7 @@
     {
      "data": {
       "text/plain": [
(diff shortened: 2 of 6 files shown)


chrismre:
WOW!

ajratner:
Awesome!

Quick comments:

- Do we need a way to insert a rule set which replaces the existing rule set, rather than extends it?
- Aside from this, if it's running feel free to merge- I'll do a more careful review of the log reg code & compare with the latest version I have in Julia later (perhaps when we add L1), but looks good for now!

ajratner:
@henryre I'd say once you've considered the minor thing I noted above, go ahead and merge!  And we can do another run over all the learning & inference code together this coming week too

henryre:
Good point. I'll make the fix and merge in the AM

ajratner:
Sounds great!

PR #47, opened 2016-03-13
Adds basic DeepDive calibration plots in matplotlib

### LINKED ISSUE #20: Output marginals + calibration plots + histograms
We need to output the basic deepdive calibration plots (notebooks are perfect for this!), as well as potentially some other histograms which guide users towards correct error analysis / debugging procedures.

We also need to output the marginals, which is a minor sub-function to add in.

henryre:
In-notebook rendering shown in the gene example. These are simple, but get the job done. We can revisit them down the road when we put in some more integrations.

Fixes #20


ajratner:
LGTM!

PR #50, opened 2016-03-17
Adds a basic RegexMatcher and wrapper for combining multiple Matchers

### LINKED ISSUE #5: Create a RegexMatch entity mention operator
This will be almost identical to the existing `DictionaryMatch` operator.

This `RegexMatch` operator, if essentially copied from `DictionaryMatch`, will be able to trivially do e.g. POS tag sequence matches (using `match_attrib=poses`); however this could be wrapped / presented as a separate operator...?

henryre:
Improvements to be made include allowing access to the unparsed sentence

Fixes #5 ... and closes out v0.1 @ajratner ???

### DIFF
--- a/ddlite.py
+++ b/ddlite.py
@@ -91,6 +91,23 @@ class RegexMatch(Matcher):
       start = bisect.bisect(start_c_idx, match.start())
       end = bisect.bisect(start_c_idx, match.end())
       yield list(range(start-1, end))
+      
+class MultiMatcher(Matcher):
+  """ 
+  Wrapper to apply multiple matchers of a given entity type 
+  Priority of labeling given by matcher order
+  """
+  def __init__(self, label, matchers):
+    self.label = label
+    self.matchers = matchers
+  def apply(self, s):
+    applied = set()
+    for m in self.matchers:
+      for rg in m.apply(s):
+        if rg[0] not in applied:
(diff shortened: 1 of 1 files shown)


ajratner:
Code looks great.  Two points: 
* the way in which overlapping matches are dealt with seems slightly opaque... in RegexMatcher we just yield all, but in MultiMatcher we yield the first that occurs for a given start index... I think in both, we should have an explicit / customizable policy ('take all matches', or 'take longest match', probably the former?)
* Why does MultiMatcher need its own label?  Might be slightly nicer if it was compositional and just took a variable number of Matcher operators as args, e.g. `MultiMatcher(RegexMatcher..., DictionaryMatcher...)`?

henryre:
Agree with 1, but we probably still want to avoid tagging something by, say, regex and dictionary.

Idea for the label was I might have a list of chem names in a dictionary and a regex for CxHyOz and want to tag all as 'CHEMICAL'. We can add a default label that composes the individual ones

ajratner:
So if we just want to avoid tagging the same span then we should check both the start and end index

And okay, got it re: label.  Maybe let's just make this an optional argument, where if it's not specified then the label of the specific extractor gets called?

henryre:
@ajratner see updates

PR #55, opened 2016-03-25
Delete old example

henryre:
Difficult to maintain a simple example for now. New `Relations` example notebook can be written later.


ajratner:
Makes sense

Issue #60, opened 2016-03-25
Error analysis workflow 1


ajratner:
1. User gets _random_ subsample of candidates in Mindtagger, and labels them
2. User gets statistics over the labeling functions, some w.r.t. to this label set (e.g. empirical acc., etc.)
3. Learn model
4. Get precision stats
5. Log (?) and repeat

Stats to show for label function development:
- Coverage
- Overlap
- Conflict
- Empirical accuracy
- Show labeling functions and/or candidates that are conflict heavy (+ low emp. accuracy lfs)

Questions:
- Should we be proscriptive, and automatically (opaquely) split their label set into a "label fn. validation set" and a test set (as default option which can be turned off)?
- How to integrate ground truth that they bring in externally?

PR #63, opened 2016-03-27
Add labeler evaluation utilities

### LINKED ISSUE #52: Output simple rule statistics (coverage, overlap, conflict, empirical rule accuracy?)

henryre:
- Adds plots for coverage (bar plot), overlap (histogram), conflict (heat plot) to assess how the labeler set as a whole performs on the candidates
- Adds tables for labelers with high conflict, high abstaining, and low empirical accuracy
- Adds utilities for dealing with ground truth and holdout sets

The notebook wasn't updated since it will be overhauled to demonstrate the new MindTagger workflow in a coming PR. There's a picture below with some toy labelers as a preview. We can demonstrate best practice re: evaluating empirical labeler accuracy on training vs. holdout sets in the notebooks. These utilities use the generality of the ground truth utilities, so the user can specify exactly how accuracy is assessed.

label_stats

Fixes #52 
Starts to f*x mega issue / @ajratner 's notepad #60


ajratner:
High level, before looking at code:

1. This looks awesome!!!  Plots are sick

2. I know how annoying this is going to sound, given our already too-long convo the other day... but "labelers" sounds odd to me... as if we're referring to different users (which isn't too off conceptually, but may be confusing).  Stick with "labeling function"?  And leave as global var in case it changes... :)?

ddlite.py
```diff
@@ -358,28 +426,134 @@ def apply_labelers(self, labelers_f, clear=False):
...
     for i,c in enumerate(self.C._candidates):    
       for j,labeler in enumerate(labelers_f):
         self.labelers[i,j + nr_old] = labeler(c)
-        
-  def learn_weights(self, nSteps=1000, sample=False, nSamples=100, mu=1e-9, 
-                    holdout=0.1, use_sparse = True, verbose=False):
+    
+  def _coverage(self):
+    cov = [(self.labelers == lab).sum(1) for lab in [1,-1]]
+    abst = self.num_labelers() - cov[0] - cov[1]
```
ajratner: The syntax here is a little confusing... for example, `self.num_labelers()` seems like it should refer to the number of labeling functions, not the nnz...

ddlite.py
```diff
@@ -358,28 +426,134 @@ def apply_labelers(self, labelers_f, clear=False):
...
+    tab.set_num(n)
+    tab.set_title("Labeler", "Mean conflicts per candidate")
+    return tab
+    
+  def _abstain_frac(self, label_idx):
+    label_csc = abs_sparse(self.labelers.tocsc()[:,label_idx])
+    return 1 - float((label_csc == 1).sum()) / self.num_candidates()
+    
+  def top_abstain_labelers(self, n=10):
+    """ Show the labelers with the highest fraction of abstains """
```
ajratner: Maybe simpler to say 'lowest coverage'?

ajratner:
- The coverage plot seems like the most potentially confusing one to me right now.  E.g. is this # of labels emitted of each type, or total # of candidates which _were given at least one of these labels_?  Should be the latter, correct?  **Perhaps we should remove the "Abstain" category, doesn't add much?**

- We could add a bit more description to the plots that print out- easy way to avoid user confusion

ajratner:
Actually, the heat map plot you have seems to basically subsume the other two plots, and is awesome... maybe we should only show that plot, + the summary stats printed out?

Either way I think looks great, but the first two plots (reading L-R) can probably be made a bit clearer at least

Otherwise is awesome.  Should wrap up without too much more work, then start getting user feedback!

ddlite.py
henryre: It does refer to the number of labeling functions (#abstain for given candidate = #labelers - #pos - #neg)

ddlite.py
henryre: Agree

henryre:
Yeah the first two plots are **labeled** poorly. I think they both have some value (your intuition for what the first one was is correct), so I'll make the **labels** more clear

henryre:
And you're killing me with the constant branding iteration...

henryre:
@ajratner Fixes above issues. We can make sure everyone knows `LF` means labeling function in the README and notebooks

ajratner:
Renaming:
- label class balance (+remove abstain)
- Label count histogram
- Label heat map

ajratner:
More precise definition for "Mean number of conflicts"

Metrics:
- number of candidates which have _some_ conflict that LF_i labels
- number of _other_ LFs which LF_i disagrees with, averaged across the candidates

PR #64, opened 2016-03-28
Add original sentence text to Sentence

### LINKED ISSUE #62: Add raw (untokenized) text as attribute to Sentence object
For regex matching, it would be very helpful to have access to the text of a single sentence without any tokenization. When tagging chemical names, for example, we frequently get these tokenization type artifacts:
 Li ( 3 ) PS ( 4 ) vs.  Li(3)PS(4) 
Some of these can be fixed with modified regexes, but it would be nice to operate on the original text itself. As far as mapping back to tokens for entity tags, we could just consider a match as anything that overlaps with the original span.

henryre:
CoreNLP output format changed to JSON to provide character offsets
Extraction notebook updated, learning notebook to follow

@jason-fries 
Fixes #62

### DIFF
--- a/GeneTaggerExample_Extraction.ipynb
+++ b/GeneTaggerExample_Extraction.ipynb
@@ -93,16 +93,16 @@
      "name": "stdout",
      "output_type": "stream",
      "text": [
-      "CPU times: user 12.4 s, sys: 216 ms, total: 12.6 s\n",
-      "Wall time: 23.1 s\n",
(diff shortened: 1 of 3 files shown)


henryre:
@ajratner can you check what I did for ordering the dependency attributes? JSON returns dependency structure without labels/parents for the corresponding token, so I sorted them by the ordering of the dependents. This seemed to work in my tests, but let me know if you see something amiss here

henryre:
Closing to fix offset-to-token issue

Issue #14, opened 2016-02-27
Write tests!


ajratner:

henryre:
Between basic inference tests and the Travis build, I'm closing this for now.

PR #67, opened 2016-03-29
Minor fixes to make master branch run once again + a Travis config to automate future tests

netj:
With a couple of questions:
- Why was treedlib submodule taken out?
- What's the best way to provide feedback on these notebooks?
- Can we set up Travis config to make sure these notebooks at least run end-to-end without Python errors, so we know when new PRs break something? (I can help, of course :)

Update: Travis config is included now, #14 is closer than we thought!


henryre:
Like F. Scott Fitzgerald with the green lights. LGTM! Thanks again @netj!!

PR #73, opened 2016-03-29
Add unparsed sentence text and document offsets to Sentence objects

### LINKED ISSUE #62: Add raw (untokenized) text as attribute to Sentence object

### LINKED ISSUE #72: Add document offsets for tokens in `Sentence`
Facilitates applying gold standard labels (among other things)
@jason-fries 

#statspadding

henryre:
* Allows better regex matching
* Allows user to apply offset-based gold standard labels
* Switches to richer JSON CoreNLP output

Fixes #62 
Fixes #72


ajratner:
LGTM

PR #81, opened 2016-03-31
Cross-validation and elastic net logistic regression

### LINKED ISSUE #23: Lasso logistic regression
DeepDive can do l1 logistic regression, so ddlite should too

henryre:
* Adds elastic net regularization for logistic regression via proximal gradient descent
* Adds cross-validation for regularization penalty
* Logreg learning function fits a sequence of penalties using warm starts for speed
  * Sequence can be user-defined or automatic
  * Automatic sequence gives a log-spaced sequence starting at a weak bound for the minimum penalty giving all zero weights and a user supplied minimum ratio
* Cross-validation uses spread-based statistic for the marginal probabilities to encourage high confidence predictions
  * User can pick the penalty with maximum spread or the max penalty within 1 se of maximum spread for final model
  * Returns weights at chosen regularization penalty learned on entire training data  
* Advanced users can see a diagnostic plot of spread and sparsity over penalties
notebook_elasticnet
* Inference tests and learning notebook updated

Still to do:
* Allow support for validation sets in CV
* Discussion of best practices in README or notebook
* Clean up parameter passing
* Figure out how to turn the diagnostic plot into a heatmap

Fixes #23 #31


ddlite.py
```diff
@@ -871,47 +888,177 @@ def transform_sample_stats(Xt, t, f, Xt_abs = None):
...
+          print "SGD converged for mu={:.3f} after {} steps".format(mu, step)
+        break
+
+      # Update weights
+      w -= rate*g
+      
+      # Apply elastic net penalty
+      soft = np.abs(w) - alpha * mu
+      #          \ell_1 penalty by soft thresholding        |  \ell_2 penalty
+      w = (np.sign(w)*np.select([soft>0], [soft], default=0)) / (1+(1-alpha)*mu)
```
ajratner: Seems like it would be ~+1 lines of code to allow users to choose L1 or L2 only?

ajratner:
- The diagnostic plot is very nice (star is next level) but we should label it better; for example I don't think "marginal spread" is standard enough that we shouldn't include the actual equation, or more detail; and the two lines should be labeled- otherwise this plot might just confuse users
- Agree with your first three to-do bullet points, if they're quick? (in general I think we need to shift from tool-building to testing for a bit- see my email)

Again though, fine by me to merge whenever convenient!

ddlite.py
henryre: alpha=0 for l2 only and alpha=1 for l1 only (documented in the new README)

henryre:
Cool, I'll label/document better and then add those as issues

PR #83, opened 2016-03-31
Add basic interface to set custom LF matrix

### LINKED ISSUE #82: Interface for access to LF matrix

henryre:
Basic checks for shape and sparse-castability

Fixes #82


ajratner:
Only superficial comments:
- `LFs` sounds like I'm accessing a list of functions `l`, when really I'm getting a matrix with rows `map(l,C)`.  I thought we were calling this something like `label_matrix` or `X`
- You have X^T w.r.t. to how we've been talking about it, just be prepared / note in the README

No showstoppers here, can merge whenever

henryre:
Agreed. Will document the transpose and rename `LFs`

PR #93, opened 2016-04-05
Fix NumPy version issue for set differences

henryre:


henryre:
@ajratner @thodrek fyi

PR #102, opened 2016-04-07
Workflow diagram

ajratner:
There's a big file in here by accident I think...

Anyway, questions:

1. Too high-level?  Detailed enough?*
2. Makes sense?

* If not, can just annotate in the README with more detailed descriptions underneath the diagram


ajratner:
@jason-fries could you take a quick look at this too if you have time?  Trying to make a rough workflow diagram for everyone tomorrow... Thanks!

ajratner:
@henryre so it seems like I added your relations commit as well by accident, was rushing at gate... Feel free to scrap this PR/branch and make a new one if easiest.  My only edit is adding in the pptx w the draft diagram...

henryre:
No problem, I'll deal with it. Couple of notation fixes to be made, but I'll handle it. I agree we should put them in the README with some annotation

Issue #112, opened 2016-04-08
Requesting Intersection Matching


ddindi1:
I'd like to request for RegexMatch to handle intersections. For my use case I would like to match a regular expressions only to nouns to limit the number of noisy matches. 

I would also like to request for the RegexMatch to match over compound phrases. For instance, if I have a noun phrase "macular degeneration" and I have a regex that identifies terms containing "gener" , I would like the candidate to be "macular degeneration" as opposed to "degeneration" given that degeneration is part of a compound phrase. 

David

PR #122, opened 2016-04-10
Matcher stuff

ajratner:
Refactors the basic candidate extractor object (now called `CandidateExtractor` for clarity) to be _compositional_.

In part this was because of requests for intersection operations, and the semantics of what an `Intersection` operator should be were slightly unclear/ambiguous, whereas compositional operators seemed better.

For example, suppose we want all phrases which (a) match a dictionary `d` and (b) contain the sub-phrase `gener` (ex from #112); now we can write:
```python
RegexFilterConcat(DictionaryMatch(label='D', dictionary=d), regex_pattern=r'.*gener')
```

Or, another request from the hackathon, all noun phrases matching a dictionary:
```python
RegexFilterAll(DictionaryMatch(label='D', dictionary=d), regex_pattern=r'NN.*')
```

@henryre when you get a chance this week could you take a look at this and see if it makes sense?  This adds a bit of complexity for the ddlite-developer but ultimately I hope it makes the ddlite-user's experience cleaner... but let's discuss

If we can agree on something here, then you can add in to the demo examples, and then we can add in the fuzzy phrase matching in this new format


ajratner:
Fixes #112.  Fixes #121.

ajratner:
@henryre note also that we have to tweak your original regex ngram matcher (which I renamed `RegexNgramMatcher` for clarity), got lazy here...

henryre:
Sounds good. Will check it out and make the demos/Travis run

henryre:
@ajratner LOVE the use of `re.compile()`

henryre:
Need to update documentation on using matchers, but the examples have some now. Compositional instead of intersection is the right design choice here. Thus, LGTM!

PR #123, opened 2016-04-10
Remove dd-biolib submodule to avoid cyclic dependency

alldefector:
... so that there won't be a cycle after https://github.com/HazyResearch/dd-biolib/pull/5

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,6 +1,4 @@
 [submodule "treedlib"]
 	path = treedlib
 	url = https://www.github.com/HazyResearch/treedlib.git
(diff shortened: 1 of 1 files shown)


ajratner:
@henryre @jason-fries I think the plan when we left off was to have dd-biolib be a submodule of ddlite (and thus remove this cyclic dependency in the other direction)?

See the [PR in dd-biolib](https://github.com/HazyResearch/dd-biolib/pull/5#issuecomment-207936990)

alldefector:
Abandoning this since ddlite will be the entry point / container of notebooks.

PR #144, opened 2016-04-29
Fix accuracy plot and F1 computation for weird np arrays

henryre:


jason-fries:
Was F1 computation wrong in some cases? Should we regenerate our current benchmarks?

henryre:
No, not the version you used. There was something weird going on with the numpy array shapes for an update I made and recall ended up being like 250 sometimes

PR #149, opened 2016-05-10
Print lf stats

ajratner:
This should be a pretty straight-forward addition, except @henryre we should make sure to sync this with the stats that are already in there, so as not to terribly confuse people...

Take a look when you get a chance, lmk what you think


henryre:
Blew this up to:
1. calculate metrics in the same place so that we don't accidentally diverge on definitions
2. calculate metrics only on the devset (might be overkill in terms of fidelity to the LF iteration process, but is on message with empirical accuracy calcs)

ddlite.py
```diff
@@ -601,35 +601,39 @@ def delete_lf(self, lf):
...
+            for lab in [1,-1]]
+
+  def coverage(self, cov=None):
+    cov = self._cover() if cov is None else cov    
+    return float(np.sum((cov[0] + cov[1]) > 0)) / len(self.devset())
 
-  def lf_stats(self):
+  def overlap(self, cov=None):    
+    cov = self._cover() if cov is None else cov    
+    return float(np.sum((cov[0] + cov[1]) > 1)) / len(self.devset())
```
ajratner: I still think my code was easier on the eyes, but fine, use ravel :)

ajratner:
We can discuss calculating on dev set vs. full set, but (a) this should at least be an optional arg, and (b) we should print this out *explicitly* with the stats, as I can see people being confused here

If you add that in quickly then LGTM!

PR #174, opened 2016-05-16
Add LF weights as default for computing marginals

henryre:


henryre:
Killing me Travis/oracle-java8-installer. Going rogue and merging in.

PR #201, opened 2016-05-26
Lots of stuff (insanity peppers?)

### LINKED ISSUE #131: Color Code Labeler Output for Judgement (-1,1)
Could we color code the results table to show the label judgement for each function? This would help keep track of which functions are labeling what when trying to balance positive and negative training examples.  Also, we should merge all the tables into one, basically something like this:

labeling

### LINKED ISSUE #133: Add relation windows

### LINKED ISSUE #157: Enforce separation of dev + test sets
This is probably a child issue to some other issues, but putting it in...

### LINKED ISSUE #176: Add new tests, README entries, comments

### LINKED ISSUE #179: Put use_LFs back in as an option + document
Perhaps also throw a warning if no separate dev + test set, or something about this...?

### LINKED ISSUE #183: First Gradient Mag number in learning alg printout bug?

### LINKED ISSUE #184: Store labels by unique candidate ID dict instead of NumPy array
If candidate extractors change, we can easily reload tags (cf TPalo)

### LINKED ISSUE #185: ValueError: Must have validation set if single mu not provided
Hi Henry, team,
I keep getting this error when I try to run the learning. What can I be doing wrong?

### LINKED ISSUE #186: In Learning example predicted probability for every candidate is always 0.5
Hi everyone,

In example GeneTaggerExample_Learning, after running the example code, I try to get the predicted probability for each candidate, but it is always 0.5 no matter what subset I choose. Is there anything I need to watch out when I use the function get_predicated_probability?

Thank you.

### LINKED ISSUE #190: Add DeepDive-style "unique" IDs to candidates

### LINKED ISSUE #194: Provide default reg. value

henryre:
Something is wrong with Travis :cry: (related to Java 8 apt-get install), @netj mind taking a look? In the meantime, I validated the examples from a clean pull and all looks well.

* Updated README
* Add separate module for learning
* Deprecate unsupervised cross-validation
* Add diagnostic plots for validation set
* Add unique IDs for candidates
* Add utilities for Relation objects
* Add separate class for managing ground truth labels (better)
* Implement dev set split for measuring accuracy generalization
* Utilities for exporting and importing ground truth labels by candidate ID
* Get rid of MindTagger vs. gold labels
* Add warnings for small dev sets/holdout sets
* Fix metrics
* Improve existing LF tables
* Add summary LF table
* Fix bias term regularization
* Add default regularization term
* Add options for using LF weights in prediction
* Add pre-pipeline integration tooling
* Add more LFs to gene tagger example

Fixes #194 
Fixes #190 
Fixes #186 
Fixes #185 
Fixes #184 
Fixes #183 
Fixes #179 
Fixes #176 
Fixes #157 
Fixes #133 
Fixes #131


.travis.yml
```diff
@@ -67,6 +66,7 @@ script:
   - python test/MatcherTests.py
   - python test/InferenceTests.py
   # Run notebooks end-to-end
+  - mkdir examples/gene_tag_example/gt
```
netj: maybe `mkdir -p`?

henryre: Probably not...that was an artifact of something else and I just took it out. The logs seem to indicate it's an apt-get issue for the JDK

henryre:
@netj  the Travis error that keeps getting hit is:
`The command "sudo -E apt-get -yq --no-install-suggests --no-install-recommends --force-yes install oracle-java8-installer" failed and exited with 100 during .`

netj:
@henryre yeah it seems like oracle.com's fault.. lame.  I tried to cache jdk/jre8 for this kind of latency/reliability reason.. but didn't find a nice way.

Also I noticed these lines from `python test/InferenceTests.py` from probably the test that runs after merging:
```
TypeError: learn_elasticnet_logreg() got an unexpected keyword argument 'maxIter'
```

henryre:
Oh thanks! I'll fix this. Maybe we can just replace Travis with @netj ! (joke)

henryre:
Hey it's working all of a sudden! :100:

henryre:
Starting to block, so merging this in while @ajratner is at his school reunion in Boston (well not IN Boston) (no, not Tufts)

ajratner:
Just sent you email about this :). Still in search of wifi... Hopefully
almost there. Will try to look over as much as I can between tonight and
tomorrow AM, but after about noon tomorrow... Don't let me block you...

ddlite.py
```diff
@@ -78,6 +90,8 @@ class candidate_internal(object):
   See entity_internal and relation_internal for examples
   """
   def __init__(self, all_idxs, labels, sent, xt):
+    self.uid = "{}::{}::{}::{}".format(sent.doc_name, sent.sent_id,
```
xiaoling: @henryre what's the difference between doc_name and doc_id? Is it possible to have multiple docs in the same file? Do you think using `doc_id` here would be safer?

henryre: @xiaoling `doc_id` vs. `doc_name` is an unfortunate distinction due to "legacy" usage. `doc_id` is the integer index in the corpus, and `doc_name` is the actual file name. The idea is that if I add more documents to the corpus, the `doc_id`s might change but I may have information attached to specific candidates based on their metadata from the old run that I'd want to retrieve in the new run

xiaoling: I see. thanks for the clarification! The doc_id field we have has a different semantics. I will change accordingly.

PR #207, opened 2016-05-31
Concat matcher

ThomasPalomares:
As discussed this morning, PR for ConcatDictionaryMatch that merge consecutive candidates from a dictionary.
Moreover,  there is a 'dictionary_optional' option for which candidates will be merged with the candidates from the main dictionary. But candidates from just this dictionary won't be extracted.
For instance, if 'decrease' and 'of' are in the optional dictionary while 'germination' is in the main one, 'decrease of germination' will be extracted as candidate while just 'of' won't be extracted.

The class is made from the syntax from ddlite_matcher.py in master but can be remade to be adapted with the version in the candidate-extractor-2 branch that @ajratner is working on if needed.

Feel free to modify anything or to just add this as a "concat" option in the original DictionaryMatch class

### DIFF
--- a/ddlite_matcher.py
+++ b/ddlite_matcher.py
@@ -121,13 +121,15 @@ class ConcatDictionaryMatch(CandidateExtractor):
   """
   Selects according to ngram-matching against a dictionary i.e. list of words.
   If candidates are juxtaposed, will merge it in one candidate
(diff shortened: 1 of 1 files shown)


ajratner:
Covered in new PR; thanks though!

PR #225, opened 2016-06-08
Added Parsers + Serialization Objects

jason-fries:
This reorganizes the parser code a bit, adding CoreNlpParser and SpacyParser classes and a generalized serialization wrapper that accepts a parser class and caches output optionally as pickle files, a text dump, or database instance. The database instance is only a stub function.


ajratner:
I'm almost done with refactoring the `DocParser` code (mostly just adding a simple XPath query based `XMLDocParser`, and some cleaning...)

After that, I'll check this out!

ajratner:
LGTM!  I'll run on my new example / write some overall tests and let you know if any problems, but will merge now

@jason-fries Can I delete the branch?

Issue #233, opened 2016-06-11
LSTM module


ajratner:
Ideally there is an object (e.g. something like `NoiseAwareLSTM`) which has a `train(candidates, marginals)` method--which takes in the output of the first pipeline stage--and then a `classify` method once trained

@SenWu thoughts here?

ajratner:
Added already by @SenWu !

PR #240, opened 2016-06-14
add LSTM module in DDLite

SenWu:
1. add LSTM module in DDLite
2. update GeneTaggerExample_Learning.ipynb to show how to run LSTM in DDLite


ajratner:
Hey @SenWu looks great!

A few things I think we should address before committing:
* **Code separation:** Just put the LSTM code in its own file at least
* **Options:** I know you had several configurations that you played around with when doing experiments- e.g. for entity extraction: (i) mention text only, (ii) words in window of size _d_ around mentions only, (iii) both; and I assume there are the same or more options for relations.  Can we expose these options here?
* **Returned marginals / predictions:** We should be clear and either return binary predictions or marginal probabilities; ideally both, which we should be able to do here.  In particular the code in this line in the example notebook is a little unclear:
`lstm_pred=[1. if i>0.5 else -1. for i in np.array(DDL.lstm_pred)[DDL.holdout()]]`
I am assuming `DDL.lstm_pred` here are the marginal probs?  Should be labeled as such, and then we can compute this line for the user and save as `lstm_pred`

ddlite.py
```diff
@@ -1297,6 +1313,355 @@ def show_log(self, idx=None):
...
+    f_grad = theano.function([x, mask, y, w], grads, name='f_grad')
+
+    lr = tensor.scalar(name='lr')
+    f_grad_shared, f_update = optimizer(lr, self.lstm_tparams, grads, x, mask, y, w, cost)
+
+    train_data, x, y, w = [], [], [], []
+    lf_probs = self.get_predicted_probability_lf()
+    for idx in self.training():
+        x.append(self.lstm_X[idx])
+        if lf_probs[idx]>0.5:
```
ajratner: Out of curiosity why this if-then?

ajratner:
@SenWu that's all from me for now!  Lmk when I should look again

ajratner:
Fixes #240 , fixes #233

ddlite.py
SenWu: Just to get likely-labels

ajratner: No I just mean is there any reason to use e.g. `y=1, w=0.65` versus `y=0, w=0.35` in your code?

SenWu:
Updated. Need sync about the tests.

ajratner:
Looks great!  Comments:
- What does word_window etc mean for relations?
- Add a few lines explaining options to the doc strings / add docstrings (we'll be transitioning to auto-generated documentation soon based on the doc strings)

For tests, can you sync with @netj or @henryre maybe?

SenWu:
1. For the meaning of word_window for relations, you can find it here: http://web.stanford.edu/~jurafsky/mintz.pdf
2. Updated doc.
3. Added a LSTM test.

ajratner:
Just add #1 to doc string for others' reference? Then LGTM!

SenWu:
Updated. Also merged up-to-date master.

henryre:
@SenWu Few things:
1. Looks like you're trying to merge into pipeline-clean, not master
2. There's a .DS_store file included
3. Storing marginals: we should probably store the latest LSTM marginals in `DDLiteModel.marginals` as well, so the utilities are agnostic to the method used. In that case, no need to store the hard class predictions.

SenWu:
@henryre Thanks for your feedback.
1. Let me close this PR, and open another one.
2. Removed. 
3. Yes, we store it in to `DDLiteModel.marginals`. Here is the link: https://github.com/HazyResearch/ddlite/blob/sen_lstm/ddlite.py#L1104

PR #266, opened 2016-06-21
Require Theano >= 0.8.2

pmlandwehr:
Testing one of the notebooks, it breaks with Theano 0.7, so should require a recent build.


ajratner:
@pmlandwehr thanks for this!

@SenWu thoughts?

SenWu:
LGTM. Merged.

PR #301, opened 2016-07-08
import defaultdict and shuffle before use

bhancock8:
I needed to explicitly import defaultdict and shuffle before I could use them in this ipynb. I added them at line 342. Is this necessary?


ajratner:
Great thanks!

PR #332, opened 2016-07-22
Update corpus object to refer to 'contexts' instead of 'sentences'

bhancock8:
All of our materials have been updated to use the new method calls (e.g., "get_contexts" instead of "get_sentences"), and a new object SentencesCorpus is a subclass of Corpus and accepts the "sentence" versions of those methods as well, for backwards compatibility.


ajratner:
@bhancock8 thanks for doing this!  LGTM

PR #338, opened 2016-07-23
ORM

stephenbach:
Not ready for a merge yet, but starting a PR to discuss.


ajratner:
High-level points to start (copying from issue):

1. Add to `tutorial/CDR_tutorial`
2. Make connection to database & table more explicit

Re: (2), I think that there should be a nice python interface for setting the db connection string (which can then be stored as `SNORKELDB` env var); this is simple and will make the mapping more clear / explicit

snorkel/parser.py
```diff
@@ -10,12 +10,133 @@
...
+
+
+class Document(SnorkelBase):
+    __tablename__ = 'document'
+    id = Column(Integer, primary_key=True)
+    corpus_id = Column(Integer, ForeignKey('corpus.id'))
+    corpus = relationship(Corpus, backref=backref('documents', uselist=True, cascade='delete,all'))
+    name = Column(String)
+    file = Column(String)
+    attribs = Column(PickleType)
```
ajratner: This `attribs` dict may contain an `xml.etree` object, which usually breaks pickle... we should check for issues here (specifically, see if this comes up when adding to `tutorial/CDR_tutorial`)

ajratner:
LGTM otherwise!

snorkel/candidates.py
```diff
@@ -1,24 +1,70 @@
...
+        """Default iterator is over Candidate objects"""
+        for candidate in self.candidates:
+            yield candidate
+
+    def get_candidates(self):
+        return self.candidates
+
+    def get_candidate(self, candidate_id):
+        """Retrieve a Candidate by id"""
+        session = SnorkelSession.object_session(self)
```
ajratner: What is the reason to get a new session each time, vs keeping one open with each Candidates object?

snorkel/candidates.py
```diff
@@ -101,80 +155,39 @@ def _extract_multiprocess(self, candidate_space, matcher, contexts, parallelism=
...
-
-        # Set basic object attributes
-        self.id          = "%s:%s-%s" % (self.sent_id, char_start, char_end)
-        self.char_start  = char_start
-        self.char_end    = char_end
-        self.char_len    = char_end - char_start + 1
-        self.word_start  = self.char_to_word_index(char_start)
-        self.word_end    = self.char_to_word_index(char_end)
-        self.n           = self.word_end - self.word_start + 1
-
```
ajratner: We'll need some of these attribs back- sent_*, metadata?

ajratner:
@stephenbach Make sure that this works with `CDR_tutorial` as that's the new code; the Gene tagger examples are just legacy supported until we can get the new tutorial done end-to-end...

ajratner:
We also have to get a handle on circular imports...

This is part of the drive to modularize properly anyway

snorkel/candidates.py
stephenbach: This shouldn't actually create a new Session, it just returns whichever session the Candidates object belongs to.

ajratner: So then getting very minor, but what happens if this is just saved as a class variable upon init?  Will they be switching sessions?

stephenbach: Good question. In that case the object's session is None. A possible fix is

```python
def get_candidate(self, candidate_id):
        session = SnorkelSession.object_session(self)
        if session is None:
            session = SnorkelSession()
            session.add(self)
            cand = session.query(Candidate).filter(Candidate.id == candidate_id).one()
            session.close()
            return cand
        else:
            return session.query(Candidate).filter(Candidate.id == candidate_id).one()
```

but this is running off the rails. See http://docs.sqlalchemy.org/en/latest/orm/session_basics.html#session-frequently-asked-questions

I think the issue is that canned queries without some externally supplied session in which to execute them is not the SQLAlchemy way AFAIK. So better fixes are probably either (1) take a session as an argument or (2) get rid of these methods and use session queries directly for specific candidates, documents, etc.

snorkel/candidates.py
stephenbach: Sentence level helper methods added in c9f9c7c

metadata has been renamed to meta because metadata is a reserved SQLAlchemy name.

stephenbach:
Ah, now I see how circular imports are messing up the CDR tutorial. Some of this is shaped by SQLAlchemy constraints, which I'm still exploring. @ajratner We should chat about this.

stephenbach:
I've collected all data models into the `snorkel.models` package so that
```python
import snorkel.models
```
or
```python
from snorkel.models import Corpus
```
works and initializes the entire database schema.

Additional details for extending the data models before initialization are included in the docs for `snorkel/models/__init__.py`

The old GeneTagger example now works without `from snorkel.snorkel import *`. Working on doing the same for the CDR tutorial, and then we should be ready for a merge.

stephenbach:
I redid the data model to make everything a Context, from Corpus to Document to Sentence. (Previously Corpora and Documents were their own types of containers.) This way, when we extract candidates, labeling functions and featurizers can traverse a general context graph.

Also used compound primary and foreign keys to avoid the id='ngram-sent-123456-0-0-5' madness. Now context and candidate primary keys are ints, and top level contexts like Corpora can have their own string primary keys in their subclass tables for easy retrieval (with their integer candidate id as a foreign key, which other contexts will use to refer to it).

The CDR tutorial is still erroring because it is trying to persist duplicate ngrams. Now that ngrams are uniquely IDed by context id, start, and end positions, this should be easier to debug.

chrismre:
So awesome.

ajratner:
Sounds great 👍 !!!  Lmk as soon as you have a clue about / example of the duplicate candidates and I can help debug

ajratner:
Preparing to merge into dev instead, closing this PR

PR #341, opened 2016-07-23
SlotFillMatch matcher

### LINKED ISSUE #311: SlotFill template-based Matcher
Can we implement a Matcher class that acccepts slot-filled strings akin to Python's  string format function?

For example:
`TemplateMatcher("{0} and/or {0} cancer", anatomy_dictionary)`
which could match strings like `breast and/or ovarian cancer`

Ideally this could operate on character level spans as well to match 
`TemplateMatcher("{0}/{0}", disease_abbrv, char_level=True)`
which could match strings like `ALD/AMN`

ajratner:
Fixes #311

### DIFF
--- a/snorkel/matchers.py
+++ b/snorkel/matchers.py
@@ -158,6 +158,38 @@ class Concat(NgramMatcher):
         return 0
 
 
+class SlotFillMatch(NgramMatcher):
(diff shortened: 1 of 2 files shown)


ajratner:
Note: this is for **single candidates that are defined compositionally**; for splitting single tokens into multiple candidates see #342

PR #344, opened 2016-07-25
Split tokens to expand Ngrams candidate space

ajratner:
Note: this only splits **single tokens** for simplicity.  I can do more than this, it just gets a bit uglier... but let me know

Another question is whether by default `Ngrams` should have `split_tokens=['-', '/']` (as it does here), or `split_tokens=None`...


ajratner:
@SenWu @jason-fries okay to merge this?

jason-fries:
Go for it

PR #361, opened 2016-07-29
Learning refactor

ajratner:
@jason-fries @SenWu We should test this the full way through, from extraction to end... help with this?  Point me to latest?

I still have to update the gene tagger example (tmrw AM), but again you can see `examples/disease_learner.jason.ipynb` for some example of how it all goes!


ajratner:
@jason-fries @SenWu So I definitely need to fix all the gene tagger / CDR examples for this, but could you:

1. Point me to / help me test this on a real example, one of yours, with the new candidates?
2. Post any comments on overall stuff?

Can also talk / work on in person tmrw, thanks!

PR #399, opened 2016-08-16
Adds support for reverse dictionaries

alekdimi:
Reverse dictionaries can be used to filter out candidates that are common English words.  All that is needed is for a reverse dict is to set the reverse keyword to True (which is by default False, so that backward compatibility is preserved).

### DIFF
--- a/snorkel/matchers.py
+++ b/snorkel/matchers.py
@@ -111,10 +111,7 @@ class DictionaryMatch(NgramMatcher):
         p = c.get_attrib_span(self.attrib)
         p = p.lower() if self.ignore_case else p
         p = self._stem(p) if self.stemmer is not None else p
-        if self.reverse:
-            return False if p in self.d else True
-        else:
-            return True if p in self.d else False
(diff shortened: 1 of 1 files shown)


snorkel/matchers.py
```diff
@@ -110,7 +111,10 @@ def _f(self, c):
         p = c.get_attrib_span(self.attrib)
         p = p.lower() if self.ignore_case else p
         p = self._stem(p) if self.stemmer is not None else p
-        return True if p in self.d else False
+        if self.reverse:
+            return False if p in self.d else True
+        else:
+            return True if p in self.d else False
```
ajratner: how about
```python
return (not self.reverse) if p in self.d else self.reverse
```
:D?

Thanks for this addition!!

alekdimi: giphy
Mind blown. Yep, that would work! Should I implement the change?

ajratner: Only if you want :)  Love the GIF

ajratner:
LGTM!

PR #415, opened 2016-08-30
Minor text fixes and initial implicit span code

lukehsiao:
- Adds [y/n] prompt to run script
- turns off automatic browser launching in run script
- add code to the Tables_Total_Recall notebook for expanding implicit ranges and suffixes.


lukehsiao:
Ah, fixing this offline then I'll open a new PR once it can be automatically merged.

PR #418, opened 2016-09-05
Suggested minor tweaks to CDR Tutorials 1 & 2

bhancock8:
Wow--that is a sexy tutorial (especially that status bar!). Nice nice work, @ajratner and @stephenbach. Most of my edits are small and optional. Not sure I saw any straight-up typos. A couple of specific notes:

1) I added the int() commands to utils to make a deprecation warning go away about how in the future you can't pass a float to a %d argument.

2) A little documentation (or even just a link) on how to clear a specific table once it's been created may be nice, since I ran into the error: "IntegrityError: (sqlite3.IntegrityError) UNIQUE constraint failed: candidate_set.name [SQL: u'INSERT INTO candidate_set (name) VALUES (?)'] [parameters: ('CDR Training Candidates',)]" quite a few times when running ce.extract if I didn't run each cell exactly once in order after creating a fresh snorkel.db file. So I always just ended up deleting snorkel.db and running Tutorial 1 again all the way through. If there isn't an easy fix for this though, probably okay to push without it, since it isn't actually broken, just inconvenient.

3) On Tutorial 2, I was getting this error the first time we try to define the Viewer, so I wasn't able to test any of those last cells that depend on the viewer object being instantiated.

---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
<ipython-input-12-cb7184511c3e> in <module>()
----> 1 from snorkel.viewer import SentenceNgramViewer
      2 
      3 # NOTE: This if-then statement is only to avoid opening the viewer during automated testing of this notebook
      4 if not AUTOMATED_TESTING:
      5     sv = SentenceNgramViewer(c[:300], session, annotator_name="Tutorial Part 2 User")

/Users/bradenhancock/snorkel/snorkel/viewer.py in <module>()
     27 
     28 
---> 29 class Viewer(widgets.DOMWidget):
     30     # TODO: Update this docstring
     31     """

/Users/bradenhancock/snorkel/snorkel/viewer.py in Viewer()
     42     the Viewer object!
     43     """
---> 44     _view_name         = Unicode('ViewerView').tag(sync=True)
     45     _view_module       = Unicode('viewer').tag(sync=True)
     46     cids               = List().tag(sync=True)

AttributeError: 'Unicode' object has no attribute 'tag'

### DIFF
--- a/snorkel/utils.py
+++ b/snorkel/utils.py
@@ -12,8 +12,8 @@ class ProgressBar(object):
 
     def bar(self, i):
         """Assumes i ranges through [0, N-1]"""
-        b = np.ceil(((i+1) / self.nf) * self.length)
-        sys.stdout.write("\r[%s%s] %d%%" % ("="*b, " "*(self.length-b), 100*((i+1) / self.nf)))
+        b = int(np.ceil(((i+1) / self.nf) * self.length))
+        sys.stdout.write("\r[%s%s] %d%%" % ("="*b, " "*(self.length-b), int(100*((i+1) / self.nf))))
         sys.stdout.flush()
 
     def close(self):
--- a/tutorial/CDR_Tutorial_1.ipynb
+++ b/tutorial/CDR_Tutorial_1.ipynb
@@ -6,7 +6,7 @@
    "source": [
     "# Chemical-Disease Relation (CDR) Tutorial\n",
     "\n",
-    "In this example, we'll be writing an application to extract *mentions of* **chemical-induced-disease relationships** from Pubmed abstracts, as per the [BioCreative CDR Challenge](http://www.biocreative.org/resources/corpora/biocreative-v-cdr-corpus/).  At core, we will be constructing a model to classify _candidate_ CDR mentions as either true or false."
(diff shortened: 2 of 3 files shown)


ajratner:
Hey @bhancock8,

Thanks so much for the helpful edits!!!

Re: (2), you can just use sqlalchemy syntax (or delete it right in the database terminal), but SB and I both agree that it would be better to have convenience / helper functions here, so we'll be adding these soon!

Re: (3) Could you check your version of jupyter notebook (`jupyter notebook --version`) and see what it is?  Mine is `4.2.1`.  I believe this problem can be resolved by upgrading jupyter with pip; could you check it out and let us know / put a note in the install section of the README if so?

Thanks!!

lanzhuzhu:
I met the same problem as (3).when I run this : sv = SentenceNgramViewer(train_cands[:300], session) in tutorial 2  and  following the tutorial 3, the same problem occured when it come to sv = SentenceNgramViewer(dev_cands, session). This means that I can't open a viewer. The error it print is : TraitError: The 'kernel' trait of a Comm instance must be a Kernel, but a value of type 'NoneType' (i.e. None) was specified.  what's more, i checked the version of jupyter notebook. Mine is  4.2.3, the newest version. I do not what's wrong with it ,could you please give me some suggestions?

ajratner:
@lanzhuzhu this PR appears to be closed but are you still running into this issue?

PR #423, opened 2016-09-10
Make ProgressBar for CorpusParser active only for max_docs option

bhancock8:
I pushed a ProgressBar for the CorpusParser a little prematurely--we won't always be able to fit all the texts in memory. But with these changes, we can not keep any in memory and still have a ProgressBar whenever they give us a max_docs value. If they run out of docs before they hit max_docs (which I don't think will happen very often, since usually you'd use that to intentionally subsample your corpus, I think), then it'll just jump to 100% when it finishes.

Also, pardon the funky capitalization of the branch name. Typed that one a little fast. :)

### DIFF
--- a/snorkel/parser.py
+++ b/snorkel/parser.py
@@ -30,20 +30,19 @@ class CorpusParser:
         corpus = Corpus(name=name)
         if session is not None:
             session.add(corpus)
-        texts = []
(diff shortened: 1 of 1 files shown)


ajratner:
LGTM!

PR #439, opened 2016-09-24
Update GridSearch and add to tutorial

### LINKED ISSUE #436: GridSearch back up and running in a tutorial
Looks like most of the code is already there. Hopefully very little tweaking is required to get it in one of the tutorials again.

henryre:
Berg gets back on the board

Fixes #436

### DIFF
--- a/snorkel/learning.py
+++ b/snorkel/learning.py
@@ -97,7 +97,7 @@ class NoiseAwareModel(object):
         """Return numpy array of elements in {-1,0,1} based on predicted marginal probabilities."""
         return np.array([1 if p > b else -1 if p < b else 0 for p in self.marginals(X)])
 
-    def score(self, X_test, L_test, gold_candidate_set, b=0.5, set_unlabeled_as_neg=True):
+    def score(self, X_test, L_test, gold_candidate_set, b=0.5, set_unlabeled_as_neg=True, disp=True):
         if L_test.shape[1] != 1:
             raise ValueError("L_test must have exactly one column.")
         predict = self.predict(X_test, b=b)
@@ -142,8 +142,9 @@ class NoiseAwareModel(object):
                     tn.add(candidate)
 
         # Print diagnostics chart and return error analysis candidate sets
-        score(test_candidates, np.asarray(test_labels), np.asarray(predict), gold_candidate_set,
-              train_marginals=train_marginals, test_marginals=test_marginals)
+        if disp:
+            score(test_candidates, np.asarray(test_labels), np.asarray(predict), gold_candidate_set,
+                  train_marginals=train_marginals, test_marginals=test_marginals)
         return tp, fp, tn, fn
 
     def save(self, session, param_set_name):
--- a/snorkel/learning_utils.py
+++ b/snorkel/learning_utils.py
@@ -193,18 +193,23 @@ class GridSearch(object):
     Runs hyperparameter grid search over a Learner object with train and test methods
     Selects based on maximizing F1 score on a supplied cross-validation (cv) set.
     """
-    def __init__(self, learner, param_names, param_val_ranges):
-        self.learner           = learner
-        self.param_names       = param_names
-        self.param_val_ranges  = param_val_ranges
+    def __init__(self, model, X, training_marginals, param_names, param_val_ranges):
+        self.model              = model
+        self.X                  = X
+        self.training_marginals = training_marginals
+        self.param_names        = param_names
+        self.param_val_ranges   = param_val_ranges
+        
+    def search_space(self):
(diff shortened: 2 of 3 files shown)


snorkel/learning.py
```diff
@@ -97,7 +97,7 @@ def predict(self, X, b=0.5):
         """Return numpy array of elements in {-1,0,1} based on predicted marginal probabilities."""
         return np.array([1 if p > b else -1 if p < b else 0 for p in self.marginals(X)])
 
-    def score(self, X_test, L_test, gold_candidate_set, b=0.5, set_unlabeled_as_neg=True):
+    def score(self, X_test, L_test, gold_candidate_set, b=0.5, set_unlabeled_as_neg=True, disp=True):
```
ajratner: `display` for readability?

snorkel/learning_utils.py
```diff
@@ -193,18 +193,23 @@ class GridSearch(object):
...
+        self.model              = model
+        self.X                  = X
+        self.training_marginals = training_marginals
+        self.param_names        = param_names
+        self.param_val_ranges   = param_val_ranges
+        
+    def search_space(self):
+        return product(*self.param_val_ranges)
+
+    def fit(self, X_test, L_test, gold_candidate_set, b=0.5, set_unlabeled_as_neg=True, **model_hyperparams):
```
ajratner: Some minor things:
* I would call e.g. `X_validation` to make it clear that they shouldn't use the test set here?
* `L_test` is a vector of labels, so should be e.g. `validation_labels` (upper case single letter reserved for matrices)

snorkel/learning_utils.py
```diff
@@ -214,24 +219,33 @@ def fit(self, candidates, gold_labels, **model_hyperparams):
...
             # Train the model
-            self.learner.train(**model_hyperparams)
+            self.model.train(self.X, self.training_marginals, **model_hyperparams)
 
             # Test the model
-            scores   = self.learner.test(candidates, gold_labels, display=False, return_vals=True)
-            p, r, f1 = scores[:3]
+            tp, fp, tn, fn = self.model.score(X_test, L_test, gold_candidate_set, b, set_unlabeled_as_neg, disp=False)
+            p, r = float(len(tp)) / (len(tp) + len(fp)), float(len(tp)) / (len(tp) + len(fn))
+            f1 = 2.0 * (p * r) / (p + r) if (p + r) > 0 else 0
```
ajratner: We should put these as standard utility fns

snorkel/learning_utils.py
```diff
@@ -214,24 +219,33 @@ def fit(self, candidates, gold_labels, **model_hyperparams):
...
-        self.learner.model.w = w_opt
+        self.model.w = w_opt
 
         # Return DataFrame of scores
         self.results = DataFrame.from_records(run_stats, columns=self.param_names + ['Prec.', 'Rec.', 'F1'])
         return self.results
-
+    
+    
+class RandomSearch(GridSearch):
```
ajratner: Minor: I was thinking `RandomSearch` would be over a continuous space of param values, rather than filtering a set of given discreet values as here; I think that would be more convenient because then the user only needs to specify a range rather than a set of values?

henryre: Shhhhhhhhh values

henryre:
Ok @ajratner changes made

snorkel/learning_utils.py
ajratner: Well played

ajratner:
@henryre looks good, added a bit of text in tutorial.  Two questions remaining:

- What's default if they don't have any dev labels?  What were you doing before (we can discuss this in person)
- Did you check to make sure adding this in preserves or improves the tutorial score (if not, we need to debug)?

henryre:
@ajratner 

- Let's add the no labels option later. Needs a discussion about best practices. Previous solution was: pick a random hold out set, then train the models and select the w s.t. sqrt(np.mean((p - 0.5)^2)) is maximized where p were the predicted marginals for the hold out set (i.e. measures how spread out the marginals are)
- 90% certain it does improve performance (from memory), but will double check shortly

henryre:
lol progressbar broke the build

stephenbach:
Might be helpful. Same issue: https://github.com/fchollet/keras/issues/2110

ajratner:
So let's just put in how to proceed without grid search (if you don't have dev set labeled)- this will preempt a lot of questions!

henryre:
Dev set F1: 0.4 -> 0.57
Test set F1: 0.7 -> 0.63

ajratner:
LGTM- will merge in after Travis passes; we should merge into `dev` and `master`

PR #462, opened 2016-10-06
Bug fixes - ProgressBar, odds ratio thresholding, etc.

bhancock8:
Credit for the ProgressBar bug discovery goes to Xiao.


bhancock8:
@ajratner As we said in the nineties, "Hit it, Rockapella!"

PR #467, opened 2016-10-14
Add experimental featurizers

henryre:
Adds context hierarchy featurization without changing default behavior


snorkel/annotations.py
```diff
@@ -37,8 +37,10 @@ def get_row_index(self, candidate):
 
     def get_key(self, j):
         """Return the AnnotationKey object corresponding to column j"""
-        return object_session(self.key_set).query(AnnotationKey)\
-                .filter(AnnotationKey.id == self.col_index[j]).one()
+        for rng, key_set in self.key_sets.iteritems():
+            if rng[0] <= j < rng[1]:
+                return object_session(key_set).query(AnnotationKey)\
+                        .filter(AnnotationKey.id == self.col_index[j]).one()
```
ajratner: So here `object_session` is just getting the database session being used by the object it's provided; but there's nothing specific to that object.  So there shouldn't be any need for this code, or for passing along the `rng` data for the key sets; the DB should handle this!

PR #480, opened 2016-11-10
Adds CID fields to Candidate object, renames get_arguments

ajratner:
Renamed `get_arguments` to `get_contexts` to avoid ambiguity


henryre:
Should we pull this in before being able to populate the CID fields?

ajratner:
@henryre yeah ready to be pulled in whenever

henryre:
LGTM!

PR #483, opened 2016-11-15
Remove accuracy sort

henryre:
Probably a CDRtifact. Won't work in general.


ajratner:
If there's one thing this great nation has shown, it's that not everyone can stomach the kind of radical change that is sometimes necessary.

This kind of a PR worries me for a whole litany of reasons.  Number one: we just don't know what's going to happen.  No way to tell.  We just don't have the science.

But, uncertainty is the spice of life.  So, let's do this.

PR #484, opened 2016-11-15
Break out scorer

henryre:
On the path to document-level scorers and optional pre-scoring operations (filtering, etc.)

### DIFF
--- a/snorkel/learning/utils.py
+++ b/snorkel/learning/utils.py
@@ -82,9 +82,9 @@ class MentionScorer(Scorer):
 
 
 def print_scores(ntp, nfp, ntn, nfn, title='Scores'):
-    prec = ntp / float(ntp + nfp)
-    rec  = ntp / float(ntp + nfn)
-    f1   = (2 * prec * rec) / (prec + rec)
+    prec    = ntp / float(ntp + nfp) if ntp + nfp > 0 else 0.0
+    rec     = ntp / float(ntp + nfn) if ntp + nfn > 0 else 0.0
+    f1      = (2 * prec * rec) / (prec + rec) if prec + rec > 0 else 0.0
     print "========================================"
     print title
     print "========================================"
--- a/snorkel/learning/utils.py
+++ b/snorkel/learning/utils.py
@@ -98,102 +98,10 @@ def print_scores(ntp, nfp, ntn, nfn, title='Scores'):
     print "========================================\n"
 
 
-#TODO: update doc string
-def score(test_candidates, test_labels, test_pred, gold_candidate_set, train_marginals=None, test_marginals=None):
-    '''
-    Compute score with true recall
-    
-    :param candidates       candidate set
-    :param candidates_gold  candidate set gold (true candidate)
-    :param gold             true set gold
-    :param pred             model predictions 
-    
-    '''
-    # false negatives from complete gold set (missing from candidate set)
-    gold_fn = [c for c in gold_candidate_set if c not in test_candidates]
-
-    # Print calibration plots
-    if train_marginals is not None and test_marginals is not None:
-        print "Calibration plot:"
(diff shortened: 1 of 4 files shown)


snorkel/learning/utils.py
```diff
@@ -1,14 +1,86 @@
...
 warnings.filterwarnings("ignore", module="matplotlib")
 
 
+class Scorer(object):
+    """Abstract type for scorers"""
+    def __init__(self, test_candidates, test_labels, gold_candidate_set):
+        """
+        test_candidates: candidates corresponding to test_labels
+        test_labels: ground truth labels for test candidates
+        gold_candidate_set: full set of gold labeled candidates
```
ajratner: I think this has been a point of confusion for people, so let's specify the exact types here?  So:

- List of `Candidate`s
- `csrLabelMatrix`
- `CandidateSet`

Right?

snorkel/learning/utils.py
```diff
@@ -1,14 +1,86 @@
...
+                    if test_label == 1:
+                        tp.add(candidate)
+                    else:
+                        fp.add(candidate)
+                else:
+                    if test_label == -1:
+                        tn.add(candidate)
+                    else:
+                        fn.add(candidate)
+            except KeyError:
```
ajratner: I'm not sure we should catch this error... seems like inviting disaster in a scoring fn?

snorkel/learning/utils.py
```diff
@@ -1,14 +1,86 @@
...
+        fn = set()
+
+        for i, candidate in enumerate(self.test_candidates):
+            try:
+                test_label_index = self.test_labels.get_row_index(candidate)
+                test_label   = self.test_labels[test_label_index, 0]
+
+                # Set unlabeled examples to -1 by default
+                if test_label == 0 and set_unlabeled_as_neg:
+                    test_label = -1
```
ajratner: Looking at the code below, this `set_unlabeled_as_neg` flag isn't actually respected- it's always True.

I think we should actually just check to make sure all the test labels are in {-1,1}, else throw an error?

snorkel/learning/utils.py
```diff
@@ -1,14 +1,86 @@
...
+            except KeyError:
+                test_label_array.append(-1)
+                if test_marginals[i] > b:
+                    fp.add(candidate)
+                else:
+                    tn.add(candidate)
+
+        # Print diagnostics chart and return error analysis candidate sets
+        if display:
+            predict = marginals_to_labels(test_marginals, b)
```
ajratner: If we're going to use this `marginals_to_labels` fn, we should use consistently i.e. above too

snorkel/learning/utils.py
```diff
@@ -1,14 +1,86 @@
...
+        gold_candidate_set: full set of gold labeled candidates
+        """
+        self.test_candidates = test_candidates
+        self.test_labels     = test_labels
+        self.gold_cs         = gold_candidate_set
+
+    def score(self, test_marginals, train_marginals=None, b=0.5, set_unlabeled_as_neg=True, display=True):
+        raise NotImplementedError()
+
+
```
ajratner: I think we should also add the equivalent `RelationScorer` or whatever you had in mind, with `NotImplementedError()`, just so the intended design doesn't get forgotten if we forget to finish this?

snorkel/learning/utils.py
henryre: I'll respond verbally in a few seconds.

henryre:
Agree with all of these points. `MentionScorer` was just a copy-paste from the old `score` method. We're going to redo scoring soon, so might be better to just pull this in now with the appropriate bug fixes

snorkel/learning/utils.py
henryre: But yes.

ajratner:
Working on Emily's stuff / basic testing, merging in

PR #487, opened 2016-11-15
Standing PR for multicontext featurization

henryre:
Still needs some testing, will assign when ready


henryre:
@ajratner ok ready to rock

snorkel/features/context_features.py
```diff
@@ -0,0 +1,40 @@
...
+
+
+def get_sentence_token_count_feats_base(candidate, attr, ngram, stopwords):
+    sentence = candidate.get_parent().parent
+    return get_token_count_feats(candidate, sentence, attr, ngram, stopwords)
+
+
+def get_document_token_count_feats(stopwords=None, ngram=1, attr='lemmas'):
+    return partial(get_document_token_count_feats_base, attr=attr, ngram=ngram,
+                   stopwords=stopwords)
```
ajratner: What's this partial fn used for?  Necessary?

ajratner: /these partial fns.  More of them :)

snorkel/features/relative_features.py
```diff
@@ -0,0 +1,148 @@
...
+            # Apply TDL features
+            for f in get_tdl_feats(xmltree.root, s1_idxs, s2_idxs,
+                                   stopwords=stopwords):
+                yield 'TDL_' + f, 1
+    else:
+        raise NotImplementedError("Only handles unary or binary candidates")
+
+
+def get_span_feats_stopwords(stopwords):
+    return partial(get_span_feats, stopwords=stopwords)
```
ajratner: We should probably load in some stopwords by default?

henryre: This actually hurt performance on CDR. My stopword set may have been too aggressive. But we're obvs going to make that pass on TDL anyways sooooo

henryre:
@ajratner Doc strings are checked in

ajratner:
@henryre awesome will look over in a little bit

snorkel/models/candidate.py
```diff
@@ -80,6 +80,10 @@ def get_contexts(self):
         """Get a tuple of the consituent contexts making up this candidate"""
         return tuple(getattr(self, name) for name in self.__argnames__)
 
+    def get_parent(self):
```
ajratner: What I think we ideally want to do here (not now, but down the road) is return the lowest shared context in the context hierarchy.  For now though, perhaps we should either (a) explicitly check to make sure they have same parent, else throw error, or (b) explicitly name this `get_first_parent`?

snorkel/features/relative_features.py
```diff
@@ -0,0 +1,190 @@
...
+
+def get_entity_word_idxs(sentence, et, cid):
+    """Get indices of @sentence tokens with type @et and id @cid"""
+    itr = enumerate(zip(sentence.entity_types, sentence.entity_cids))
+    return [i for i, (t, c) in itr if c == cid and t == et]
+
+
+def get_first_document_span_feats(candidate, stopwords=None):
+    """Base function for sentence dependency path features where the sentence
+       is the first sentence in the parent @Document of @candidate which
```
ajratner: This seems a bit arbitrary- worth generalizing at all?

snorkel/models/candidate.py
henryre: Let's go with (a). The derived methods should error out if this doesn't hold.

snorkel/features/relative_features.py
henryre: Agree it should be generalized, but is practically useful enough to not be arbitrary. Let's generalize if we hear of a related use case. Can't think of one right now.

henryre:
How can there be conflicts with files that don't exist in the base branch...

PR #490, opened 2016-11-16
Split filenames by last '.' instead of first

aliskin:
Running HTMLDocParser on files doc.{abc}.html causes sqlite errors.


henryre:
Looks good, thanks for the change!

PR #495, opened 2016-11-20
Cascade delete

ajratner:
One of the major user annoyances / points of friction currently is difficulty in deleting tables.  Part of this can be solved by simplifying the schema a bit (e.g., getting rid of many-to-many sets!), but mainly just three important things, which this PR aims to address:

1. `cascade_delete_set`: Easy function for deleting many-to-many sets, and cascading the delete all the way through, e.g. `Corpus -> Documents -> Sentences -> Spans -> Candidates -> Labels`;
2. Set up cascades in data model properly to make deleting easier / cascade fully through
3. `AnnotatorLabel`: A new class to **preserve labels created by human annotators** in a stable format, a.k.a. the one thing we shouldn't delete, so they are not lost when deletes cascade.

These are all implemented, but I still need to finish an easy function to reload `Labels` given just the stable `AnnotatorLabel` table.  The only annoying point here is reconstructing missing `Candidates` and constituent `Contexts`, given just the `AnnotatorLabel` table... probably just need to redefine the stable_id to be recursively constructed.  Will aim to finish this later tonight

### DIFF
--- a/snorkel/db_helpers.py
+++ b/snorkel/db_helpers.py
@@ -1,4 +1,4 @@
-from .models import Corpus, Document, Sentence
+from .models import Corpus, CandidateSet, AnnotationKeySet, ParameterSet
 from sqlalchemy.orm import object_session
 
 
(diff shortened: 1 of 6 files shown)


ajratner:
For now, `reload_annotator_labels` just reloads all `Labels` that correspond to an existing `Candidate`, marking labels that are in `AnnotatorLabels` but were skipped due to the corresponding `Candidate` not existing.

Ideally we improve on this, but seems like it gets a bit messy.  Rest of the PR open for comments

PR #514, opened 2017-01-09
Autoclear

stephenbach:
Starting PR. Goals are:
1. Simpler interface for UDFs. Users should not need to work with UDFRunner. Instead it should be wrapped for specific tasks by subclasses such as CorpusParser.
2. UDFs clear the relevant tables before execution by default, making correct usage easy.


snorkel/parser.py
```diff
@@ -7,45 +7,79 @@
...
+                                           parse_tree=parse_tree,
+                                           fn=fn,
+                                           in_queue=in_queue)
+
+    def clear(self, session, **kwargs):
+        session.query(Context).delete()
+        # We cannot cascade up from child contexts to parent Candidates, so we delete all Candidates too
+        session.query(Candidate).delete()
+
+    class CorpusParserUDF(UDF):
```
ajratner: Does this class need to be defined inside the other class?

snorkel/udf.py
```diff
@@ -1,77 +1,46 @@
...
+            session.commit()
+            session.close()
 
         # Execute the UDF
         if parallelism is None or parallelism < 2:
             self.apply_st(xs, progress_bar=progress_bar, **kwargs)
         else:
             self.apply_mt(xs, parallelism, **kwargs)
 
+    def clear(self, session, **kwargs):
```
ajratner: Makes sense to put this here, circumvents the annoyances that I was facing putting in `UDF`, nice

snorkel/parser.py
```diff
@@ -7,45 +7,79 @@
...
-from .models import Document, Sentence, construct_stable_id
-from .udf import UDF
-from .utils import ProgressBar, sort_X_on_Y
+from .models import Candidate, Context, Document, Sentence, construct_stable_id
+from .udf import UDF, UDFRunner
+from .utils import sort_X_on_Y
 
 
-class DocParser:
+class CorpusParser(UDFRunner):
```
ajratner: Is this implicitly making it easier to e.g. split sets by Document, rather than by Sentence?  Fine with that just clarifying

snorkel/candidates.py
```diff
@@ -31,72 +28,86 @@ class CandidateExtractor(UDF):
...
+                        continue
+                    elif not self.symmetric_relations and ai > bi:
+                        continue
+
+                # Assemble candidate arguments
+                for i, arg_name in enumerate(self.candidate_class.__argnames__):
+                    candidate_args[arg_name + '_id'] = args[i][1].id
+
+                # Checking for existence
+                if not clear:
```
ajratner: We should automatically set `clear=True` if we know we're deleting all the candidates before running, right?

snorkel/candidates.py
```diff
@@ -31,72 +28,86 @@ class CandidateExtractor(UDF):
...
+                                                 candidate_class=candidate_class,
+                                                 cspaces=cspaces,
+                                                 matchers=matchers,
+                                                 self_relations=self_relations,
+                                                 nested_relations=nested_relations,
+                                                 symmetric_relations=symmetric_relations,
+                                                 in_queue=in_queue)
+
+    def clear(self, session, **kwargs):
+        session.query(Candidate).filter(Candidate.split == kwargs['split']).delete()
```
ajratner: Nice

snorkel/parser.py
stephenbach: It's not changing any functionality, just a bit of renaming. CorpusParser is actually more like the old SentenceParser, since that's the actual UDF we're distributing.

snorkel/candidates.py
stephenbach: Set by parent UDFRunner in 4ad529b.

snorkel/parser.py
stephenbach: No, just an organizational thing.

ajratner:
Reviewing + checking parallelism on raiders now!

PR #519, opened 2017-01-14
Minor usability / aesthetic edits

ajratner:
* Add alias `apply_existing` (so we have `apply` and `apply_existing`, vs. `fit` and `transform`... thoughts?)
* Nest `load_matrix` subclasses under `UDFRunner` subclasses for convenience


ajratner:
Should be ready to merge in, just someone review / any comments on the naming of alias?

stephenbach:
Looks good, merging.

PR #521, opened 2017-01-17
Fix to Annotator.reduce

ajratner:


ajratner:
Also working on multi-thread on pg, merging in!

PR #535, opened 2017-01-18
Better training set rebalancing

### LINKED ISSUE #533: Add better class rebalancing to LogisticRegression

henryre:
See `LabelBalancer`

cc: @jason-fries 

Fixes #533

### DIFF
--- a/snorkel/contrib/learning/relstm.py
+++ b/snorkel/contrib/learning/relstm.py
@@ -148,6 +148,8 @@ class reLSTM(TFNoiseAwareModel):
             @dim: embedding dimension
             @batch_size: batch size for mini-batch SGD
             @rebalance: bool or fraction of positive examples desired
+                        If True, defaults to standard 0.5 class balance.
+                        If False, no class balancing.
             @dropout_rate: rate for tensorflow.nn.dropout(...)
             @max_sentence_length: maximum sentence length for candidates
             @print_freq: number of epochs after which to print status
--- a/snorkel/contrib/learning/relstm.py
+++ b/snorkel/contrib/learning/relstm.py
@@ -2,7 +2,7 @@ import cPickle
 import numpy as np
 import tensorflow as tf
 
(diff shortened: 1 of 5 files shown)


ajratner:
@henryre I thought there was a TF class for doing this?  Did you check out?

henryre:
That's batch normalization. Will do this for deepier models. This is much more basic.

ajratner:
K will review in a bit

snorkel/contrib/learning/relstm.py
```diff
@@ -147,7 +147,7 @@ def train(self, candidates, training_marginals, n_epochs=25, lr=0.01,
             @lr: learning rate
             @dim: embedding dimension
             @batch_size: batch size for mini-batch SGD
-            @rebalance: rebalance training examples?
+            @rebalance: bool or fraction of positive examples desired
```
ajratner: Just make this a tiny bit more clear- e.g. state that 0.5 would represent standard class balancing, `False` means no class balancing, what is `True`?

snorkel/learning/utils.py
```diff
@@ -11,6 +11,51 @@
...
+class LabelBalancer(object):
+    def __init__(self, y):
+        """Utility class to rebalance training labels
+        For example, to get the indices of a training set
+        with labels y and around 90 percent negative examples,
+            LabelBalancer(y).get_train_idxs(rebalance=0.1)
+        """
+        self.y = np.ravel(y)
+    
+    def _get_pos(self, split):
```
ajratner: `split` is a weird name for this

snorkel/contrib/learning/relstm.py
henryre: Updated comment.

snorkel/learning/utils.py
henryre: Disagree, and this is an internal method so YOLO.

Issue #368, opened 2016-08-01
CoreNLP char_offsets/token mappings do not match raw document text


varun-tandon:
I am having an issue with the DictionaryMatch where it matches incorrect words. 

I construct the DictionaryMatch object like so: 
`DiseaseMatch = DictionaryMatch( d = ["CRPC"], ignore_case= True)`
And I create Entities like so:
` ngrams = Ngrams(n_max=1) `
 ` E = Entities(sentences, CandidateExtractor(ngrams, DiseaseMatch))`

When I print out these Entities I notice the DictionaryMatch object **extracts words that are not part of the Dictionary,** in this case, it extracts the word "with", despite "CRPC" being the only word in the dictionary:
```
<Entity: [u'with'][9]>
<Entity: [u'CRPC'][30]>
<Entity: [u'CRPC'][39]>
<Entity: [u'CRPC'][13]>
<Entity: [u'CRPC'][7]>
<Entity: [u'CRPC'][10]>
<Entity: [u'CRPC'][15]>
<Entity: [u'CRPC'][25]>
```
The sentence it is extracting "with" from is:

```
 We conclude that plasma AGR2 may be used along with CGA , and other serum markers , to help define NECRPC .
```

Here are some other examples:

```
Dictionary= ["CSPC"]

Incorrect Entities: 
<Entity: [u'2500'][18]>

Sentence: Blood was collected from 44 patients with metastatic prostate cancer separated
as: castration sensitive prostate cancer (CSPC, n ¼ 5); castration resistant prostate cancer
(CRPC, n ¼ 36); and neuroendocrine-predominate CRPC defined by PSA  1 ng/ml in the
presence of wide-spread metastatic disease (NE-CRPC, n ¼ 3).
```
```
Dictionary=["PSA"]

Incorrect Entities: 
1. <Entity: [u'was'][19]>
2. <Entity: [u'and'][13]>

Sentences: 
1.A third pattern with stable AGR2 levels despite decreasing PSA (Fig. 3D , F , and G) was also observed . 
2. In the five subjects with CSPC, the median (range) age and serum PSA were 58.1 (51.1–66.6) years and 428.6 (2.4–7,606) ng/ml, respectively.
```
```
Dictionary=["CGA"]

Incorrect Entities:
<Entity: [u'the'][24]>

Sentence:
The median (range) of CGA levels in normal subjects was 70 (22–714) ng/ml, which is consistent with the assay’s normal value of <100 ng/ml.
```
[...]

PR #550, opened 2017-01-25
Fix CoreNLP Divergence Bug

bhancock8:
A previous issue (#368) identified that at times the raw text and the CoreNLP parse for a sentence diverged, and we had a check in place to go with the CoreNLP parse when this happened. The cause of this bug was not accounting for the fact that Penn Tree Bank tokens (e.g., ')') were being converted from 1 character long to 5 characters long (e.g., '-RRB-'), with offsets treating them as 1 and summing of word lengths counting them as 5. This fix replaces the PTB characters immediately, before character counting begins, so the extra divergence check is no longer required and our code can be simplified.

### DIFF
--- a/snorkel/parser.py
+++ b/snorkel/parser.py
@@ -271,6 +271,7 @@ class CoreNLPHandler(object):
                 dep_lab.append(deps['dep'])
                 dep_order.append(deps['dependent'])
 
+            parts['text'] = ''.join(t['originalText'] + t['after'] for t in block['tokens'])
             # make char_offsets relative to start of sentence
             abs_sent_offset = parts['char_offsets'][0]
             parts['char_offsets'] = [p - abs_sent_offset for p in parts['char_offsets']]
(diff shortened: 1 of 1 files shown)


snorkel/parser.py
```diff
@@ -223,7 +223,10 @@ def __init__(self, tok_whitespace=False, split_newline=False, parse_tree=False):
                         backoff_factor=0.1,
                         status_forcelist=[ 500, 502, 503, 504 ])
         self.requests_session.mount('http://', HTTPAdapter(max_retries=retries))
-    
+        
+        # Use this for recongizing PTB tokens
+        self.ptb_rgx = re.compile(r'-[A-Z]{2}B-')
```
ajratner: I'm slightly concerned that with weird text (like stuff with Chemicals, etc.) this regex might not be specific enough.  We can just list out the PTB tokens in the `PTB` dict and use this?

snorkel/parser.py
```diff
@@ -233,6 +236,11 @@ def _kill_pserver(self):
 
     def parse(self, document, text):
         """Parse a raw document as a string into a list of sentences"""
+        def ptb_clean(text):
```
ajratner: Does this function need to be inside the other function?  Make separate?

bhancock8:
Found a much sexier solution--just do the conversion inline with the dictionary 'get' operator. No extra function or regex required.

ajratner:
Great!

PR #558, opened 2017-02-06
Fix SGD for small batches

henryre:
Running a `SparseLogisticRegression` batch with an all-zero sparse matrix results in an error during training. This can happen for the last batch in an epoch since it may have very few rows.

### DIFF
--- a/snorkel/learning/logistic_regression.py
+++ b/snorkel/learning/logistic_regression.py
@@ -48,7 +48,7 @@ class LogisticRegression(TFNoiseAwareModel):
             raise Exception("Sparse input matrix. Use SparseLogisticRegression")
         return X
 
-    def _run_batch(self, X_train, y_train, i, r):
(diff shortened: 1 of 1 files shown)


henryre:
travis has given the green light, but will @ajratner?

snorkel/learning/logistic_regression.py
```diff
@@ -48,7 +48,7 @@ def _check_input(self, X):
             raise Exception("Sparse input matrix. Use SparseLogisticRegression")
         return X
 
-    def _run_batch(self, X_train, y_train, i, r):
+    def _run_batch(self, X_train, y_train, i, r, *args):
```
ajratner: ...???

henryre: SparseLR inherits, and its `_run_batch` uses another arg

PR #560, opened 2017-02-06
Small parser patch

henryre:
IDK why this was broken or what this is actually fixing.

### DIFF
--- a/snorkel/parser.py
+++ b/snorkel/parser.py
@@ -263,7 +263,7 @@ class CoreNLPHandler(object):
                 dep_lab.append(deps['dep'])
                 dep_order.append(deps['dependent'])
 
-            parts['text'] = ''.join(t['originalText'] + t['after'] for t in block['tokens'])
+            parts['text'] = ''.join(t['originalText'] + t.get('after', '') for t in block['tokens'])
             # make char_offsets relative to start of sentence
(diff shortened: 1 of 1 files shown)


henryre:
Actually this looks like it's stemming from a merge issue in
* https://github.com/HazyResearch/snorkel/commit/10f23e6ca0742ba0ed05e6393db94bd515e4778d
* https://github.com/HazyResearch/snorkel/commit/35e714902e04c58118ce2445fbaef3ad7d138af4

@bhancock8 can you check this out and make the appropriate changes and/or rollback `parser.py` to `ab7b23`

bhancock8:
Yep--your fix is exactly what was needed, and it makes sense. Looks like the `get` function being dropped was just a transcription error. The original line looked like this:

`parts['text'] = ''.join(t['originalText'] + t.get('after', '') if i < L - 1 else t['originalText'] for i,t in enumerate(block['tokens']))`

And after we worked out the bug in CoreNLP being caused by PTB characters, that extra "if" statement wasn't needed. This patch looks good to me.

henryre:
Ok cool, can you run a diff on the rest of the file to make sure everything else was preserved?

bhancock8:
Yep--diff looks good. That's everything.

PR #568, opened 2017-02-17
Precision-recall curve for NoiseAwareModel

### LINKED ISSUE #466: Output precision-recall curves with score function

henryre:
Decided to put it in a function separate from score so that people won't accidentally threshold-hack

prcurve

Fixes #466

### DIFF
--- a/snorkel/learning/utils.py
+++ b/snorkel/learning/utils.py
@@ -74,6 +74,51 @@ class Scorer(object):
 
 class MentionScorer(Scorer):
     """Scorer for mention level assessment"""
+
+    def score_b(self, test_marginals, b, get_candidates=False, set_unlabeled_as_neg=True, set_at_thresh_as_neg=True):
+        test_label_array = []
+        tp, fp, tn, fn = set(), set(), set(), set()
+        it = self.test_candidates if get_candidates else self.test_labels
+        for i, x in enumerate(it):
+            idx = self.test_labels.get_row_index(x) if get_candidates else i
+            label = self.test_labels[idx, 0]
+
+            # Set unlabeled examples to -1 by default
+            if label == 0 and set_unlabeled_as_neg:
+                label = -1
+          
+            # Bucket the candidates for error analysis
+            test_label_array.append(label)
+            if label != 0:
+                if test_marginals[i] > b:
+                    if label == 1:
+                        tp.add(x if get_candidates else i)
+                    else:
+                        fp.add(x if get_candidates else i)
+                elif test_marginals[i] < b or set_at_thresh_as_neg:
+                    if label == -1:
+                        tn.add(x if get_candidates else i)
+                    else:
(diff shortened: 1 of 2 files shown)


snorkel/learning/utils.py
```diff
@@ -74,6 +74,51 @@ def score(self, test_marginals, train_marginals=None, b=0.5, set_unlabeled_as_ne
 
 class MentionScorer(Scorer):
     """Scorer for mention level assessment"""
+
+    def score_b(self, test_marginals, b, get_candidates=False, set_unlabeled_as_neg=True, set_at_thresh_as_neg=True):
```
ajratner: I don't really get why we need to separate into `score` and `score_b`?  Maybe it's just the name in particular that bugs me.  Maybe `score_b` -> `_score_buckets`?

ajratner: I don't really get why we need to separate into `score` and `score_b`?  Maybe it's just the name in particular that bugs me.  Maybe `score_b` -> `_score_buckets`?

snorkel/learning/utils.py
```diff
@@ -74,6 +74,51 @@ def score(self, test_marginals, train_marginals=None, b=0.5, set_unlabeled_as_ne
...
+                        fn.add(x if get_candidates else i)
+
+        return (tp, fp, tn, fn, test_label_array)
+
+    def pr_curve(self, test_marginals, set_unlabeled_as_neg=True, set_at_thresh_as_neg=True):
+        """Display a precision-recall curve for various threshold values"""
+        bxr = np.linspace(0, 1.0, 21)
+        cts = [self.score_b(test_marginals, b, False, set_unlabeled_as_neg, set_at_thresh_as_neg)[:4] for b in bxr]
+        p, r, f1 = zip(*[scores_from_counts(*x) for x in cts])
+        with plt.style.context('fivethirtyeight'):
```
ajratner: srsly?  That's so pre-election

ajratner:
Also this is super nice looking and awesome

snorkel/learning/utils.py
henryre: Modularity. Reuse of code. Etc. This is called both by `pr_curve` (over multiple threshold values) and by `score`. Can rename to `_score_threshold`

snorkel/learning/utils.py
henryre: They gave it a 30% chance...

henryre:
Agree with integrating with parameter searches. Will wrap it in

ajratner:
@henryre Should we just merge this in without integrating with grid search, and do that later?

ajratner:
@henryre thoughts on what to do with this?

henryre:
@ajratner sorry just saw. Definitely still a worthwhile addition, but going to close because we're too out of date.

PR #579, opened 2017-03-11
Parser: really remove HTML comments.

rychev:
The current regex never matches anything. 

(Also see the comments to the first answer on http://stackoverflow.com/questions/1936466/beautifulsoup-grab-visible-webpage-text )

### DIFF
--- a/snorkel/parser.py
+++ b/snorkel/parser.py
@@ -1,13 +1,13 @@
 # -*- coding: utf-8 -*-
 import atexit
 from bs4 import BeautifulSoup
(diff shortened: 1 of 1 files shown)


henryre:
LGTM, thanks for the fix!

henryre:
@rychev Can you add the original regex back in as an `elif`? Might match on someone else's data. Ping me when you do, and I'll pull in the changes.

ajratner:
Closing for now, feel free to reopen!

PR #591, opened 2017-04-16
Added 2 Preprocessors: CSV input paths processor, Apache Tika processor

thammegowda:
+ Added `CSVPathsPreprocessor`:  this one helps to read the documents referenced in a CSV file. It seems similar to TSVPreprocessor, but the difference is - in TSV processor text content itself is stored, whereas in my case, texts are stored in separate files and its paths are referenced in CSV file. 

+ Added `TikaPreprocessor`: this one uses apache tika to retrieve text content from PDF and other complex file types.

### DIFF
--- a/python-package-requirement.txt
+++ b/python-package-requirement.txt
@@ -13,3 +13,4 @@ requests
 scipy>=0.18
 sqlalchemy>=1.0.14
 tensorflow>=1.0
+tika
--- a/snorkel/parser.py
+++ b/snorkel/parser.py
@@ -123,6 +123,49 @@ class TextDocPreprocessor(DocPreprocessor):
             yield Document(name=name, stable_id=stable_id, meta={'file_name' : file_name}), f.read()
 
 
+class CSVPathsPreprocessor(DocPreprocessor):
+    """This `DocumentPreprocessor` treats inputs file as index of paths to
+     actual documents; each line in the input file contains a path to a document.
+
+     **Defaults and Customization:**
+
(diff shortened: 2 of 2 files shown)


thammegowda:
If you are curious how we are using these:

we have our corpus spread out as many individual .txt and  .ann files, Example: https://github.com/wkiri/MTE-corpus/tree/master/lpsc14-text

The .txt files are generated from .pdf s by Apache tika, it will be nice to streamline this process.

Some thing like:
```bash
!find pdf_dir -name *.pdf > input.csv # list of files
```

```python
from snorkel.parser import TikaPreprocessor, CSVPathsPreprocessor, CorpusParser
CorpusParser().apply(CSVPathsPreprocessor('input.csv', parser_factory=TikaPreprocessor))
````

chrismre:
Awesome!!!

ajratner:
This is awesome!!!  I'll review today!

ajratner:
Hey @thammegowda this is awesome, and LGTM!  I'll go ahead and merge in once checks have passed.

Two things that came up while @henryre and I were looking over just now:

(1) At some point soon, we should add some material to the tutorials (either a new tutorial or to existing tutorial) showcasing TIKA + Snorkel usage!  Added issue so we don't forget this: #593

(2) Minor, but we realized that we should probably make separate classes for (i) loading files from user's system (e.g. your `CSVPathsPreprocessor`) and (ii) parsing the data (e.g. with TIKA). Posted an issue so we don't forget, let us know (in comments there) if you have any thoughts here: #592

chrismattmann:
BOOM

PR #603, opened 2017-04-27
Tweak generative model defaults.

stephenbach:
The defaults are unnecessarily high. Turning them down.


ajratner:
http://www.rentistoodamnhigh.org/index.html

PR #633, opened 2017-05-24
Parallelize candidate extraction

lukehsiao:


snorkel/contrib/fonduer/fonduer/candidates.py
```diff
@@ -1,21 +1,173 @@
...
+        Here, we define a context as a Phrase.
+        :param context:
+        :param clear:
+        :param split: Which split to use.
+        """
+        # Generate TemporaryContexts that are children of the context using the candidate_space and filtered
+        # by the Matcher
+        for i in range(self.arity):
+            self.child_context_sets[i].clear()
+            for tc in self.matchers[i].apply(self.candidate_spaces[i].apply(self.session, context)):
```
lukehsiao: Besides moving this code to being under Fonduer contrib, this is the only real change.

```py
for tc in self.matchers[i].apply(self.candidate_spaces[i].apply(self.session, context)):
```
Now passes in `self.session`, where it didn't before.

tutorials/fonduer/hardware/hardware_spaces.py
```diff
@@ -245,3 +245,34 @@ def apply(self, context):
                     meta           = None)
             else:
                 yield ts
+
+
+class OmniNgramsVolt(OmniNgrams):
```
lukehsiao: This is just here for the ce_v_max tests. We can remove it again.

Issue #631, opened 2017-05-24
Get correct LF accuracies


ajratner:
When deps are present

ajratner:
Merged into dev

PR #637, opened 2017-05-26
Supervised inputs to gen model

### LINKED ISSUE #627: Add in priors + fixed weights for LFs
Internal representation passed to generative model will now be, in addition to labeling function matrix `L` (M x N):
- `priors`: An M-dim vector specifying the prior _probabilities_ of all the LFs
- `is_fixed`: An M-dim sparse vector specifying if the LF weight is fixed (e.g. for hand-labeled data LF) or not

If `prior` is not == 0.5, then a fixed factor will be added to re-center the weight for that LF (which will still be regularized to zero, otherwise annoying NS changes needed)

This will cover (a) passing labeled data, and (b) setting individual priors of the LFs, e.g. based on the dev set accuracies (see separate issue!)

Front end input will be something like:
```
gen_model.train(L_train, LF_priors=[0.5,...,0.5], labeled_data=None, labeled_prior=1.5)
```

TODO list:
- [x] Implementation of new fixed factors in Snorkel
- [x] Test of the above
- [x] Implementation in front end
- [x] Remove redundant `y`
- [x] Change LF initial acc to at prior val
- [x] Implement proper calculation of LF accs. with deps
- [x] Write a few additional sanity-check level tests

ajratner:
Fixes #627


ajratner:
Fails tests after merging into dev... (acc of the "fixer" LF way off).  Will look at in AM

snorkel/learning/gen_learning.py
```diff
@@ -184,16 +188,23 @@ def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, lf_cl
...
     optional_names = ('lf_prior', 'lf_propensity', 'lf_class_propensity')
-    dep_names = ('dep_similar', 'dep_fixing', 'dep_reinforcing', 'dep_exclusive')
-
-    def train(self, L, y=None, deps=(), init_acc = 1.0, init_deps=1.0, init_class_prior=-1.0, epochs=10, step_size=None, decay=0.99, reg_param=0.1, reg_type=2, verbose=False,
-              truncation=10, burn_in=5, timer=None):
+    dep_names = (
+        'dep_similar', 'dep_fixing', 'dep_reinforcing', 'dep_exclusive'
+    )
+
+    def train(self, L, y=None, deps=(), LF_priors=None, labels=None,
```
ajratner: @stephenbach is `y` used anywhere?  If not, can just use this instead of `labels`

snorkel/learning/gen_learning.py
```diff
@@ -377,10 +412,20 @@ def _compile(self, L, y, init_acc, init_deps, init_class_prior):
...
+                weight[w_off]['isFixed'] = True
+                weight[w_off]['initialValue'] = np.float64(0.5 * np.log(LF_priors[i] / (1 - LF_priors[i])))
+                w_off += 1
+            # Learnable acc for LF
+            if (not is_fixed[i]):
+                weight[w_off]['isFixed'] = False
+                if LF_priors[i] == 0.5:
+                    weight[w_off]['initialValue'] = np.float64(init_acc + .1 - .2 * self.rng.random())
+                else:
+                    weight[w_off]['initialValue'] = np.float64(0)
```
ajratner: @bryan-he Should we just have all LF accs initialize at the prior value, rather than having two separate ways of handling?  Thoughts here?

snorkel/learning/gen_learning.py
stephenbach: `y` is a vector of true labels that will be treated as ground truth, making the learning fully supervised.

ajratner: So... does that obviate this PR?  Or is the idea that we wanted to have finer-grained control over the prior on the ground truth labels?  Either way we should resolve to only one way of handling input ground truth labels...

stephenbach: The latter. `y` was mostly for testing. My understanding is that @thodrek needs strong priors to handle "ground truth" but doesn't want to actually set it to a deterministic constraint, due to numeric issues.

ajratner: Okay, makes sense- just confirming.  So can we get rid of / replace `y`?  Any preference here?

(Also: I'll tag you as reviewer again once it's ready for review, some small bug w one of the tests still)

snorkel/learning/gen_learning.py
bryan-he: If all of the priors are 0.5, we have to give them a weight > 0, otherwise we might converge to the mirror image of the correct set of weights.

bryan-he:
This pull request now also has the diagnostic information via inference.

It turns out that LF propensity is enough of a dependency to mess with the accuracies that were found, so we needed this to get the checks for the test working.

bryan-he:
Resolves #631

ajratner:
Have to finish fixing gen learning test then will ping for review...

ajratner:
@stephenbach This should pass tests now and is then ready for review!

snorkel/learning/gen_learning.py
```diff
@@ -390,8 +507,9 @@ def _compile(self, L, y, init_acc, init_deps, init_class_prior):
         # Compiles variable matrix
         #
         for i in range(m):
-            variable[i]['isEvidence'] = False if (y is None or i not in y) else True
-            variable[i]['initialValue'] = self.rng.randrange(0, 2) if (y is None or i not in y) else 1 if y[i] == 1 else 0
+            variable[i]['isEvidence'] = False
+            # TODO: Change this to range (0, cardinality)!
```
stephenbach: What is the significance of this TODO?

snorkel/learning/gen_learning.py
```diff
@@ -184,55 +188,153 @@ def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, lf_cl
...
-        :param deps: collection of dependencies to include in the model, each element is a tuple of the form
-                     (LF 1 index, LF 2 index, dependency type), see snorkel.learning.constants
-        :param init_acc: initial weight for accuracy dependencies (in log scale)
-        :param init_deps: initial weight for additional dependencies, except class prior (in log scale)
-        :param init_class_prior: initial class prior (in log scale), note only used if class_prior=True in constructor
+        :param deps: collection of dependencies to include in the model, each 
+                     element is a tuple of the form 
+                     (LF 1 index, LF 2 index, dependency type),
+                     see snorkel.learning.constants
+        :param LF_priors: Priors on the LF accuracies
```
stephenbach: What is the expected type of this input?

snorkel/learning/gen_learning.py
```diff
@@ -184,55 +188,153 @@ def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, lf_cl
...
-
-    def train(self, L, y=None, deps=(), init_acc = 1.0, init_deps=1.0, init_class_prior=-1.0, epochs=10, step_size=None, decay=0.99, reg_param=0.1, reg_type=2, verbose=False,
-              truncation=10, burn_in=5, timer=None):
+    dep_names = (
+        'dep_similar', 'dep_fixing', 'dep_reinforcing', 'dep_exclusive'
+    )
+
+    def train(self, L, deps=(), LF_priors=None, LF_prior_default=0.7, 
+        labels=None, label_prior=0.95, init_deps=1.0, 
+        init_class_prior=-1.0, epochs=100, step_size=None, decay=0.99,
```
stephenbach: Is 100 epochs necessary for good performance? I find that closer to 10 is sufficient.

snorkel/learning/gen_learning.py
```diff
@@ -184,55 +188,153 @@ def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, lf_cl
...
-                     (LF 1 index, LF 2 index, dependency type), see snorkel.learning.constants
-        :param init_acc: initial weight for accuracy dependencies (in log scale)
-        :param init_deps: initial weight for additional dependencies, except class prior (in log scale)
-        :param init_class_prior: initial class prior (in log scale), note only used if class_prior=True in constructor
+        :param deps: collection of dependencies to include in the model, each 
+                     element is a tuple of the form 
+                     (LF 1 index, LF 2 index, dependency type),
+                     see snorkel.learning.constants
+        :param LF_priors: Priors on the LF accuracies
+        :param LF_prior_default: Default prior for the LFs
```
stephenbach: I don't understand what a default prior is, what happens if I don't set it, and how to specify non-default priors (whatever those are).

snorkel/learning/gen_learning.py
```diff
@@ -184,55 +188,153 @@ def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, lf_cl
...
+        :param labels: Optional ground truth labels
+        :param label_prior: Prior on the optional ground truth labels
```
stephenbach: What is a label prior? How does it differ from a class prior?

snorkel/learning/gen_learning.py
ajratner: I believe one issue (that @bryan-he pointed out) is that the step size will still be quite large after only 10 epochs.  This could be addressed by dynamically setting the step size decay factor... but either way thinking (from tests even) was that 10 is too small?  @bryan-he any thoughts?

ajratner:
Main point to discuss (can do in person @stephenbach @bryan-he ): Do we want to use priors?  Adds more (fixed) factors to graph, but seems like should be more correct.  @bryan-he and I felt was the way to go but can discuss further!

PR #646, opened 2017-05-30
Categorical

ajratner:
Adds:
* Support for categorical variables to `GenerativeModel` and `(Sparse)LogisticRegression` end models
* Support for categorical `Candidate` sub-classes with verbose value names
* Saving of training / predicted marginals in sparse format
* Scoring utils, etc. upgraded for categorical support
* New toy demo notebook for categorical setting

Note, does *not* yet add:
* Support for non-binary labeling in the `Viewer`: Since we are moving to using the `Viewer` for viewing data, and external tools (e.g. BRAT) for labeling anyway
* Support for categorical to all end models in `snorkel.contrib`: This is a simple change to make whenever we do a cleaning pass on the `contrib` models


ajratner:
Reminder: Clear Travis caches so that NS/dev is installed!  Everything but test_categorical should pass now...

snorkel/annotations.py
```diff
@@ -336,33 +375,88 @@ def load_matrix(self, session, split, key_group=0, **kwargs):
...
-def _to_annotation_generator(fns):
-    """"
-    Generic method which takes a set of functions, and returns a generator that yields
-    function.__name__, function result pairs.
-    """
-    def fn_gen(c):
-        for f in fns:
-            yield f.__name__, f(c)
-    return fn_gen
+def save_marginals(session, X, marginals, training=True):
```
stephenbach: If we don't actually like the term 'marginal,' is this the time to switch to something else?

snorkel/learning/gen_learning.py
```diff
@@ -535,42 +506,55 @@ def _compile(self, L, init_deps, init_class_prior, LF_acc_priors, is_fixed):
...
+                                     "Valid values are 0 to %d. " % (i, j, data, cardinality))
 
         #
         # Compiles factor and ftv matrices
         #
-
         # Class prior
         if self.class_prior:
+            if cardinality != 2:
+                raise NotImplementedError("Class Prior not implemented for categoricals.")
```
stephenbach: "categorical classes" (spell it out for users)

snorkel/learning/gen_learning.py
```diff
@@ -609,11 +592,12 @@ def _compile(self, L, init_deps, init_class_prior, LF_acc_priors, is_fixed):
 
         for optional_name in GenerativeModel.optional_names:
             if getattr(self, optional_name):
+                if optional_name != 'lf_propensity' and cardinality != 2:
+                    raise NotImplementedError(optional_name + " not implemented for categoricals.")
```
stephenbach: "categorical classes"

snorkel/learning/gen_learning.py
```diff
@@ -636,21 +620,27 @@ def _compile(self, L, init_deps, init_class_prior, LF_acc_priors, is_fixed):
...
-            for i in range(len(mat.data)):
-                f_off, ftv_off, w_off = self._compile_dep_factors(L, factor, f_off, ftv, ftv_off, w_off,
-                                                                  mat.row[i], mat.col[i],
-                                                                  dep_name_map[dep_name][0],
-                                                                  dep_name_map[dep_name][1])
-
+            if mat.nnz > 0:
+                if dep_name not in CATEGORICAL_DEPS and cardinality != 2:
+                    raise NotImplementedError(
+                        dep_name + " not implemented for categoricals.")
```
stephenbach: "categorical classes"

stephenbach: I'm not going to mark anymore of these :)

snorkel/learning/gen_learning.py
```diff
@@ -535,42 +506,55 @@ def _compile(self, L, init_deps, init_class_prior, LF_acc_priors, is_fixed):
             weight[i]['isFixed'] = False
             weight[i]['initialValue'] = np.float64(init_deps)
 
-
         #
         # Compiles variable matrix
         #
+        # Internal representation:
```
stephenbach: Why does the internal representation not match the way users write LFs, i.e., abstain is 0?

snorkel/learning/gen_learning.py
```diff
@@ -314,35 +241,39 @@ class prior (in log scale)
...
                  tend to mix poorly when there are many very accurate labeling
                  functions. In this case, this function will assume that the
                  classes are approximately balanced.
         """
         if self.fg is None:
-            raise ValueError("Must fit model with train() before computing diagnostics.")
+            raise ValueError(
+                "Must fit model with train() before computing diagnostics.")
 
         burnin = 500
```
stephenbach: This seems excessive, even to me... Is this for the class imbalanced -> slow mixing issue we discussed? How long does this method take to run?

snorkel/learning/gen_learning.py
```diff
@@ -169,13 +74,13 @@ class GenerativeModel(object):
...
-    :param lf_propensity: whether to include labeling function propensity 
+    :param lf_propensity: whether to include labeling function propensity
         factors
-    :param lf_class_propensity: whether to include class-specific labeling 
+    :param lf_class_propensity: whether to include class-specific labeling
         function propensity factors
     :param seed: seed for initializing state of Numbskull variables
     """
-    def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False, 
+    def __init__(self, class_prior=False, lf_prior=False, lf_propensity=False,
```
stephenbach: My understanding is that if you run this code against an old version of Numbskull with binary classes, it will silently do the wrong thing. That seems *very bad*.

I don't think we can cleanly handle the new NS / old Snorkel case, but can we at least raise an error in the new Snorkel / old NS case by checking for some flag set in the NS package?

snorkel/annotations.py
bryan-he: Marginals isn't formally the right term, posteriors is. We could change to that?

snorkel/learning/gen_learning.py
bryan-he: This runs pretty quickly (for test_categoricals.py, it's about 0.6 seconds).

The key point is that the factor graph at this point is reduced to just 1 data point (https://github.com/HazyResearch/snorkel/blob/categorical/snorkel/learning/gen_learning.py#L231), since we aren't looking at the values the LFs give anymore.

Should I cut the number of total iterations down, or is this fast enough?

snorkel/learning/gen_learning.py
bryan-he: This cleans up the factor function definitions (now, internally, the true class labels and labeling functions match except in the case of abstain, instead of them being off-by-one).

snorkel/learning/gen_learning.py
stephenbach: Eh, good enough for now :)

PR #654, opened 2017-06-02
Snark

stephenbach:


snorkel/contrib/snark/annotations.py
```diff
@@ -0,0 +1,147 @@
...
+from snorkel.models.views import create_serialized_candidate_view
+
+
+class SparkLabelAnnotator:
+    """
+    Distributes candidates to a Spark cluster and applies labeling functions over them.
+
+    Distributed candidate sets are cached by split for the annotator, so repeated calls to apply() will
+    operate over the same split of candidates as the first call for that split, regardless of how the
+    underlying database changes. To reload a candidate set split, construct a new SparkLabelAnnotator.
```
ajratner: This could be pretty bad failure mode if user doesn't read doc string... should print warning or something?

ajratner: Maybe have this caching be an option at very least?

snorkel/contrib/snark/annotations.py
```diff
@@ -0,0 +1,147 @@
...
+        key_query = LabelKey.__table__.insert()
+        label_query = Label.__table__.insert()
+
+        for lf in LFs:
+            lf_id = self.snorkel_session.execute(key_query, {'name': lf.__name__, 'group': 0}).inserted_primary_key[0]
+
+            labels = self.split_cache[split].map(lambda c: (c.id, lf(c)))
+            labels.filter(lambda (_, value): value != 0 and value is not None)
+            for cid, value in labels.toLocalIterator():
+                self.snorkel_session.execute(label_query, {'candidate_id': cid, 'key_id': lf_id, 'value': value})
```
ajratner: Should do bulk insert here?

snorkel/contrib/snark/annotations.py
```diff
@@ -0,0 +1,147 @@
...
+# Note: This should ideally not be hard-coded...
+CONTEXT_OFFSET = 2
+# Note: We store the sentence here, not the sentence_id
+SPAN_COLS = ['id', 'sentence', 'char_start', 'char_end', 'meta']
+SENTENCE_COLS = ['id', 'document_id', 'position', 'text', 'words', 
+    'char_offsets', 'lemmas', 'pos_tags', 'ner_tags', 'dep_parents',
+    'dep_labels', 'entity_cids', 'entity_types']
+
+
+def wrap_candidate(row, class_name='Candidate', argnames=None):
```
ajratner: Move this to candidate.py

ajratner:
One other issue to consider: running notebook 1 to prepare the data is slow... and seems odd that this is not done in Spark.  Potential solutions:
* Preprocess and commit zipped version so no need to re-run
* Make Spark preprocessor as well

Will think about after everything else done

PR #657, opened 2017-06-04
Crowdsourcing tutorial

thodrek:
Please review text in notebooks.


ajratner:
Actually one other change: I think we should separate the Spark / preprocessing stuff from the rest of the tutorial.  If people don't want to install spark and/or run all of the preprocessing this shouldn't block them from the rest of the tutorial

PR #660, opened 2017-06-05
Scoped categorical

ajratner:
Adds support for "scoped" categorical mode in `GenerativeModel` by setting `GenerativeModel.train(..., scoped_categorical=True)`.  In this setting, we assume there is a large cardinality but that each candidate has a much smaller effective cardinality; entries are re-mapped to row-specific values so that sampling is not massively inefficient.

Notes:
* No support currently for pre-specified per-candidate support sets; these are inferred as the union of all LF labels applied to a given candidate
* +1 smoothing is used w.r.t. cardinality: when computing the LF accuracies, we add +1 to the inferred cardinality to avoid weirdness when e.g. inferred cardinality would be 1 (could be quite common)
* A sparse matrix is returned (re-mapped to the correct original value range)

### DIFF
--- a/snorkel/learning/gen_learning.py
+++ b/snorkel/learning/gen_learning.py
@@ -67,8 +67,8 @@ class GenerativeModel(object):
         'dep_similar', 'dep_fixing', 'dep_reinforcing', 'dep_exclusive'
     )
 
-    def train(self, L, deps=(), LF_acc_priors=None, LF_acc_prior_default=0.7, 
-        labels=None, label_prior=0.99, init_deps=0.0,
+    def train(self, L, deps=(), LF_acc_prior_weights=None, LF_acc_prior_weight_default=1,
+        labels=None, label_prior_weight=5, init_deps=0.0,
         init_class_prior=-1.0, epochs=30, step_size=None, decay=1.0,
         reg_param=0.1, reg_type=2, verbose=False, truncation=10, burn_in=5,
         cardinality=None, timer=None, scoped_categorical=False):
@@ -87,12 +87,12 @@ class GenerativeModel(object):
                      element is a tuple of the form 
                      (LF 1 index, LF 2 index, dependency type),
                      see snorkel.learning.constants
-        :param LF_acc_priors: An N-element list of prior probabilities for the
+        :param LF_acc_prior_weights: An N-element list of prior weights for the
             LF accuracies
-        :param LF_acc_prior_default: Default prior probability for each LF 
-            accuracy; if LF_acc_priors is unset, each LF will have this prior
+        :param LF_acc_prior_weight_default: Default prior for the weight of each LF
+            accuracy; if LF_acc_prior_weights is unset, each LF will have this
         :param labels: Optional ground truth labels
-        :param label_prior: The prior probability that the ground truth labels
+        :param label_prior_weight: The prior probability that the ground truth labels
             (if provided) are correct
         :param init_deps: initial weight for additional dependencies, except
                           class prior (in log scale)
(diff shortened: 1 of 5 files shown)


ajratner:
@stephenbach Ready for review!

snorkel/learning/gen_learning.py
```diff
@@ -134,12 +87,12 @@ def train(self, L, deps=(), LF_acc_priors=None, LF_acc_prior_default=0.7,
...
-        :param LF_acc_priors: An N-element list of prior probabilities for the
+        :param LF_acc_prior_weights: An N-element list of prior weights for the
             LF accuracies
-        :param LF_acc_prior_default: Default prior probability for each LF 
-            accuracy; if LF_acc_priors is unset, each LF will have this prior
+        :param LF_acc_prior_weight_default: Default prior for the weight of each LF
+            accuracy; if LF_acc_prior_weights is unset, each LF will have this
         :param labels: Optional ground truth labels
-        :param label_prior: The prior probability that the ground truth labels
+        :param label_prior_weight: The prior probability that the ground truth labels
```
stephenbach: Clarify that it's in log scale.

snorkel/learning/gen_learning.py
```diff
@@ -159,6 +112,10 @@ class prior (in log scale)
         :param cardinality: number of possible classes; by default is inferred
             from the label matrix L
         :param timer: stopwatch for profiling, must implement start() and end()
+        :param scoped_categorical: If True, remap the values of a categorical
+        label matrix so that the support for each candidate is a contiguous
+        range, so that only relevant values are sampled during learning /
```
stephenbach: This needs to be clearer. That they're remapped to be contiguous is an implementation detail.

snorkel/learning/gen_learning.py
```diff
@@ -210,11 +168,42 @@ class prior (in log scale)
             L = sparse.csr_matrix(L)
         L = L[idxs, :]
 
+        # Optionally: Remap the values of L (in categorical setting) so that the
```
stephenbach: Recommend moving this to another branch and doing user-provided scoping for now. :)

snorkel/learning/gen_learning.py
```diff
@@ -134,12 +87,12 @@ def train(self, L, deps=(), LF_acc_priors=None, LF_acc_prior_default=0.7,
                      element is a tuple of the form 
                      (LF 1 index, LF 2 index, dependency type),
                      see snorkel.learning.constants
-        :param LF_acc_priors: An N-element list of prior probabilities for the
+        :param LF_acc_prior_weights: An N-element list of prior weights for the
             LF accuracies
-        :param LF_acc_prior_default: Default prior probability for each LF 
-            accuracy; if LF_acc_priors is unset, each LF will have this prior
+        :param LF_acc_prior_weight_default: Default prior for the weight of each LF
```
stephenbach: Ah, actually none of these make clear it's in log scale. That's my bad. :)  Maybe just a note above in the doc string that all weights are in log scale?

snorkel/learning/gen_learning.py
```diff
@@ -367,31 +365,45 @@ def marginals(self, L):
...
+            return marginals
 
         # Categorical setting
         else:
-            marginals = np.zeros((m, self.cardinality), dtype=np.float64)
+            all_marginals = []
+
+            # Get the marginal (posterior) probability for each candidate
             for i in range(m):
+                marginals = np.zeros(self.cardinalities[i], dtype=np.float64)
```
stephenbach: Do we need to tell users how to interpret indices of this vector if their LFs output strings, etc.? (Think I already saw some comment about this...)

snorkel/learning/gen_learning.py
ajratner: @stephenbach any thoughts on what to do if there are labels outside of the specified range for a candidate?  Default is just throw an error (what I plan to do for now), but also could union with the support set / leave this as an option...

stephenbach: Throw an error is reasonable.

PR #671, opened 2017-06-19
Save pickle of LFs with all dependencies

ajratner:
Also moves the input of the annotation generator to `Annotator.apply`, which seems cleaner as this function can change after the `Annotator` is instantiated...

This is just a stub- ideas on how to configure this LF pickling?  Simplest temporary solution would be to just hard-code as True for the workshop


ajratner:
Getting the old `The command "conda update --yes conda" failed and exited with 1 during .` error on one of the Travis builds... can try to re-un later

ajratner:
Confused about what's going on with Travis.... tested locally and changes work fine

ajratner:
Tests make sense now; just need to change the CDR / categorical notebooks for new `LabelAnnotator.apply` syntax; can do when done w something else

ajratner:
@jason-fries @stephenbach ready to be reviewed / merged!

Issue #569, opened 2017-02-18
Make all learning code work with `session=None`


ajratner:
A lot of the learning code (by design) works just with generic `scipy.sparse` matrices, however some of it--e.g. scoring, hyperparam search--needs a `session` object and will throw an error.

We should just add in some default modes so that these methods can run with `session=None`; this should both be relatively easy *and* super high impact :)!

ajratner:
@henryre we can work on this early next week maybe, if I don't get to myself this weekend

henryre:
If we assume that the candidate/feature input and the label matrix are row-aligned, we really don't need the session at all (correct me if I'm wrong). Seems like a reasonable requirement.

ajratner:
Yes exactly

PR #681, opened 2017-06-23
TensorFlow end model bindings refactor + fixes

ajratner:
Main changes:
- Fixes the grid search issue, but more broadly refactors + fixes model saving / loading (now, any keyword arguments passed into `_build` will automatically be saved / loaded for reconstructing the model.  These are referred to as `model_kwargs` and include anything needed to build the model and/or apply it at test time; this is opposed to `training_kwargs` which are only needed to train the model
- Creates a new graph & session for each training session to keep graphs separate and clean, fixing e.g. `GridSearch`
- Moves training loop into parent class so more code reuse, easier to plug in new models
- General fixes + refactoring
- Adds a new test (notebook)
- Reorganizes contrib dir, all models in there (open to other options here)


ajratner:
Unfortunately the changes were extensive and so, along with the renaming, the diffs are not all appropriately showing up... happy to discuss this PR in person / over slack to make easier to read.  Shouldn't change much in terms of existing functionality though!

ajratner:
One thing I'd love review feedback on: the scoring functions.  In trying to get it to pass tests, I wrote a lighter-weight `score` function which just computes the score--and assumes that the candidates and labels are collated, thus requires no `session` object (per #569 )--and renaming the old `score` function as `error_analysis`.
* One question: does this make sense / seem like right way to go?
* Also, we should probably make sure the generative and discriminative models are in sync on score functions otherwise will be confusing... not sure what right way to do this is

ajratner:
One other thing: the scores in the intro tutorial after grid search seem pretty low.  This might be a new error I introduced (?) Or just a function of overfitting to a minuscule dev set.  Don't know whether we want to fix this in this PR or after putting in the new dataset.

ajratner:
@stephenbach @jason-fries @henryre ready for review :D!

ajratner:
@stephenbach @jason-fries ready for review again!

PR #688, opened 2017-06-28
Partially fixes #582

### LINKED ISSUE #582: Array slicing breaks csr_AnnotationMatrix objects
I'd like to take a subset of candidate instances to train the generative model and then save those marginals for use in the discriminative model. 

I thought this could be done just taking a subset slice of the label matrix.  However, when I use  slice notation on a csr_AnnotationMatrix object to generate my model input, I run into errors when attempting to save the marginals later on in the program. For example:

```
L_train = labeler.load_matrix(session, split=0)
print L_train.get_candidate(session,0)
```
will correctly access and the mapped candidate object. 
However, if we take a slice of the label matrix, e.g., 
```
L_train_subset = L_train[0:n,...]
print L_train_subset.get_candidate(session,0)
```
the `get_candidate` function breaks with the following error
```
File "/home/user/code/dd-patient-notes/snorkel/snorkel/annotations.py", line 39, in get_candidate
    return session.query(Candidate).filter(Candidate.id == self.row_index[i]).one()
TypeError: 'NoneType' object has no attribute '__getitem__'
```
everything works otherwise in terms of training, just the mapping between candidate and row index seems to break.

ajratner:
I got this working for slices (e.g. `a:b`) but can't figure out how to handle / what subfunction of `csr_matrix` class handles slicing by array / list... have to pause on this for now, maybe is worth pulling in as is, or someone else can pick up from where I left off?

### DIFF
--- a/snorkel/annotations.py
+++ b/snorkel/annotations.py
@@ -55,6 +55,29 @@ class csr_AnnotationMatrix(sparse.csr_matrix):
         """Return the cow index of the AnnotationKey"""
         return self.key_index[key.id]
 
+    def _get_sliced_indexes(self, s, axis, index):
+        idxs = np.arange(self.shape[axis])[s]
+        index_new, inv_index_new = {}, {}
+        for i, k in index.iteritems():
+            if i in idxs:
+                i_new = np.where(idxs == i)[0][0]
+                index_new[i_new] = k
+                inv_index_new[k] = i_new
+        return index_new, inv_index_new
+
+    def _get_submatrix(self, row_slice, col_slice):
+        # Get the slice of the matrix
+        X = super(csr_AnnotationMatrix, self)._get_submatrix(row_slice, 
+            col_slice)
+        X.annotation_key_cls = self.annotation_key_cls
+
+        # Remap the row and column indexes
+        X.row_index, X.candidate_index = self._get_sliced_indexes(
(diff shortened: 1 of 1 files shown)


ajratner:
Hey @jason-fries or @stephenbach thoughts on merging this into master?  The issue we can only do slices like `X[a:b]` but not `X[s]` where `s` is an array... but might as well put in the former capability for now (won't negatively change anything)?

jason-fries:
I like the flexibility of Numpy's slice notation. I put some thoughts earlier on this issue at https://github.com/HazyResearch/snorkel/issues/582#issuecomment-311566563

I'd vote to keep things as it, but i don't feel that strongly.

ajratner:
Sorry I could have been clearer: Numpy's standard slice notation is already in place.  The only problem (without this PR) is that using slicing in any way breaks the index that allows you to use the e.g. `get_candidate` or `get_key` helper methods.

This PR just extends the set of slice operations where the index will not be broken (to include all `a:b`-type ones).  So it's a minimal but purely additive fix

Also, as of latest changes (on `new_intro` branch), the learning / grid search methods (basically everything but the `classifier.error_analysis` method) will work without this index anyways (i.e. #569 has been fixed)

stephenbach:
LGTM!

PR #702, opened 2017-07-10
Python 3 compatibility

hangyao:
Changed print statement to print().
Changed dict.iteritems() to dict.items().
Changed xrange() to range().

### DIFF
--- a/snorkel/annotations.py
+++ b/snorkel/annotations.py
@@ -19,7 +19,7 @@ from .utils import (
     matrix_fn,
     matrix_tn
 )
-
+from future.utils import iteritems
 
 class csr_AnnotationMatrix(sparse.csr_matrix):
     """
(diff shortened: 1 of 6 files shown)


hangyao:
@ajratner, would you like to review and merge this refixed PR?

Thank you!

snorkel/annotations.py
```diff
@@ -61,7 +61,7 @@ def _get_sliced_indexes(self, s, axis, index):
         else:
             idxs = np.array([s])
         index_new, inv_index_new = {}, {}
-        for i, k in index.iteritems():
+        for i, k in index.items():
```
ajratner: I think this loads them all into memory- would like to avoid this.  Unless am wrong here?

snorkel/db_helpers.py
```diff
@@ -37,7 +37,7 @@ def reload_annotator_labels(session, candidate_class, annotator_name, split, fil
 
         # Assemble query and check
         candidate_query = session.query(candidate_class)
-        for k, v in candidate_args.iteritems():
+        for k, v in candidate_args.items():
```
ajratner: Ditto re: comment above

ajratner:
Looks great, thanks!  Should I merge in after tests pass?

hangyao:
Yes I think it is good to go. Thank you.

Issue #705, opened 2017-07-10
save_dir parameter not respected in GridSearch and GenerativeModel


jason-fries:
Even if we provide a custom `save_dir` all model checkpoints are still written to `checkpoints`. This cause problems when running multiple jobs using grid search on the same machine, since model snapshots overwrite each other.

PR #706, opened 2017-07-10
save_dir param used during checkpointing, fixes issue #705

jason-fries:
Fixes saving issues with GridSearch. This is needed to run multiple, simultaneous jobs.


ajratner:
I.e. why not just pass in `save_dir` to the `fit` method, and then pass it through to the `model.save` call?

jason-fries:
That works for discriminative models, where `save_dir` is already a parameter of `TFNoiseAwareModel`'s `train` method. For the `GenerativeModel` , the train class doesn't include the `save_dir` param which causes an error when it's passed into `fit`. I just added it for expediency's sake, so I'm not wedded to where it goes.

PR #708, opened 2017-07-11
Parallel grid search fixes

ajratner:


jason-fries:
@ajratner - just ran through on EHR task.
Everything LGTM!

PR #717, opened 2017-07-13
Adds psycopg2 dependency for PostgreSQL

MattMorgis:
First time trying to use PostgreSQL, the user is always given the following message:
```
ImportError: No module named psycopg2
```

This adds `psycopg2` to the list of dependencies


ajratner:
Thanks!!

PR #725, opened 2017-07-20
Make candidates with repeated mention types deterministic.

bhancock8:
In the spouse example, there is a sentence that mentions Daniel Ammann first and Pernilla Amman second. However, running the tutorial repeatedly, sometimes Daniel Ammann is arg0, and sometimes he's arg1. It'd be nice for the candidates that are created to be deterministic, and in particular, for arg0 to refer to the first person in the sentence and arg1 refer to the second person. In candidates where the mentions are of different types, this isn't an issue. This PR ensures that for candidates where the mentions are of the same type, the mentions have the same order in the candidate as they do in the sentence.

It also updates a doc string to list the correct default value.

### DIFF
--- a/snorkel/candidates.py
+++ b/snorkel/candidates.py
@@ -25,7 +25,7 @@ class CandidateExtractor(UDFRunner):
     :param nested_relations: Boolean indicating whether to extract Candidates that relate one Context with another
                              that contains it. Only applies to binary relations. Default is False.
     :param symmetric_relations: Boolean indicating whether to extract symmetric Candidates, i.e., rel(A,B) and rel(B,A),
-                                where A and B are Contexts. Only applies to binary relations. Default is True.
+                                where A and B are Contexts. Only applies to binary relations. Default is False.
     """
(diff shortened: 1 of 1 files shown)


snorkel/candidates.py
```diff
@@ -88,12 +88,14 @@ def apply(self, context, clear, split, **kwargs):
...
-                # "symmetric" relations
+                # "symmetric" relations. For symmetric relations, if mentions are of the same type, maintain
+                # their order in the sentence.
                 if not self.self_relations and a == b:
                     continue
                 elif not self.nested_relations and (a in b or b in a):
                     continue
-                elif not self.symmetric_relations and (b, a) in extracted:
+                elif not self.symmetric_relations and ((b, a) in extracted or
+                    (self.matchers[0] == self.matchers[1] and a.char_start > b.char_start)):
```
ajratner: I'm not sure that this `self.matchers[0] == self.matchers[1]` check will always work... but this seems fine to add in either way!

ajratner:
Can merge in once tests pass, thanks!

PR #734, opened 2017-07-22
Fixes unfinished / incorrect marginals calculation for scoped categorical

ajratner:
This originally started as a branch to add to the save/load functions for the `GenerativeModel`, as for categorical vars was throwing an error about not having `self.cardinalities`... but then I realized that we would need to recalculate this for an unseen label matrix at test time anyway, which somehow had never been implemented, so added that.

Passes tests locally, although we've been having weird travis timeout issues so may fail due to that...


stephenbach:
LGTM

PR #737, opened 2017-07-27
Remove future imports

henryre:
Unused and can break compatibility

### DIFF
--- a/snorkel/annotations.py
+++ b/snorkel/annotations.py
@@ -19,7 +19,6 @@ from .utils import (
     matrix_fn,
     matrix_tn
 )
-from future.utils import iteritems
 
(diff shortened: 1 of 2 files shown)


ajratner:
I think this might be for Python 3 compatibility?  See https://github.com/HazyResearch/snorkel/pull/702/files#diff-398030fcfc468a8ee24a7f8c21053a96L22

henryre:
Well they're both unused, and sometimes conda messes up the six/future install. I vote we just remove.

ajratner:
There's one usage in `db_helpers.py`

henryre:
It's not using this import (just the `dict` method), so that call breaks in Python 3. I have a few more Python 3 fixes, so let me close this for another PR.

PR #738, opened 2017-07-27
Load marginals custom query

ajratner:


ajratner:
@stephenbach ready for review!

Issue #509, opened 2016-12-13
Python 3 compatability


dhimmel:
I didn't see [any dependencies](https://github.com/HazyResearch/snorkel/blob/40dfdc586b72412f9825b0f5f16d02b0ed98626a/python-package-requirement.txt) that are 3 incompatible. Is the lack of Python 3 support just because it's not a priority?

It would be nice to snorkel in Python 3 pipelines. Especially since for analyzing text the default unicode strings would be convenient.

I tend to be extreme on these issues, but I'd even advocate for dropping 2 support to make development cleaner.

This is not a major issue for me, as workarounds are easy -- but just wanted to get the perspectives of the developers.

ajratner:
@dhimmel primarily just busy :)  We will most likely want to move to Python 3 at some point once we catch up on other development priorities!

ajratner:
Should be closed now!

dhimmel:
Tagging @danich1. Hopefully we can upgrade `greenelab/snorkeling` to python 3 now.

PR #741, opened 2017-07-28
Python 3 fixes

henryre:


dhimmel:
@danich1 and I are really interested in upgrading to Python 3 for our [snorkeling project](https://github.com/greenelab/snorkeling). @ajratner I'm happy to help review this PR, if that will speed things up.

@henryre do you think you could resolve conflicts and update this PR for the current repo state? @ajratner is `master` the right branch for this PR?

Also @henryre do you have a sense of how many additional fixes are needed for full Python 3 compatability?

snorkel/learning/disc_learning.py
```diff
@@ -205,8 +205,7 @@ def train(self, X_train, Y_train, n_epochs=25, lr=0.01, batch_size=256,
                 epoch_losses.append(epoch_loss)
 
             # Reshuffle training data
-            train_idxs = range(n)
-            np.random.shuffle(train_idxs)
+            train_idxs = np.random.permutation(n)
```
dhimmel: Not Python 3 related but does appear to do the same thing as before with more brevity.

snorkel/parser/parser.py
```diff
@@ -19,8 +19,11 @@ def to_unicode(self, text):
         :return:
         '''
         text = text.encode('utf-8', 'error')
-        text = text.decode('string_escape', errors='ignore')
-        text = text.decode('utf-8')
+        try:
```
dhimmel: Perhaps a comment to make it more explicit which route is Py2 versus Py3

dhimmel:
Refs https://github.com/HazyResearch/snorkel/issues/509

henryre:
This is a pretty outdated and incomplete diff. In general, I don't think the patch-based approach to 2/3 compatibility is worth it for a small team (harder to maintain, inevitable breaks and code rot, etc.). My recommendation would be going all in on Python 3 for future releases (e.g. https://github.com/numpy/numpy/blob/master/doc/neps/dropping-python2.7-proposal.rst)

@ajratner @stephenbach

dhimmel:
> My recommendation would be going all in on Python 3 for future releases

I agree and usually make my packages Python 3 only. Support 2 & 3 is a hassle and takes a deep understanding of the differences.

But the question is whether the maintainers feel ready to drop Python 2 support? I personally would encourage it since snorkel is still early in its overall life cycle.

And unless the plan is to stay Python 2 forever, sooner is probably better than later.

bjelkenhed:
Python 3 support would be really great, it is very hard to make things work well in Python 2 for other languages than English (Swedish in our case). Snorkel is currently the only tool we use that requires Python 2.

ajratner:
Hi guys,

Sorry for delayed response!  We will indeed be switching to Python3 in the next version of Snorkel (v0.7). However, this probably won't be happening for another one or two months, so I'd suggest going ahead with the patches to support 2 + 3 in v0.6.  Happy to help a bit here and any outside help is much appreciated!  Can either wrap this PR up or start a clean one

Thanks!
Alex

dhimmel:
> any outside help is much appreciated

Cool. Tag me on any Python related PRs and I'm happy to help with the PR review burden.

ajratner:
Thanks!

ajratner:
Closing- see PR #847

PR #745, opened 2017-07-29
Pass in the dev/validation set to the model.train function in GridSearch

ajratner:


ajratner:
@stephenbach ready whenever passes tests!

PR #747, opened 2017-07-30
Fixes #746: Batch marginals to avoid OOM on GPU

### LINKED ISSUE #746: Need to batch test set eval for TF models
Otherwise get OOM errors running on GPU with larger models

ajratner:


snorkel/learning/disc_learning.py
```diff
@@ -237,6 +237,21 @@ def train(self, X_train, Y_train, n_epochs=25, lr=0.01, batch_size=256,
...
+        Compute the marginals for the given candidates X.
+        Split into batches to avoid OOM errors, then call _marginals_batch.
+        """
+        N = len(X) if self.representation else X.shape[0]
+        n_batches = int(np.floor(N / batch_size))
+
+        # Iterate over batches
+        batch_marginals = []
+        for i, b in enumerate(range(0, N, batch_size)):
+            X_batch = X[b:min(n, b+batch_size)]
```
henryre: not necessary: just do `b+batch_size`

henryre:
Looks like it works, but would have 
1. an option to run all at once (`batch_size=-1`)
2. it default to this since it will be faster for most uses

PR #750, opened 2017-07-31
elasticsearch

jasontlam:


ajratner:
@jasontlam tag me whenever ready for review again- excited!!

jasontlam:
@ajratner

elasticSnorkel.ipynb
```diff
@@ -0,0 +1,410 @@
+{
```
ajratner: I would put this notebook in the `tutorials/advanced` directory

elasticSnorkel.ipynb
```diff
@@ -0,0 +1,410 @@
...
+      "health status index  uuid                   pri rep docs.count docs.deleted store.size pri.store.size\n",
+      "yellow open   corpus VlTi5iDNQXi2cLp9sYAldQ   5   1      67820            0     45.6mb         45.6mb\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "count = 0\n",
+    "print \"Beginning Indexing\"\n",
+    "for p in session.query(Document):\n",
```
ajratner: Here is where we should really both (1) automate and (1) generalize:

First, the goal should be to make this a one liner.  We can assume that the context hierarchy being used is just the default one (`Document` -> `Sentence` -> `Span`) and that the user wants to index their entire set of loaded contexts in the database, to make things simple.

Second, and more important: this code is currently only useful if someone is focused on candidates which are pairs of NER tags, which is only a small amount of Snorkel applications.  The spec we should aim for is to take in an arbitrary set of candidates (again, feel free to assume we want to just use all candidates) and then can just encode these candidates in a similar serialized format, to keep the rest of the code mostly consistent

elasticSnorkel.ipynb
```diff
@@ -0,0 +1,410 @@
...
+     ]
+    }
+   ],
+   "source": [
+    "distance = 100\n",
+    "value = \"married\"\n",
+    "candField = \"candidates\"\n",
+    "candTag = \"person\"\n",
+    "sentField=\"sentence\"\n",
+    "result = eSearch.searchBetweenCandidates(candField,candTag,sentField,value,distance)\n",
```
ajratner: Can we make this a bit more general?  What about just calling it something like 'search', and then have some keyword argument (e.g `position`) with values:
* 'entire'
* 'between'
* 'left'
* 'right'

This would then subsume both this and the previous `searchIndex` example!

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "## Intializing Elasitcsearch\n",
+    "\n",
+    "We begin by intializing Elasticsearch with the names of our index and type. You should have Elasticsearch installed and already run the \"bin/elasticsearch\" command. For more information you can visit: https://www.elastic.co/guide/en/elasticsearch/guide/current/running-elasticsearch.html.\n",
+    "\n",
+    "Elasticsearch uses slightly different terminology from tradational databases. \n",
```
ajratner: Is this info necessary?  Cut and point to ES docs?

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "from elastics import elasticSession ,printResults,deleteIndex\n",
+    "\n",
+    "# import os\n",
+    "\n",
+    "# # TO USE A DATABASE OTHER THAN SQLITE, USE THIS LINE\n",
+    "# # Note that this is necessary for parallel execution amongst other things...\n",
+    "# os.environ['SNORKELDB'] = 'postgres:///snorkel-intro'\n",
+    "\n",
+    "session = SnorkelSession()\n",
+    "eSearch=elasticSession(\"corpus\",\"articles\",\"sentence\")"
```
ajratner: Do they need to provide these args (and what are they- "corpus", "articles"?)  Can we skip this and just assume default context hierarchy?

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Check Connection\n",
+    "\n",
+    "We begin indexing, we can check our connection with Elasticsearch as well as our existing indices. "
```
ajratner: Cut this step?  Do they need to do this?

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+      " \n",
+      "health status index  uuid                   pri rep docs.count docs.deleted store.size pri.store.size\n",
+      "yellow open   corpus HCOENVAeRoKTrHduXdVChQ   5   1      67820            0       24mb           24mb\n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "eSearch.elasticIndex(Document)\n",
+    "eSearch.getIndices()"
```
ajratner: Can this just be run automatically when the `elasticSession` object is created?  That would be way cleaner

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Visualize Index\n",
+    "\n",
+    "To better visualize our data we get a mapping of index. We can see here that the each document in articles has 3 different fields: lineNum,candidates and sentence."
```
ajratner: Skip this?  Seems like a detail user shouldn't need to know

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+   ],
+   "source": [
+    "eSearch.getIndexMap()"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Visualizing a Document\n",
```
ajratner: Skip this too (or put this kind of stuff at the end)?  Again, goal should be to get a user up and using ES in a few lines!  Otherwise, realistically, no one will use...

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+      "-------------------\n",
+      "sentence\n",
+      "All of the women were married (or divorced!), and some had children.   \n",
+      "\n"
+     ]
+    }
+   ],
+   "source": [
+    "query=\"married children\"\n",
+    "searchResult = eSearch.searchIndex(\"entire\",query)\n",
```
ajratner: Could we maybe have kwargs to make syntax cleaner?

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+    "query=\"white trousers\"\n",
+    "searchResult = eSearch.searchIndex(\"entire\",query,distance=0,size=3)\n",
+    "printResults(searchResult,\"sentence\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "### Positional Search\n",
```
ajratner: The utility of this one made less sense to me (why can't we just use phrase search for multi-word?), maybe cut or put at end?

tutorials/advanced/elasticSnorkel.ipynb
```diff
@@ -0,0 +1,583 @@
...
+   "source": [
+    "searchResult = eSearch.searchIndex(\"position\",\"Pennsylvania\",\"Colorado\",distance =20)\n",
+    "printResults(searchResult,\"sentence\")"
+   ]
+  },
+  {
+   "cell_type": "markdown",
+   "metadata": {},
+   "source": [
+    "# Creating the Candidate Field"
```
ajratner: Can we just cut all this setup stuff and have the user feed the candidate class, e.g. `Spouse`, into the `elasticSession` constructor, and then have this stuff done automatically on init?  Is there a reason not to do this?  E.g. is this very slow?

ajratner:
All good- just what we discussed on slack!

ajratner:
It looks much better- I realized though the one thing I think that's missing (and sorry for not catching before).  Can we have the search return Snorkel objects?  Right now the only thing the user can do is print out results... if the search returned `Sentences` and/or `Candidates`, however, then (a) these could be viewed in the `Viewer` with no extra work; (b) the user could then actually _use_ these results for testing LFs, etc.

Thoughts?  Any roadblocks to this?  Hopefully it's a small change, then everything else LGTM. If not, let's chat on slack / in person so I understand.  Thanks!!

ajratner:
Minor changes only:
- Change e.g. `elasticSession` -> `ElasticSession`, `searchIndex` -> `search_index`
- Mention in the tutorial that `Sentence`/`Candidate` objects are returned
- Is `size` the number of results to return, and does it default to 5?
- Is `slop` a standard elastic search keyword, or something you added?  If former, maybe just link to their defn?  If latter, be a little clearer on explanation: is this word-level edit distance (i.e. number of words that would need to be added/removed/moved)?
- The text under "Positional Search" needs to be cleaned up / edited; also, I'd suggest removing this / moving to bottom, I think it's more in the weeds / adv functionality that most people won't need
- change "{in,b,a}Cand" -> "{between,before,after}_cand"
- Remove the "Visualizing a Document" section

Then we should merge and get some feedback from others in the lab! (once tests pass)

Issue #718, opened 2017-07-13
intro tutorial super slow on gen_model.marginals(L_train)


bhancock8:
I'm running the intro tutorial out of the box with a fresh install of v0.6, and in Intro_Tutorial_2, the call "train_marginals = gen_model.marginals(L_train)" is taking over 30 minutes; it is making it through the triple nested loops, just very slowly. I'm assuming this is unexpected behavior?

stephenbach:
Yup, definitely unexpected. Not seeing this locally. Can you confirm you're actually on 0.6.1?

bhancock8:
On master and up-to-date with origin/master. Pulled 10 minutes ago and relaunched the notebooks; problem still there.

bhancock8:
Looks like other funny business is going on. I'm now seeing it hang for "self.clear(session, **kwargs)" in UDFRunner class as well. Don't see how the two are connected yet, but I'll keep digging and report back here.

bhancock8:
With 500 documents it runs in 20 seconds. I assume that's still unusual? The tutorial completes with reasonable results.

bhancock8:
I cleared the database and ran again with the full set of spouse documents. It took 7.5 mins for the gen_model.marginals(L_train) command, and the final F1 score was 0.41. I think it's right; it's just slow?

bhancock8:
It happens for both the CDR tutorial (1 min for 8272 marginals) and intro tutorial (7.5 min for 22195 marginals)--my final F1 scores and the dimensions (and nnz) for L_train match the saved values in the tutorials. For reference, how long does this line take when you run it in either tutorial?

stephenbach:
Ah, ok, the issue is just a regression in how the label matrix subclass handles slicing. Reverting the commit that changed snorkel.annotations.csr_LabelMatrix drops the execution time on full intro docs to ~5 seconds. We're looking into how to fix or just revert. Will release an 0.6.2 shortly with a fix.

stephenbach:
A workaround in the meantime is

```
from scipy.sparse import csr_matrix

train_marginals = gen_model.marginals(csr_matrix(L_train))
```

ajratner:
@bhancock8 thanks for catching this!!  FYI fixing now, and regarding the `clear` operation being slow, we'll pin that to #719

bhancock8:
Beautiful! Thanks a bunch for the quick fixes!

ajratner:
No thanks for catching it so quickly :)!

PR #753, opened 2017-08-01
Annotation matrix slicing

### LINKED ISSUE #582: Array slicing breaks csr_AnnotationMatrix objects

bhancock8:
This modification enables the full range of normal numpy slicing on annotations matrices. 

Instead of just replacing _get_submatrix, replace the higher level built-in function __getitem__, and for instances where the result it not a matrix (e.g., L_train[x,y] returns a scalar), it recognizes that the row_index and col_index are gone and won't try to keep up with the extra bookkeeping.

Further extends/fixes #582 and #688.

### DIFF
--- a/snorkel/annotations.py
+++ b/snorkel/annotations.py
@@ -68,12 +68,7 @@ class csr_AnnotationMatrix(sparse.csr_matrix):
                 idxs = np.arange(self.shape[axis])[s]
         elif isinstance(s, int):
             idxs = np.array([s])
-<<<<<<< HEAD
-        else:
-            # s is an array of ints
-=======
         else: # s is an array of ints
->>>>>>> 9b25520e... Make csr_AnnotationMatrix.__getitem__ return an int or a csr_AnnotationMatrix.
             idxs = s
             # If s is the entire slice, skip the remapping step
             if idxs == range(len(idxs)):
@@ -89,17 +84,6 @@ class csr_AnnotationMatrix(sparse.csr_matrix):
     def __getitem__(self, key):
         X = super(csr_AnnotationMatrix, self).__getitem__(key)
 
-<<<<<<< HEAD
-        # Remap the row and column indexes if applicable.
-        if hasattr(X, 'row_index') and hasattr(X, 'col_index'):
-            X.annotation_key_cls = self.annotation_key_cls
-            
-            row_slice, col_slice = self._unpack_index(key)
-            X.row_index, X.candidate_index = self._get_sliced_indexes(
-                row_slice, 0, self.row_index, self.candidate_index)
-            X.col_index, X.key_index = self._get_sliced_indexes(
-                col_slice, 1, self.col_index, self.key_index)
-=======
         # If X is an integer, just return it
         if isinstance(X, int):
             return X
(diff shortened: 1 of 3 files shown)


bhancock8:
All clear!

bhancock8:
I haven't noticed any, but I'll do some profiling with and without to check it.

ajratner:
Yeah sorry for my delay in responding here!  In particular check for massive slowdowns in computing marginals, that's what we ran into before

ajratner:
Actually, for reference, here we are... some meddling contributor was the one who caught it :D... https://github.com/HazyResearch/snorkel/issues/718

snorkel/annotations.py
```diff
@@ -66,27 +66,30 @@ def _get_sliced_indexes(self, s, axis, index, inv_index):
...
-        # # Remap the row and column indexes
-        X.row_index, X.candidate_index = self._get_sliced_indexes(
-            row_slice, 0, self.row_index, self.candidate_index)
-        X.col_index, X.key_index = self._get_sliced_indexes(
-            col_slice, 1, self.col_index, self.key_index)
+    def __getitem__(self, key):
+        X = super(csr_AnnotationMatrix, self).__getitem__(key)
+
+        # Remap the row and column indexes if applicable.
+        if hasattr(X, 'row_index') and hasattr(X, 'col_index'):
```
ajratner: If you rely on the standard `__getitem__` method, I believe it will in some cases just return a generic `csr_matrix`, which will not have these attributes, e.g. see https://github.com/scipy/scipy/blob/v0.19.1/scipy/sparse/csr.py#L406.  Then the matrix returned won't have the row/col indexes...

ajratner: Otherwise this solution is much more elegant than before :)!

bhancock8: I modified it to make sure that all cases are covered. In short, if we indexed a single element, that just gets returned. If we index a submatrix or array, we explicitly make that returned value a csr_AnnotationMatrix and update its indexes before returning it. And then in the two files where we do a lot of read-only slicing (gen_learning and disc_learning), we explicitly convert to plain old csr matrices at the beginning (which is super easy; it just drops the auxiliary fields), so it won't bother doing any of the extra bookkeeping that we don't need to be keeping track of anyway. No measurable slowdown and we now support all indexing patterns that sparse arrays support. :)

bhancock8:
All tests passing and no observed slowdowns. Merging it in.

PR #759, opened 2017-08-04
Annotated Images with person-bike pair

paroma:
Changes to csv generator and image_generator notebook for (person,bike) candidate type


bhancock8:
LGTM

PR #773, opened 2017-08-10
Babbler runs start to finish

paroma:
- most files changed in an unclean manner, shouldn't merge this :)


bhancock8:
It was helpful for comparison. :)

PR #775, opened 2017-08-11
Re-constructs model for each hyperparameter config in grid search

ajratner:


ajratner:
Gah now a timeout bug!  Will try to cut down the test length because this keeps happening with all the PRs...

PR #777, opened 2017-08-12
Refactor GridSearch + TFNoiseAwareModel to be fully seeded

ajratner:
Refactored `GridSearch` to simplify, then ironed out the seeding for both `GridSearch` and `TFNoiseAwareModel`, using `RandomState`s to keep contained within class.  Tested deterministic execution locally


ajratner:
@stephenbach or @jason-fries this is ready to merge in!  Everything should be deterministic end-to-end now.  I already pushed changes to `scuba/master` that sync it with this branch and am testing / re-running all results overnight to confirm that everything is deterministic in full scuba pipeline

ajratner:
@jason-fries good catch!  Just pushed the fix.

@stephenbach (and @jason-fries if you have an opinion here): In `GenerativeModel` we set the random seed at construction; in `TFNoiseAwareModel` currently, the seeds are set (for both the `numpy.RandomState` instance and the TF graph) when `train` is called, since currently that's when the graph is constructed.

I could try to change this so that one single graph is constructed at construction... but then this leads to issues if anyone calls `train` again.  Which is why I prefer it this way.  But any objections?  If we want to test for e.g. variance, we can just start an unseeded instance?

ajratner:
Also @jason-fries @stephenbach FYI somehow I am still not getting deterministic execution of a full scuba pipeline, but will get to bottom of & fix today

ajratner:
@jason-fries Note that I changed the interface so you specify `save_dir` in the init of `GridSearch`, and it automatically appends it to `save_dir/grid_search`.  Happy to change back, this was just to simplify for scuba

PR #782, opened 2017-08-15
Fix for structure learning with propensity factor.

stephenbach:
Random segfaults are the rare type of bug that isn't a form
of regularization.


ajratner:
LGTM :D!

PR #783, opened 2017-08-17
Add option to remove reduce ops

ajratner:
This makes `LogisticRegression` and `SparseLogisticRegression` deterministic on GPU; the LSTMs unfortunately not but appears to reduce inter-run variance on GPU...


ajratner:
Should not change any behavior with current default of `deterministic=False`

PR #784, opened 2017-08-17
Data generation fix.

stephenbach:


ajratner:
LGTM!  Also- this is just part of synthetic testing, not mainline code right?  So fine if you just want to merge in I think?  Either way gets my stamp!

PR #789, opened 2017-08-23
Quick fix for memory efficient loading

danich1:
Greetings this PR is a patch for loading the annotation matrix in a memory efficiently. While using snorkel, I ran into a memory issue where sqlalchemy tried to read the entire feature table before constructing the lil matrix. This might not sound like a problem, but can be detrimental for a table of about >21M rows. As a result I revamped the algorithm to avoid reading the entire table in memory. Plus  [stack overflow](https://stackoverflow.com/questions/32743584/python-lil-matrix-vs-csr-matrix-in-extremely-large-sparse-matrices) advised that I change the lil matrix construction to a coo matrix to avoid possible memory issues in the future. Free to change this to a way you all prefer. Let me know.

### DIFF
--- a/snorkel/annotations.py
+++ b/snorkel/annotations.py
@@ -355,22 +355,31 @@ def load_matrix(matrix_class, annotation_key_class, annotation_class, session,
             kid_to_col[kid] = j
             col_to_kid[j]   = kid
 
-    # Create sparse matrix in LIL format for incremental construction
-    X = sparse.lil_matrix((len(cid_to_row), len(kid_to_col)), dtype=np.int64)
+    # Create sparse matrix in COO format for incremental construction
+    row = []
+    columns = []
+    data = []
 
     # NOTE: This is much faster as it allows us to skip the above join (which for some reason is
     # unreasonably slow) by relying on our symbol tables from above; however this will get slower with
     # The total number of annotations in DB which is weird behavior...
     q = session.query(annotation_class.candidate_id, annotation_class.key_id, annotation_class.value)
-    q = q.order_by(annotation_class.candidate_id)
 
     # Iteratively construct row index and output sparse matrix
-    for cid, kid, val in q.all():
-        if cid in cid_to_row and kid in kid_to_col:
-            # Optionally restricts val range to {0,1}, mapping -1 -> 0
-            if zero_one:
-                val = 1 if val == 1 else 0
-            X[cid_to_row[cid], kid_to_col[kid]] = int(val)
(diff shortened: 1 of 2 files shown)


snorkel/annotations.py
```diff
@@ -355,22 +355,31 @@ def load_matrix(matrix_class, annotation_key_class, annotation_class, session,
...
+    # Create sparse matrix in COO format for incremental construction
+    row = []
+    columns = []
+    data = []
 
     # NOTE: This is much faster as it allows us to skip the above join (which for some reason is
     # unreasonably slow) by relying on our symbol tables from above; however this will get slower with
     # The total number of annotations in DB which is weird behavior...
     q = session.query(annotation_class.candidate_id, annotation_class.key_id, annotation_class.value)
-    q = q.order_by(annotation_class.candidate_id)
```
ajratner: Can we make sure the matrix is ordered by candidate_id somewhere else though?  Maybe just tack this on to the query below?

ajratner: Or make sure the cid_to_row keys are properly sorted

snorkel/annotations.py
```diff
@@ -355,22 +355,31 @@ def load_matrix(matrix_class, annotation_key_class, annotation_class, session,
...
-            # Optionally restricts val range to {0,1}, mapping -1 -> 0
-            if zero_one:
-                val = 1 if val == 1 else 0
-            X[cid_to_row[cid], kid_to_col[kid]] = int(val)
+    # Cycle through each key and rely on select rather than going through the entire table
+    for f_cid in cid_to_row.keys():
+
+        for cid, kid, val in q.filter(annotation_class.candidate_id == f_cid).all():
+
+            if cid in cid_to_row and kid in kid_to_col:
```
ajratner: You don't need to check if `cid` is in `cid_to_row` here, correct?

danich1:
Greetings Alex,

Took some pondering, but I believe that I have found a better alternative towards my previous fix. I'm relying on sqlalchemy's core to efficiently parse through the table. This takes up some memory but ***way*** less than the ORM was trying to allocate. I tested it on my code and then the tutorial and there are major speed increases. ~114 ms for tutorial. After travis ci checks my code, please take a look and let me know if this method suffices. 

~David

ajratner:
@danich1 Sorry for the long lag time here- this looks great though!!  Merging in

ajratner:
Confirmed the speed-up too btw- this is great work, thanks!

PR #794, opened 2017-08-31
starting image > drink task

paroma:
lots going on:
- added drink task so the BabbleModel notebook runs end to end
- created drink folder with copies of bike stuff (notebooks were me testing, and drink_model looks at the relevant .npy files)
- changeImageExtractor because it was specific to the bike task, added an argument that defaults to bike task to not break existing scripts
- changed post-processing a bit to account for bad candidates and not throw error
- current Drink task is the"able to reach cup" query, though we have necessary files for the "holding a cup" query as well


tutorials/babble/drink/image_objects.py
```diff
@@ -0,0 +1,65 @@
+import numpy as np
```
bhancock8: Why are we reintroducing Image_Candidate and BBox objects. Do the ones in context.py that we use for the bike task not work for the drink task?

paroma: they should, i just copied most of the files that existed in the bike/ folder so we shouldn't include this second file

tutorials/babble/drink/__init__.py
```diff
@@ -0,0 +1,2 @@
+from drink_model import DrinkModel
+from image_objects import BBox
```
bhancock8: We shouldn't use this BBox object anywhere.

paroma: removing stuff!

paroma:
removed outdated files

PR #796, opened 2017-09-04
Fix link to lecture slides in README.md

vochicong:


ajratner:
@vochicong thanks so much for this!!

PR #804, opened 2017-09-20
Update to debug categorical candidates.

regoldman:
Addresses Issues 799 and 800.

### DIFF
--- a/snorkel/models/candidate.py
+++ b/snorkel/models/candidate.py
@@ -99,8 +99,8 @@ def candidate_subclass(class_name, args, table_name=None, cardinality=None,
             raise ValueError("Number of values must match cardinality.")
         if None in values:
             raise ValueError("`None` is a protected value.")
-        if any([type(v) == int and v == 0 for v in values]):
-            raise ValueError("0 is a protected value")
+        if any([type(v) == int for v in values]):
+            raise ValueError("Default usage of values is consecutive integers. Leave values unset if attempting to define values as integers.")
         cardinality = len(values)
 
     # If cardinality is specified but not values, fill in with ints
--- a/snorkel/models/candidate.py
+++ b/snorkel/models/candidate.py
@@ -99,8 +99,8 @@ def candidate_subclass(class_name, args, table_name=None, cardinality=None,
             raise ValueError("Number of values must match cardinality.")
         if None in values:
             raise ValueError("`None` is a protected value.")
-        if any([type(v) == int for v in values]):
-            raise ValueError("Values cannot be integers.")
+        if any([v == 0 for v in values]):
+            raise ValueError("0 is a protected value")
         cardinality = len(values)
 
     # If cardinality is specified but not values, fill in with ints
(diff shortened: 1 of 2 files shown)


snorkel/models/candidate.py
```diff
@@ -99,8 +99,8 @@ def candidate_subclass(class_name, args, table_name=None, cardinality=None,
             raise ValueError("Number of values must match cardinality.")
         if None in values:
             raise ValueError("`None` is a protected value.")
-        if any([type(v) == int for v in values]):
-            raise ValueError("Values cannot be integers.")
+        if any([type(v) == int and v == 0 for v in values]):
```
ajratner: So the intended usage was as follows: if you want integer-valued classes (e.g. `[1,2,3]`), you just leave `values` unset and then it defaults to that.

So technically, the only reason to specify integer values is if you want to specify a non-consecutive sequence of ints... but this will not be correctly handled in the current [code](https://github.com/HazyResearch/snorkel/blob/master/snorkel/annotations.py#L412). Overall, this is a minor issue; but either way options would be to:
* Just revert back to the previous code, but add a note to the error message saying to just leave `values` unset
* Check for non-consecutive or zero-valued ints, throw error then

snorkel/utils.py
```diff
@@ -75,25 +83,29 @@ def matrix_coverage(L):
     Given an N x M matrix where L_{i,j} is the label given by the jth LF to the ith candidate:
     Return the **fraction of candidates that each LF labels.**
     """
-    return np.ravel(sparse_abs(L).sum(axis=0) / float(L.shape[0]))
+    return np.ravel(sparse_nonzero(L).sum(axis=0) / float(L.shape[0]))
```
ajratner: Great catch here, thanks!!

snorkel/utils.py
```diff
@@ -55,6 +55,14 @@ def camel_to_under(name):
...
     return re.sub('([a-z0-9])([A-Z])', r'\1_\2', s1).lower()
 
+def sparse_nonzero(X):
+    """Sparse matrix with value 1 for i,jth entry !=0"""
+    X_nonzero = X.copy()
+    if sparse.isspmatrix_csr(X) or sparse.isspmatrix_csc(X):
+        X_nonzero.data[X_nonzero.data != 0] = 1
+    else:
+        raise ValueError("Only supports CSR/CSC matrices")
+    return X_nonzero
```
ajratner: Can you quickly add in the rest of the support just mirroring the below code (for dense and LIL matrices)?  This will be:
- Dense: `X_nonzero[X_nonzero != 0] = 1`
- LIL: `X_nonzero.data = [np.ones(len(L)) for L in X_nonzero.data]`

snorkel/utils.py
```diff
@@ -75,25 +83,29 @@ def matrix_coverage(L):
...
     Return the **fraction of candidates that each LF _conflicts with other LFs on_.**
     """
-    L_abs = sparse_abs(L)
-    return np.ravel(np.where(L_abs.sum(axis=1) != sparse_abs(L.sum(axis=1)), 1, 0).T * L_abs / float(L.shape[0]))
+    B = L.copy()
+    for row in range(0,B.shape[0]):
+        if np.unique(B.getrow(row).data).size == 1:
+            B.data[B.indptr[row]:B.indptr[row+1]] = 0
+    return matrix_coverage(sparse_nonzero(B))
+
```
ajratner: Is this slow because of the looping over rows of B?  Probably should be fine... can come back to this I guess, happy to test later

ajratner:
@regoldman Looks great!  Thanks so much and sorry for the delay!

PR #819, opened 2017-10-25
Update typos in Categorical_Classes.ipynb

chentc:
Since there are more relationships involved, consider renaming the variable name.


ajratner:
Great catch :) . Thanks!!  Merging in now

PR #825, opened 2017-11-02
matching traditional and weak train numbers

paroma:
we can't use the `traditional_supervision` function in snorkel pipeline for images...because of the candidate and image mismatch i believe

also a weird issue where the marginals are sometimes floats and sometimes sparse matrices, that was making it skip some images as well

in the notebook (not pushed), I'm hard coding`config['max_train'] = 15711` in order to make sure all the images get picked up, this is not necessary anymore since i don't even call the `traditional_supervision` function

### DIFF
--- a/snorkel/contrib/babble/pipelines/image_pipeline.py
+++ b/snorkel/contrib/babble/pipelines/image_pipeline.py
@@ -32,8 +32,7 @@ class ImagePipeline(BabblePipeline):
             num_candidates = self.session.query(self.candidate_class).filter(
                 self.candidate_class.split == split).count()
             print("Candidates [Split {}]: {}".format(split, num_candidates))
-
-
+    
     def classify(self, config=None, slim_ws_path=None):
         if config:
             self.config = config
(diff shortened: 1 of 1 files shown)


snorkel/contrib/babble/pipelines/image_pipeline.py
```diff
@@ -138,8 +142,8 @@ def scrape_output(output_file):
         if self.config['supervision'] == 'traditional':
             print("In 'traditional' supervision mode...grabbing candidate and gold label subsets.")  
             candidates_train = self.get_candidates(TRAIN)
-            L_gold_train = load_gold_labels(self.session, annotator_name='gold', split=TRAIN)
-            X_train, Y_train = self.traditional_supervision(candidates_train, L_gold_train)
+            Y_train = load_gold_labels(self.session, annotator_name='gold', split=TRAIN)
+            #Deleted call to traditional_supervision, which was pruning by number of non-zeros
```
paroma: this is where the magic happened (also I wanted to use this feature)

snorkel/contrib/babble/pipelines/image_pipeline.py
```diff
@@ -77,10 +76,15 @@ def link_images_candidates(anns, candidates, mscoco, marginals):
...
                 coco_ids[image_id] = int(mscoco_id)
+                
+                #HACK: sometimes marginals[idx] is a float, sometimes a matrix...
                 try:
                     labels[image_id] = max(labels[image_id], max(marginals[idx], 0))
                 except:
-                    import pdb; pdb.set_trace()
+                    try:
+                        labels[image_id] = max(labels[image_id], max(marginals[idx].todense(), 0))
+                    except:
```
paroma: this is for the marginals being matrices and floats interchangeably

PR #834, opened 2017-11-20
fix spacy version to 1.10.0

yayitswei:
since we're not supporting spacy 2.0 yet


ajratner:
@yayitswei this is great, thanks so much!!! 🥇

PR #839, opened 2017-11-28
add figure extraction support in fonduer

SenWu:


snorkel/contrib/fonduer/fonduer/models/context.py
```diff
@@ -521,3 +548,88 @@ def __ne__(self, other):
...
+            % (self.__class__.__name__, self.figure.document.name.encode('utf-8'), \
+                self.figure.position, self.figure.url)
+
+    def _get_instance(self, **kwargs):
+        return TemporaryImage(**kwargs)
+
+
+class Image(Context, TemporaryImage):
+    """
+    A span of characters, identified by Context id and character-index start, end (inclusive).
```
lukehsiao: An image is a "span of characters"? 

We should clarify the comments so that it's clear how this is different that just a Span?

snorkel/contrib/fonduer/fonduer/models/context.py
```diff
@@ -58,6 +60,31 @@ def __gt__(self, other):
...
+    __mapper_args__ = {
+        'polymorphic_identity': 'figure',
+    }
+
+    __table_args__ = (
+        UniqueConstraint(document_id, position),
+    )
+
+    def __repr__(self):
+        return "Figure(Doc: %s, Position: %s)" % (self.document.name.encode('utf-8'), self.position)
```
lukehsiao: Do we want to print the URL here too, like we do for the Image?

snorkel/contrib/fonduer/fonduer/parser.py
```diff
@@ -247,18 +252,22 @@ def parse_node(node, table_info=None):
 
             for child in node:
                 if child.tag == 'table':
-                    parse_node(child, TableInfo(document=table_info.document))
+                    parse_node(child, TableInfo(document=table_info.document), figure_info)
+                elif child.tag == '':
+                    parse_node(child, table_info, FigureInfo(document=figure_info.document))
```
lukehsiao: Checking if a tag is empty to imply that we need a FigureInfo seems a little confusing, is a Figure really the only scenario where the tag is empty?

snorkel/contrib/fonduer/fonduer/parser.py
```diff
@@ -135,7 +135,7 @@ def __init__(self,
         for (pattern, replace) in replacements:
             self.replacements.append((re.compile(pattern, flags=re.UNICODE), replace))
         if self.lingual:
-            self.batch_size = 7000  # TODO(bhancock8): what if this is smaller than a block?
+            self.batch_size = 7000
```
lukehsiao: Rather than just get rid of this comment, it would probably be good to leave a comment for why `7000` is a magic number here.

SenWu: Yes, that's the limit for CoreNLP

PR #849, opened 2017-12-27
Move _preprocess_data to parent Classifier class

ajratner:
This is a simple fix to a current bug in running multi-threaded `GridSearch` with models that don't have a `_preprocess_data` method, e.g. the `GenerativeModel` (which causes errors running scuba the whole way through with `n_threads > 1` for example).

Note that calling `._preprocess_data` in the grid search to pre-load data isn't the cleanest way to go about things but leave for now...


ajratner:
Note: I don't know for sure what's going on with Travis, but I think it's just a caching issue since I've been switching around miniconda versions on another PR... maybe?  Will figure out later

PR #852, opened 2018-01-11
Flag for non-mutually-exclusive categorical classification

ajratner:
Now, if `Classifier` subclass is initialized with cardinality > 2 and `single_value=True` (default), it will have the same behavior as before: the marginals will be the result of applying a softmax, i.e. the probabilities for all classes will sum to one for each datapoint.  This will also be used during training.

However, now if `single_value=False`, a sigmoid will be applied element-wise instead--i.e. classes are not mutually-exclusive; this will also be used during training.  @regoldman this should be the quick fix for your application!

With `single_value=False`, it will *almost* be equivalent to just training `k` separate binary classifiers... almost because the LFs still only give mutually-exclusive labels.  So e.g. if there are LFs that should say 'yes' to two different classes, they would currently need to be represented as two different LFs, which would presumably change the generative model... so we'll have to think about this more.  Either way this flag shouldn't hurt or change anything when left at it's default

### DIFF
--- a/snorkel/learning/classifier.py
+++ b/snorkel/learning/classifier.py
@@ -17,9 +17,13 @@ class Classifier(object):
     # otherwise assume X is an AnnotationMatrix
     representation = False
 
-    def __init__(self, cardinality=2, name=None):
+    def __init__(self, cardinality=2, name=None, single_value=True):
         self.name = name or self.__class__.__name__
         self.cardinality = cardinality
+
+        # Determine whether predicts single value or (potentially) multiple
+        # Only valid when cardinality > 2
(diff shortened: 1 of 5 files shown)


ajratner:
Also @regoldman : there is a quicker way to implement this that doesn't require re-constructing (i.e. re-training) the models... let me know if this would be more convenient (just returning the unscaled logits instead of `marginals` and thresholding at 0 for each class)

ajratner:
This has other issues--thanks to discussion with @stephenbach!  In particular the semantics of negative votes are very unclear... going to close this for now

PR #856, opened 2018-01-16
WIP: Comparing changes for new fonduer_parser.

lukehsiao:
Do **NOT** merge this PR. Just for comparison purposes right now.


snorkel/contrib/fonduer/phantomjs/examples/arguments.js
```diff
@@ -0,0 +1,10 @@
+"use strict";
```
lukehsiao: Remove phantomjs. We have the install script, so this whole program shouldn't be versioned in this repo.

snorkel/contrib/fonduer/fonduer/parser.py
```diff
@@ -67,14 +70,14 @@ def __init__(self,
                  flatten_delim='',
                  lingual=True,                       # lingual
                  strip=True,
-                 replacements=[(u'[\u2010\u2011\u2012\u2013\u2014\u2212\uf02d]', '-')],
+                 replacements=[],#[(u'[\u2010\u2011\u2012\u2013\u2014\u2212\uf02d]', '-')],
```
lukehsiao: Will users be able to still easily look for dashes in their LFs, rather than needing to worry about all the different types (e.g. endash, emdash, subtract, etc.)? It's fine if we move that responsibility to the user, just curious about the motivation for the change.

snorkel/contrib/fonduer/fonduer/parser.py
```diff
@@ -162,7 +165,7 @@ def apply(self, x, **kwargs):
                 pass
             # Add visual attributes
             filename = self.pdf_path + document.name
-            create_pdf = not os.path.isfile(filename + '.pdf') and not os.path.isfile(filename + '.PDF')
+            create_pdf = not os.path.isfile(filename + '.pdf') and not os.path.isfile(filename + '.PDF') and not os.path.isfile(filename)
```
lukehsiao: Do we have PDF files with no extension? Seems strange to assume something is a PDF without an extension.

tutorials/fonduer/hardware/test_parser.py
```diff
@@ -0,0 +1,32 @@
...
+
+corpus_parser = OmniParser(structural=True, lingual=True, visual=False, pdf_path=pdf_path)
+corpus_parser.apply(doc_preprocessor, parallelism=PARALLEL)
+
+from snorkel.contrib.fonduer.models import Document, Phrase, Figure, Para
+
+print "Documents:", session.query(Document).count()
+print "Phrases:", session.query(Phrase).count()
+print "Paragraphs:", session.query(Para).count()
+print "Figures:", session.query(Figure).count()
```
lukehsiao: It would be very helpful as tests if the expected values were included here. Just printing these numbers alone doesn't really tell me much about what is working, and what isn't. The numbers I'm getting are:

```
Documents: 4
Phrases: 5978
Paragraphs: 3727
Figures: 47
```

which I assume is wrong, since there 7 documents in the directory, and `max_docs` is set to 6.

lukehsiao: It would be very helpful as tests if the expected values were included here. Just printing these numbers alone doesn't really tell me much about what is working, and what isn't. The numbers I'm getting are:

```
Documents: 4
Phrases: 5978
Paragraphs: 3727
Figures: 47
```

which I assume is wrong, since there 7 documents in the directory, and `max_docs` is set to 6.

PR #900, opened 2018-03-30
FIX restrict TemporarySpan.__contains__() to same sentence

moreymat:
This PR prevents `TemporarySpan.__contains__()` from returning `True` for spans from different Sentence contexts.
This condition is already enforced elsewhere in this class, in particular in the related `__eq__` and `__ne__`.


stephenbach:
Good catch. Thank you!

PR #901, opened 2018-03-30
FIX make brat.utils compatible with python 3 (and 2.6+)

moreymat:
This PR makes `brat.utils` compatible with python 3 (and 2.6+).
It enforces the modern syntax for exceptions and uses six for urllib.


stephenbach:
Thank you @moreymat !

moreymat:
You're welcome, thanks for this great library!

Issue #877, opened 2018-02-24
About Discriminative Model Loss Function（Bug？？？）


wenfeixiang1991:
**There are some knowledge about discriminative model loss function in Snorkel Project.**

1. [Snorkel: Rapid Training Data Creation with Weak Supervision (VLDB 2018)](https://arxiv.org/abs/1711.10160).It mainly saids two points about loss function. 
The first one just give a discretionally formula expression which I could know that 

> we train a discriminative model hθ on our probabilistic labels Y ̃ by minimizing a noise-aware variant of the loss l(hθ(xi),y), i.e., the expected loss with respect to Y ̃

   The second one give a words description.

> To best take advantage of this, Snorkel creates probabilistic training labels for any discrim- inative model with a standard loss function.
2. Snorkel code [disc_learning.py](https://github.com/HazyResearch/snorkel/blob/0b76242bb76098967b736b15aa699b32979bdc6b/snorkel/learning/disc_learning.py#L71) just uses `tf.nn.sigmoid_cross_entropy_with_logits` and I don't see any changes of loss function.

**So, here are my questions** 

1. What does 'standard loss function' in [Snorkel: Rapid Training Data Creation with Weak Supervision (VLDB 2018)](https://arxiv.org/abs/1711.10160) mean?
2. Is that snorkel code [disc_learning.py](https://github.com/HazyResearch/snorkel/blob/0b76242bb76098967b736b15aa699b32979bdc6b/snorkel/learning/disc_learning.py#L71) `tf.nn.sigmoid_cross_entropy_with_logits` right? Because in my point of view, cross entropy function is `Loss = -1/n * sum(y~*ln(y) + (1-y~)*ln(1-y))` which is only effective when y~ is 0 or 1, if y~ is some probabilistic number between 0 and 1, I can't see it works directly. But `1/2*(y - y~)^2` does make sense when y~ is probabilistic label.
3. If I need change loss function to match discriminative model in snorkel, how would I do that? 
4. The reason I post this issue is that the loss is not decrease when I use snorkel code re_rnn.py, which is #870 . Then I check every mistakes I probably make, nothing found but one point I notice that most my input_y is around 0.5. So I thought this could cause loss not decrease, but in snorkel theory, y always approach y~, and I didn't see that could be happen when the code is `tf.nn.sigmoid_cross_entropy_with_logits`
[...]

PR #910, opened 2018-04-06
Quick Fix for Bug in the RNN code

danich1:
Greetings everyone,

When I was working on the latest commit of snorkel, I noticed that the RNN wasn't performing dropout as one would hope. After digging into the code, I found that the dropout part was calculated; however, the variable wasn't used at all (thanks to #877). So I'm proposing a quick fix to this issue.

Thanks,
~David


stephenbach:
I'd like to get this in still on 0.7.0-alpha. Not sure if rebasing is best? Let me know if I should or if just want to submit a new PR.

danich1:
Sorry for the delay. Since this is a quick fix, let me resubmit the PR. Feel free to close this one and look for the new one shortly.

Issue #724, opened 2017-07-20
Switch default parser to Spacy?


ajratner:

jason-fries:
Also -- remove CoreNLP download from run.sh?

jason-fries:
Maybe we should have a download option like with spaCy that's internal to the CoreNLP class. It would then install the first time it's actually run?

ajratner:
Yeah def

moreymat:
Dear developers,
I am currently trying to use Snorkel in a project on French data.
Even though I really like spaCy a lot, its support for languages other than English is currently lagging behind that of CoreNLP (version 3.7.0 and later).
By the way, if you are interested in adding a (very basic) support for other languages to your wrapper around CoreNLP, I could prepare a small PR based on my own changes for French.
Best,

jason-fries:
Hi @moreymat 
I know spaCy has models for English, French, German, and Spanish but I sure the non-English models leave a lot to be desired. 

For the user, it's really simple to switch parsers -- we're still including CoreNLP support. Any enhancements you have for CoreNLP and other language support would be awesome to include. Definitely prepare a PR request and we can take a look. Thanks!

PR #876, opened 2018-02-24
Spacy 2.0 Compatibility

4d4stra:
There are a few small changes in this pull request.  
1.  A minor bug fix that resulted in a failure to download the spacy language model if it doesn't exist.
2. Some extraneous code was removed in the code used to test
3. This update checks for the spacy version (1.x or 2.x) and updates the api calls as appropriate.

The changes were tested with both spacy version 1 and 2 using the tutorial code.  I have also tested using spacy 2.0 more thoroughly at work.

PR #924, opened 2018-05-04
Switch spacy installation to conda from pip

dhimmel:
Hopefully this will reduce the 550 second [install time](https://travis-ci.org/HazyResearch/snorkel/jobs/374664220#L1708) of the conda environment. It [looks like](https://travis-ci.org/HazyResearch/snorkel/jobs/374670307#L566) it reduces install time to 330 seconds.

Refs https://github.com/HazyResearch/snorkel/pull/876
Closes https://github.com/HazyResearch/snorkel/issues/724 (which actually has already been addressed).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -19,7 +19,6 @@ before_install:
     --output-document=miniconda.sh
   - bash miniconda.sh -b -p $HOME/miniconda
   - source $HOME/miniconda/etc/profile.d/conda.sh
-  - hash -r
   - conda config --set always_yes yes --set changeps1 no
   - conda info --all
 
@@ -28,6 +27,7 @@ install:
   - conda env create --quiet --file=environment.yml
   - conda activate snorkel
   - pip install .
+  - conda list
 
 script:
 
--- a/.travis.yml
+++ b/.travis.yml
@@ -39,19 +39,18 @@ script:
   - runipy test/learning/test_parallel_grid_search.ipynb
 
   # Runs intro tutorial notebooks
-  - cd tutorials
-  - runipy intro/Intro_Tutorial_1.ipynb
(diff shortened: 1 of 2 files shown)


dhimmel:
See build to confirm that spacy 2 is being installed:

+ https://travis-ci.org/HazyResearch/snorkel/jobs/374670307#L742
+ https://travis-ci.org/HazyResearch/snorkel/jobs/374670308#L730

dhimmel:
I narrowed down the issue. The following cell is going to take about 16 minutes to run on my device (it's about 10 seconds for every percentage point):

screenshot from 2018-05-03 20-30-18

One solution which is not ideal is [to use](https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received) `travis_wait`. However, we don't want build times to slow down, so it may be worth looking into why this command is now slower and think about whether we should address that.

stephenbach:
One other thing: it should only process 500 docs on travis, based on the variable set in the first cell. Perhaps it needs to be lowered further? But annoying if spacy 2 is actually the issue.

dhimmel:
Downgrading to version 1 of spacy fixed the slowness issue (see https://github.com/HazyResearch/snorkel/pull/924/commits/c697d24dbea6f453820683cd50ded05b84e35f21). This PR is still productive because it switches the spacy installation to use conda from pip, which prevent installation errors on my system and is faster. So we should consider merging... if so I'd rename to "Switch spacy installation to conda from pip" and use a squash merge so we don't merge in the intermediate commits.

Alternatively, we could look a bit more into speeding up spacy 2. Possibly related to https://github.com/explosion/spaCy/issues/1508

PR #931, opened 2018-05-11
 Quick fix so tensorflow rnn will use dropout

danich1:
This PR is designed to update 0.7.0-alpha, with the dropout fix for tensorflow. Since the old PR is pointing to an outdated branch, I am re-opening the request here. The original PR can still be found at this [link](#910). 

~David


stephenbach:
Locally, looks like turning down dropout to .1 makes the test pass (dropout was too high for desired accuracy because... it didn't do anything). So merging, then will update test. Thanks!

PR #923, opened 2018-05-03
Version 0.7

stephenbach:
A PR for collecting sub PRs, discussion, etc. (not ready for review yet)

Goals for this version:
Pytorch classifiers
New installation model that uses conda for everything and is installable from pip
Switch to Spacy parser as default and support Spacy 2

PR #933, opened 2018-05-15
Minor updates,mostly for large datasets

MatthieudeR:
Changes include: 
* Added filtered candidate  extraction (only extract candidate from documents that pass through a filter)
* Added options for safer (but slower) candidate extraction and adding into the db
* Added support for different optimizer in the config
* Added option to run corpus extraction in batches (Only in single-threaded mode)
* Added support for LR schedule 
* Added support for model selection based on F-beta score


dhimmel:
@stephenbach here we have a major PR being opened against the `master` branch. I would not recommend merging it before #923, since it could cause conflicts and we really want new work being tested against the new installation framework.

What is the timeline for the merging of the `0.7.0-alpha` branch?

MatthieudeR:
Rebased on v0.7 in #937

PR #937, opened 2018-05-23
Minor Edits, mostly for big datasets (  #933 rebased for v 0.7)

MatthieudeR:
Changes include:

* Added filtered candidate extraction (only extract candidate from documents that pass through a filter)
* Added options for safer (but slower) candidate extraction and adding into the db
* Added support for different optimizer in the config
* Added option to run corpus extraction in batches (Only in single-threaded mode)
* Added support for LR schedule
* Added support for model selection based on F-beta  score
#933


ajratner:
Hey @MatthieudeR - thanks for submitting this, and sorry for the delay in getting to it!

Unfortunately, this is now a bit old and the build appears to be broken. In addition, there are several distinct changes in here that would each need to be documented and cleaned up for merging into the main branch.  I'm going to close this PR for now, but if you want to do the above cleanup & get it passing, feel free to re-open!  Thanks!

PR #938, opened 2018-05-25
tutorial with coral and snorkel

paroma:
coral tutorial works with snorkel with no dependencies!


ajratner:
@vincentschen thanks!  Closing now

PR #976, opened 2018-07-06
Add indices to foreign keys by default

MyHumbleSelf:
This speeds up cascade deletes a lot


MyHumbleSelf:
Yes, good point, @stephenbach! Candidate deletion seems to benefit from this as well. SQLAlchemy may create these automatically for some database backends, but definitely not for Postgres, so forcing them with `index=True` is still a good, I think.

stephenbach:
Yay! Thank you!

PR #982, opened 2018-07-16
refactor/reword image tutorial for clarity

vincentschen:
* I agree with the calls for @inafiterau 's call for better software engineering practices to avoid having to change primtiives/LFs multiple times, but abstraction takes away from the clarity of what's going on in this case. Hopefully, people can see exactly what's going on in the tutuorial and build their own classes for primitives/LFs for their purposes!


vincentschen:
@ajratner for review!

PR #989, opened 2018-07-25
Pip

stephenbach:
Note the caveats added at the top of setup.py. Packages like PyTorch and Numba recommend installation via Conda. On my Mac, with this PR, I can install Snorkel in an empty environment using just `pip install .` from inside the repo. This working depends on PyPI having the right binaries for torch and llvmlite. In particular, based on PyTorch's website, I think it won't work on Linux without manually installing PyTorch separately.

I therefore have removed "and pip" from the 0.7 release notes to clarify that the Conda installation is the recommended way of installing Snorkel, but the comments in setup.py point out users can also try just pip and how to get around common errors.

I will also try uploading Snorkel to PyPI now.

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,7 +92,7 @@ You can also check out all the great [materials](https://simtk.org/frs/?group_id
 
 ### Major changes in v0.7:
 * [PyTorch](https://pytorch.org/) classifiers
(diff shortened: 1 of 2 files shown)


ajratner:
@stephenbach any updates here?  Should this remain open?

PR #1000, opened 2018-08-03
Add multi-process UDF progress

CatalinVoss:
Keeping track of multiprocess progress while doc parsing, etc. is a pretty crucial ingredient for our application. This updates the progress bar to tqdm and changes a bunch of stuff in UDF to enable multi-process progress tracking. It is a somewhat imperfect solution, but perhaps not less imperfect than what was there before.

The `QUEUE_TIMEOUT` thing still bothers me and a much better approach may be to switch to the concurrent futures paradigm https://docs.python.org/3/library/concurrent.futures.html. Let me know what you think and feel free to edit as you see fit. It's a pretty heavy change as far as I understand, so I may be breaking stuff.


CatalinVoss:
Don't think this is this our error? ```The job exceeded the maximum time limit for jobs, and has been terminated.
```

**Comment by CatalinVoss** [comment]
These last two commits by @debnil are not really associated with this PR, but unfortunately I made the PR from our fork's master, and we added them there, so if you don't like the changes, feel free to undo. Those changes to candidates are needed for us to do multiple, independent feature extraction on the same DB. Recreating the DB for each candidate type would not be possible for us given the size.

**Comment by CatalinVoss** [comment]
Continuing this conversation, I reverted the commit and am moving us onto a separate branch again. Let's discuss if/how we want to merge other stuff in (including the subclasses in our repo). Thanks!

File: snorkel/udf.py [hunk 1]
```diff
@@ -92,54 +90,64 @@ def apply_mt(self, xs, parallelism, **kwargs):
...
-                    try:
-                        y = out_queue.get(True, QUEUE_TIMEOUT)
-                        self.reducer.reduce(y, **kwargs)
-                        out_queue.task_done()
-                    except Empty:
-                        break
-                self.reducer.session.commit()
-            self.reducer.session.close()
+        while any([udf.is_alive() for udf in self.udfs]) and count < total_count:
+            y = out_queue.get()
```
ajratner: What's the reason behind getting rid of `QUEUE_TIMEOUT` here?  If no specific one, let's at least leave in as a kwarg?

CatalinVoss: In the case where there is a reduce step, the timeout places an arbitrary restriction/estimate on the completion time of the user-defined function (which in our case, was too low). If the UDF isn't fast enough to fill up the out_queue, this will just busy-loop at which point, why not leave that up to the queue to decide… seems like `get()` is cleaner. Let me know what you think.

ajratner: No makes sense- so all good by me then!

CatalinVoss:
This failure is `The job exceeded the maximum time limit for jobs, and has been terminated.` again… 02ed0b4 is a two-character comment change…

PR #1005, opened 2018-08-08
Allow Candidates to be arbitrary n-tuples

MaxG87:
Due to a hidden assumption in the code candidates had to consist of two entries. This made it impossible to adapt the intro tutorial to NER or relation extraction for more than two entries.

This merge provides (1) minor improvements in the edited file and (2) a fix to allow arbitrary n-tuples. The code is already used for initial NER experiments successfully.

### DIFF
--- a/snorkel/learning/pytorch/rnn/rnn_base.py
+++ b/snorkel/learning/pytorch/rnn/rnn_base.py
@@ -56,16 +56,16 @@ class RNNBase(TorchNoiseAwareModel):
         outputs = torch.Tensor([])
 
         for batch in range(0, n, batch_size):
+            cur_minibatch = X[batch:batch+batch_size]
 
-            if batch_size > len(X[batch:batch+batch_size]):
-                batch_size = len(X[batch:batch+batch_size])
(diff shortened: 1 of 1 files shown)


MaxG87:
I noticed that a check failed but I am a bit puzzled about this. The failure is due to exhausted runtime but

  * the changes should not affect the runtime measurably
  * only the build for Python 2.7 fails

Is this serious? Are there hints how to investigate this, e.g. how to run the tests for Python 2.7 myself?

Vaikless:
Hello Max,

does your fix still work - and does it lead to 'good' results in a single value extraction task? 
For my bachelor's thesis I have a use-case where I need to extract a single value and only that value is labeld in my data. For example "revenue (...) 500"  500 is labeled and revenue is not.

I could try to programmatically label the span the labeled value is realted to - "revenue" in the example above.
Or I could try your fix :)

Thx and best regards from Bonn, Germany

MaxG87:
@ajratner What is missing for this PR to be accepted? I would like to get rid of this branch.

henryre:
This should be covered by #1066, closing out for now.

PR #1013, opened 2018-08-16
Candidate -> document shortcut

CatalinVoss:
This adds a `document` field to `Candidate` that is automatically populated during extraction by climbing up the context hierarchy until we arrive at the doc parent. This is useful for doc-level information tasks where our goal is to eventually reconstruct document-level labels, but we frame it as an extraction task over sub-doc-level candidates (Span, Sentence, etc.).

These commits also add an index on `Candidate.type` which speeds up stuff when running multiple extraction tasks with different Candidate-subclasses, which we'll reference in a later PR.


CatalinVoss:
Seems like tests rely on `spouse.db` and its schema. Would be a lot easier if that wasn't the case. This means we somehow have to figure out how to update the test schema every time we make a change that affects schema.

CatalinVoss:
Seems like tests rely on `spouses.db` and its schema. Would be a lot easier if that wasn't the case. This means we somehow have to figure out how to update the test schema every time we make a change that affects schema.

CatalinVoss:
@ajratner Is the code you used to generate `spouses.db` somewhere in there so we can just re-run that with the new stuff in SQLAlchemy tree and won't have to migrate the schema manually? I imagine it's just more or less the db generated by one of the tutorials?

ajratner:
Yeah it should just be the one from the intro tutorial!

CatalinVoss:
@ajratner I re-ran, Intro Tutorial 1 and 2, but the output `snorkel.db` is now 272.4 MB, so too big to put onto git. 😞 Looks like we also have to recreate `crowdsourcing.db` in the tensorflow dir. I don't have a pyspark setup on my laptop, so could you look into re-running that or we can look tomorrow?

ajratner:
@stephenbach any chance you could push the scripts you used to generate the test data?  @CatalinVoss

ajratner:
@CatalinVoss can you just try re-running on your fork?  I don't have pyspark currently installed either but is quick to install.  Then you can just filter the number of docs (for intro tutorial, `n_docs=500` or similar)?  Is there a reason this wouldn't work?

henryre:
Hi @CatalinVoss, we've moved the information extraction version of Snorkel (v0.7) to https://github.com/snorkel-team/snorkel-extraction. Feel free to reopen this PR there!

CatalinVoss:
We're no longer using snorkel. But happy to see the new stuff

PR #1014, opened 2018-08-16
Properly connect candidates and marginals

CatalinVoss:
Here, we actually add candidate as a relationship to marginal. Together with #1013, this is useful for our road back to doc labels). Any particular reason it wasn't there, or just not needed? It shouldn't add any other overhead, unless we're missing something.

This will enable you to clean up things in a few places in snorkel where you pass around tuples with marginal/candidate IDs rather than using objects.


ajratner:
LGTM, re-running tests that failed for a weird reason

PR #1019, opened 2018-08-20
Scorer cleanup

CatalinVoss:
Some scorer cleanup that should hopefully quick & easy to merge. FWIW, we're having serious trouble with the multiclass scoring, so we've redone this on our side and are refactoring.


ajratner:
@CatalinVoss thanks for this!  Re-running a test that timed out, then can merge

PR #1020, opened 2018-08-20
Fix bug in Ngram splitting bug

lukehsiao:
HiromuHota fixed this bug in Fonduer in https://github.com/HazyResearch/fonduer/pull/108 and https://github.com/HazyResearch/fonduer/pull/112. This commit fixes it for Snorkel.


ajratner:
Thanks!!  LGTM

PR #1024, opened 2018-08-27
Fix PyTorch install for Travis build

henryre:
PyTorch now installs GPU version by default. Switch install to CPU-only version.


ajratner:
LGTM once tests pass!

ajratner:
Tests are now failing on something else, merging this in first then addressing that

PR #1031, opened 2018-10-25
DB hierarchy refactor

CatalinVoss:
These changes refactor in our model hierarchy to eliminate `Context` as the in-between uber-class that everything inherits from. Rather than having corresponding entries in the `context` table and operating via joined table inheritance, `Sentence`, `Document`, `Span`, etc. all now have their own tables. This speeds up insertion and other queries significantly.

This means that `candidate_subclass()` now needs to keep track of the types of its args.

@ajratner This is somewhat invasive and related to our discussion. I find the code in `PretaggedCandidateExtractorUDF` pretty hard to follow, but this does pass all of our tests for labeling & generative model on our side now.

### DIFF
--- a/snorkel/contrib/snark/parser.py
+++ b/snorkel/contrib/snark/parser.py
@@ -31,7 +31,7 @@ class SparkCorpusParser(object):
 
     def _clear(self):
         # TODO
-        # session.query(Context).delete()
+        # session.query(Document).delete()
         # # We cannot cascade up from child contexts to parent Candidates,
         # # so we delete all Candidates too
         # session.query(Candidate).delete()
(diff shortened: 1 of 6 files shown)


ajratner:
Okay looked through- I totally get the motivation here, but definitely a big change :) . One route is that this is a separate "flat" DB structure so the user could choose this instead of the hierarchical one... will think about more tomorrow!

CatalinVoss:
OK, let's discuss on Monday. My guess is that most users may not want to choose this. I'll run some benchmarks on our end to quantify the performance gain a bit in the meantime.

CatalinVoss:
@ajratner Any thoughts on https://github.com/HazyResearch/snorkel/pull/1031/commits/32c5dd66e36d3b266f9e2170d79eb6278e0e1d7f ? Seems like this should ideally be a bit more asynchronous…

henryre:
Hi @CatalinVoss, we've moved the information extraction version of Snorkel (v0.7) to https://github.com/snorkel-team/snorkel-extraction. Feel free to reopen this PR there!

Issue #1030, opened 2018-10-18
Pandas version of crowdsourcing tutorial


tmerrittsmith:
I've edited the crowdsourcing tutorial to use pandas instead of spark. It was useful for me, so may be useful to others. Do you want it?

ajratner:
@tmerrittsmith yes!!  This would be awesome!  We definitely want to keep the spark one, but could have the pandas version as a separate notebook?  Thanks!!

tmerrittsmith:
Cool. Yeah the spark one is definitely worth keeping - it's like the Ferrari version, where mine is the second hand hatchback... 

I'll tidy up what I've got, but can provide two possibilities:
1) A straight conversion of the crowdsourcing tutorial (exactly the same data and results), just using pandas where spark was used (to be honest, I didn't use pandas that much apart from joining the csvs at the beginning)

2) Something more similar to the intro tutorial, but for tabular data. We wrote some labelling functions for a UCI repository dataset, and then used snorkel's generative model to resolve the labels.

What's the best way to get them to you, once they're ready?

ajratner:
Hah :) . Both of these sound awesome- whatever you can send, we'll look over!  And best way would be a PR (separate ones for each).  Thanks!!

tmerrittsmith:
Sorry, haven't got round to making the PR - I'll do that today.

ajratner:
@tmerrittsmith awesome thanks!!

hanzlfs:
@tmerrittsmith thanks! I think your PR is good enough to have insights.

ajratner:
@paroma assigned to you since you're looking at the PRs!

paroma:
thank you for submitting this PR! (merged with #1048)

gauthamkrishna-g:
@tmerrittsmith Could you kindly update on the tabular data PR in case you have any leads? Thanks!

PR #1048, opened 2019-01-18
Notebook removing the spark dependency in the crowdsourcing tutorial

tmerrittsmith:
Reference Issue: Pandas version of crowdsourcing tutorial #1030  
https://github.com/HazyResearch/snorkel/issues/1030


ajratner:
Hey @tmerrittsmith this seems to be failing on an unrelated issue that we're still dealing with.  Can you confirm your new notebook is working on your end?  If so, since purely additive, I can merge in.  And thanks for your work on this!!

PR #1053, opened 2019-02-05
Fix link to snorkel.stanford.edu in README

chrisji:
Previously points to https://github.com/HazyResearch/snorkel/blob/master/snorkel.stanford.edu


bhancock8:
Thank you for the fix!

PR #1054, opened 2019-02-06
Tiny update to spacy_parser

CatalinVoss:
Don't add null entity array twice...

We're converting this to `pipe()` method on our side for a pretty solid ~3x performance boost due to batching.

### DIFF
--- a/snorkel/parser/spacy_parser.py
+++ b/snorkel/parser/spacy_parser.py
@@ -160,10 +160,6 @@ class Spacy(Parser):
                 parts['dep_parents'].append(head_idx)
                 parts['dep_labels'].append(token.dep_)
 
-            # Add null entity array (matching null for CoreNLP)
-            parts['entity_cids'] = ['O' for _ in parts['words']]
-            parts['entity_types'] = ['O' for _ in parts['words']]
-
             # make char_offsets relative to start of sentence
             parts['char_offsets'] = [
                 p - parts['char_offsets'][0] for p in parts['char_offsets']


bhancock8:
@jason-fries do you have any sense of if the lines @CatalinVoss suggests removing are needed anywhere else or break anything else? I know the CIDs were introduced in part for some of our biomedical tasks.

bhancock8:
The travis build breaking, btw, is due to a change in one of our dependencies; old builds that have already passed before suddenly are not passing when you restart them. That's a separate issue that needs to be fixed.

CatalinVoss:
I only removed them bc they appeared twice, actually. Happy to share "piped" version if of interest, too… This makes a dramatic performance difference
```
    def parse_pipe(self, text_tuples):
        """
        Piped version of parse() enabling spacy minibatching for improved
        performance.

        Args:
            text_tuples: a generator of (text, context) tuples

        Yields:
            ([array of sentences contained in text], context) tuple for each
            text tuple
        """
        for doc, ctx in self.model.pipe(text_tuples, as_tuples=True):
            sentences = [self._construct_sentence_parts(sent)
                         for sent in doc.sents]
            yield sentences, ctx
```

bhancock8:
Ah, you're right! Great, merging this in then. And yes, the piped version seems well worth it. Is this snippet here all that changed? Could you submit it another PR?

CatalinVoss:
The trouble with the pipe is that you have to call into the parser with a
generator over docs. So I think it would require a bunch of changes in your
architecture. Happy to chat.

jason-fries:
@bhancock8 The entity CID fields are hack/legacy elements. I believe they're required for the CDR demo (not 100% certain)? 

Regardless, storing arbitrary, typed entity spans is better handled outside the `Sentence` object :)

PR #1081, opened 2019-03-20
Adding documentation to utilities in CDR tutorial

eric-czech:
As a part of better understanding the CDR tutorial before applying it to a similar problem, I added some documentation to the utils.py file.  I thought some extra exposition around matching tags from external parsers might help others trying to do the same thing.


bhancock8:
Awesome! Thanks for the improved documentation!

Issue #1082, opened 2019-03-20
SentenceNgramViewer 'gold' argument error 


eric-czech:
Should these lines in the viewer :

https://github.com/HazyResearch/snorkel/blob/cfbbb1dbd3a835b1ddfa129e851797a2833676b0/snorkel/viewer/__init__.py#L166-L167

instead be like this?

```python
candidates = [c for c in self.candidates if c[0].get_parent() == context]
gold = [g for g in self.gold if g[0].get_parent() == context]
```

I'm struggling to see what Context subclass or other model even has a ```context_id``` attribute so I'm assuming that may be incorrect since making that change and passing in a list of candidates for the ```gold``` argument to ```SentenceNgramViewer``` at least avoids an ```Attribute 'context_id' does not exist``` error.

bhancock8:
I think you're right. I don't think the SentenceNgramViewer is part of the unit tests at the moment, so it likely didn't get updated in a previous refactor. If you submit a PR with this fix, we'll merge it in!

PR #1089, opened 2019-03-29
Fixing context equality check in viewer for gold candidates

eric-czech:
Fix for https://github.com/HazyResearch/snorkel/issues/1082 (cc: @bhancock8).

I'd like to add a unit test that at least validates that the viewer can be created, though I was a little unclear on what the strategy with tests is.  The spouses and crowdsourcing dbs that get used in the pytorch and tensorflow tests seem like they could probably be parts of test fixtures that don't necessarily have anything to do with PyTorch or TF functionality right?  It seemed to me like making test data a more generic resource could help with testing many things, including the viewer.  I'll gladly attempt that in another PR if there are no objections that it would be a worthwhile abstraction.

### DIFF
--- a/snorkel/viewer/__init__.py
+++ b/snorkel/viewer/__init__.py
@@ -164,7 +164,7 @@ class Viewer(widgets.DOMWidget):
 
                 # Get the candidates in this context
                 candidates = [c for c in self.candidates if c[0].get_parent() == context]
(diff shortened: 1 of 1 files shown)


bhancock8:
Yep! No objections. The data should be independent from PyTorch or TF. We'd welcome that other PR.

PR #1118, opened 2019-06-12
tox setup

henryre:
Sets up tox for testing, dev, etc.

```
# set up a dev env
tox -e dev

# run tests
tox

# fix style
tox -e fix
```

**Test plan**

`tox` runs!


tox.ini
```diff
@@ -4,4 +4,32 @@ envlist =
     py35,
     py36,
     py37,
+    check,
 isolated_build = true
+
+[testenv]
+description = run the test driver with {basepython}
+deps = -rrequirements-dev.txt
```
bhancock8: -rr

henryre: intentional. silly, but intentional. i have no clue how that's parsed.

PR #1121, opened 2019-06-13
Add LF appliers

henryre:
Adds LF appliers for Pandas and Spark. API is LFs + dataframe -> csr.

Dask applier still needs to be implemented

**Test plan**
* New unit tests for base and Pandas
* Spark applier tested on EMR


ajratner:
@henryre ready for review?

snorkel/labeling/apply/lf_applier.py
```diff
@@ -0,0 +1,44 @@
+from itertools import chain
+from typing import Any, Iterable, List, Tuple
+
+import scipy.sparse as sparse
+from tqdm import tqdm
+
+from snorkel.labeling.lf import LabelingFunction
+from snorkel.types import Example
```
ajratner: I don't like the name `Example`.  What about `DataPoint`

ajratner: Open to suggestions but "example" is weird to me

snorkel/labeling/apply/lf_applier.py
```diff
@@ -0,0 +1,44 @@
...
+
+RowData = List[Tuple[int, int, int]]
+
+
+class BaseLFApplier:
+    def __init__(self, lfs: List[LabelingFunction]) -> None:
+        if not all(isinstance(f, LabelingFunction) for f in lfs):
```
ajratner: N00b question here (I've been out of the loop for a bit...): Are there any generic commands / tools for enforcing type checking automatically, given the type annotations?  Seems redundant to have explicit checks like this?

ajratner: A very cursory check: I see a few libs for letting you add a decorator to enforce types at runtime (https://github.com/RussBaz/enforce, https://github.com/Stewori/pytypes).  Don't know what the "standard" is though.  Perhaps we could just spin our own, just so that this is a decorator e.g. `@enforcetype` instead of this manual check?

snorkel/labeling/apply/lf_applier.py
```diff
@@ -0,0 +1,44 @@
...
+
+def apply_lfs_to_example(
+    example: Example, index: int, lfs: List[LabelingFunction]
+) -> RowData:
+    labels = []
+    for j, lf in enumerate(lfs):
+        y = lf(example)
+        if y != 0:
+            labels.append((index, j, y))
+    return labels
```
ajratner: What is the rationale for not just using e.g. [`sparse.lil_matrix`](https://docs.scipy.org/doc/scipy/reference/generated/scipy.sparse.lil_matrix.html) as the internal representation?

ajratner: And why not just have this be a class method?

snorkel/labeling/apply/lf_applier_pandas.py
```diff
@@ -0,0 +1,36 @@
...
+    for j, lf in enumerate(lfs):
+        y = lf(example)
+        if y != 0:
+            labels.append((j, y))
+    return labels
+
+
+class PandasLFApplier(BaseLFApplier):
+    def apply(self, df: DataFrame) -> sparse.csr_matrix:  # type: ignore
+        apply_fn = partial(apply_lfs_to_example, lfs=self._lfs)
```
ajratner: Nooo `partial` is back... wouldn't need this if `apply_lfs_to_example` was a class method

test/labeling/apply/test_lf_applier.py
```diff
@@ -0,0 +1,53 @@
...
+    def test_lf_applier_pandas_parquet(self) -> None:
+        table_write = pa.Table.from_pandas(pd.DataFrame(dict(a=DATA)))
+        stream = pa.BufferOutputStream()
+        pq.write_table(table_write, stream)
+        buffer = stream.getvalue()
+        reader = pa.BufferReader(buffer)
+        table = pq.read_table(reader)
+        applier = PandasLFApplier([f, g])
+        L = applier.apply(table.to_pandas())
+        np.testing.assert_equal(L.toarray(), L_EXPECTED)
```
ajratner: Spark test..?

snorkel/labeling/apply/lf_applier_pandas.py
```diff
@@ -0,0 +1,36 @@
...
+        tqdm.pandas()
+        labels = df.progress_apply(apply_fn, axis=1)
+        labels_with_index = [
+            [(index, j, y) for j, y in row_labels]
+            for index, row_labels in enumerate(labels)
+        ]
+        return self._matrix_from_row_data(labels_with_index)
```
ajratner: I still think it would be simpler to just construct an incremental sparse matrix (`lil_matrix`) and then convert to csr, then having this internal format -> conversion subfunction.

snorkel/labeling/apply/lf_applier.py
bhancock8: allennlp uses `Instance`, though `DataPoint` is probably the most clear

ajratner: Either way we should be consistent in our verbiage across both code and text (whenever possible).  We say "LFs to label _training data_", but "data" is plural; "datum" sounds pretentious... so that led me to "data point"

henryre: Just to clarify, Example = "training example". But I'm open to whatever. Any objections to DataPoint?

snorkel/labeling/apply/lf_applier.py
henryre: This check is completely unnecessary. I'll remove it. The onus is on the user to use the API correctly, and they'll find out pretty quickly if they don't. Even more quickly if they use mypy.

snorkel/labeling/apply/lf_applier.py
bhancock8: DataPoint sgtm. Just checking: Example can refer to both examples used for training and testing, right?

snorkel/labeling/apply/lf_applier.py
ajratner: Well I think it's nice to have it.  And more generally, it also seems nice to be able to add a decorator that says "if types were defined, enforce them at runtime", so we should have this?

snorkel/labeling/apply/lf_applier.py
henryre: You mean lil vs. csr? Does metal need csr or no? I just usually default to csr since it's the best known sparse matrix format. lil, coo, etc. are scipy land.

henryre: Conversely: why have it be a class method?

ajratner: No I just mean that you're incrementally constructing a sparse matrix (but just in a custom format) and then converting it to a `csr_matrix`.  However scipy already has a sparse matrix format, `lil_matrix`, optimized for incremental construction, and then has a method for conversion to `csr_matrix` (or any other type).

Note: this _is_ a nit, but I think will be cleaner!

snorkel/labeling/apply/lf_applier_pandas.py
henryre: Ah ok. This could easily be used as the target in a parallel/distributed map function (e.g. Dask or Spark) and sent through a pipe. If it's a class method, we'd have to send the entire class through the pipe.

test/labeling/apply/test_lf_applier.py
henryre: Yeah I'm looking for a way to mock that out. PySpark is a 200+ MB install, and putting it in requirements.txt is problematic (overwrites an existing system install). Not worth it for a single unit test (especially since it would just run it on a single node).

As per the test plan, I tested this manually on EMR. As a stopgap, I can write a unit test but disable it by default. That will make manual checks on e.g. EMR easier.

snorkel/labeling/apply/lf_applier_pandas.py
henryre: Pending answer about metal

snorkel/labeling/apply/lf_applier_pandas.py
ajratner: But the class is mainly just the lfs, and they'd have to be sent through too anyway... right?

snorkel/labeling/apply/lf_applier_pandas.py
ajratner: See above- output will still be csr, but can construct in fewer lines / faster as lil

snorkel/labeling/apply/lf_applier.py
henryre: Actually why not just use Instance if it's already in use by a popular package?

snorkel/labeling/apply/lf_applier.py
henryre: Let's talk offline

snorkel/labeling/apply/lf_applier_pandas.py
henryre: Currently, yes. But safer to not make that assumption.

snorkel/labeling/apply/lf_applier.py
henryre: Ok sg

snorkel/labeling/apply/lf_applier.py
ajratner: `DataPoint` maintains a clearer mapping with our overall explanation of how Snorkel works, e.g. in tutorials, etc.

snorkel/labeling/apply/lf_applier_pandas.py
ajratner: I think better to be simpler and cleaner now, then can optimize for distributed execution later

snorkel/labeling/apply/lf_applier.py
henryre: @ajratner I implemented the changes in the new version, but

1. the following snippet is now repeated 3 times due to the pandas.df.apply and pyspark.rdd.map APIs

```
for j, lf in enumerate(self._lfs):
    y = lf(x)
    if y != 0:
```

2. the CSR construction is now slower for all the appliers because `chain.from_iterable(...)` is faster than the loop-based `lil.tocsr()`

Screen Shot 2019-06-14 at 10 52 02 AM

ajratner: Okay the difference is small, but based on this analysis, seems fine to default to the way you had before- I stand corrected!

henryre:
@ajratner Ok good to go

snorkel/labeling/apply/lf_applier.py
```diff
@@ -0,0 +1,42 @@
...
+    def _matrix_from_row_data(self, labels: List[RowData]) -> sparse.csr_matrix:
+        row, col, data = zip(*chain.from_iterable(labels))
+        n, m = len(labels), len(self._lfs)
+        return sparse.csr_matrix((data, (row, col)), shape=(n, m))
+
+    def apply(self, data_points: Any, *args: Any, **kwargs: Any) -> sparse.csr_matrix:
+        raise NotImplementedError
+
+
+def apply_lfs_to_data_point(
```
ajratner: It still kills me inside that this isn't in `BaseLFApplier`....

ajratner: How about put it in the class for now, then we'll empirically test if it hurts parallelism.  *I'll* do it

henryre: Go for it! I'm merging this in assuming you'll do that in a separate PR

PR #1122, opened 2019-06-13
[EASY] Small patches to language tools

henryre:
Material changes:

* Remove seed-isort-config :/ . It adds new imports on the same line, which is a nightmare for merges/rebases
* Reactive flake8 F401 (unused imports)
* Add some known imports

**Test plan**
Travis? idk

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -121,3 +121,6 @@ dmypy.json
 
 # Pyre type checker
 .pyre/
+
+# Editors
+.vscode/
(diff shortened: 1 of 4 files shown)


henryre:
The missing known_third_party imports were never added outside of `git commit`. isort is still checked in `tox -e check` and `git commit`. Order fixes happen in `tox -e fix`. Missing known_third_party imports aren't added anywhere now.

PR #1123, opened 2019-06-13
Add codecov config

henryre:
See title. Used https://github.com/codecov/example-python as a guide

**Test plan**
Manual verification of Travis and codecov

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,14 @@
+[run]
+branch = True
+source = snorkel
+
+[report]
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+    def __repr__
+ignore_errors = True
+omit =
+    tests/*
+    tutorials/*
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,6 +22,7 @@ jobs:
 install:
   - pip install -U pip setuptools
   - pip install -U tox==3.12.0
+  - pip install -U codecov
   - tox --notest
 
 script:
--- a/.travis.yml
+++ b/.travis.yml
@@ -10,19 +10,22 @@ cache: pip
 dist: xenial
 
 env:
-  TOXENV=py,check,type
+  TOXENV=py,check,type,coverage
 
 jobs:
   include:
-  - name: "Tests: python3.6"
-    python: 3.6
-  - name: "Tests: python3.7"
-    python: 3.7
+    - name: "Tests: python3.6"
+      python: 3.6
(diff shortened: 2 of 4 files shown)


.travis.yml
```diff
@@ -10,19 +10,23 @@ cache: pip
...
+    - name: "Tests: python3.7"
+      python: 3.7
 
 install:
-- pip install -U pip setuptools
-- pip install -U tox==3.12.0
-- tox --notest
+  - pip install -U pip setuptools
+  - pip install -U tox==3.12.0
+  - pip install -U codecov
```
ajratner: Can we run codecov within tox https://github.com/codecov/example-python#testing-with-tox?

tox.ini
```diff
@@ -23,9 +23,14 @@ commands =
 [testenv:type]
 description = run static type checking
 basepython = python3
-usedevelop = True
 commands = mypy -p snorkel
 
+[testenv:coverage]
```
ajratner: Note (earlier in file): typo line 12

.coveragerc
```diff
@@ -0,0 +1,14 @@
...
+branch = True
+source = snorkel
+
+[report]
+exclude_lines =
+    pragma: no cover
+    raise NotImplementedError
+    if __name__ == .__main__.:
+    def __repr__
+ignore_errors = True
```
ajratner: ?

tox.ini
ajratner: what kind of coverage tests are these?  Redundant with codecov?

henryre: Not a typo: https://tox.readthedocs.io/en/latest/example/basic.html#depending-on-requirements-txt-or-defining-constraints

henryre: This command (`pytest --cov=...`) generates the `.coverage` file. The command `codecov` just looks for a file called `.coverage` and uploads it. See https://github.com/codecov/example-python/blob/master/.travis.yml

.travis.yml
henryre: Yes, but there's really no reason to. The only place we'd ever run `codecov` is Travis, so I'd prefer to keep it isolated to that environment (as opposed to putting the dep in `requirements-dev.txt` and adding possibly confusing env in `tox.ini`)

.coveragerc
henryre: This is the generic .coveragerc suggested by codecov: https://gist.github.com/codecov-io/bf15bde2c7db1a011b6e

.travis.yml
ajratner: Okay makes sense

PR #1124, opened 2019-06-14
Add LF preprocessors

henryre:
Preprocessors are shared resources that are applied to data points before they are subjected to the logic in labeling functions. The idea is that e.g. multiple LFs want to pass a text examples through SpaCy before executing, so the preprocessor is defined once and passed to all of the LFs.

* Examples of custom preprocessors are in the unit tests
* A SpaCy-based preprocessor is implemented
* The interface for TFs will be similar, so we can abstract out later (e.g. the logic for reconstructing data points with changed fields)
* We can later add caching, etc.

**Test plan**
* Added unit tests for preprocessors and everything that implements them upstream (LFs, appliers)

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -8,6 +8,7 @@ exclude_lines =
     raise NotImplementedError
     if __name__ == .__main__.:
     def __repr__
+    SparkLFApplier
 ignore_errors = True
 omit =
     test/*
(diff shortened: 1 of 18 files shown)


henryre:
The main idea is that the LFAppliers will set it appropriately. It's rare that a user will call a LabelingFunction in isolation, so likely will never have to set the mode.

PR #1134, opened 2019-06-18
Add MTL logger tools

bhancock8:
Port over from MeTaL/mmtl and Emmental

- LogManager checks when it's time to log, evaluate, and/or checkpoint
- LogWriter/TensorboardWriter write logs
- Checkpointer handles bookeeping, saving, and loading of best model

**Test plan**
- Ported existing tests for LogManager
- Checkpointer tests will be added once a model is defined that can be saved/loaded


snorkel/mtl/loggers/__init__.py
```diff
@@ -0,0 +1,15 @@
+logging_default_config = {
```
henryre: Let's get this outside of init

snorkel/mtl/loggers/log_manager.py
```diff
@@ -0,0 +1,174 @@
+import logging
+import os
+from datetime import datetime
+from typing import Dict
+
+import torch
+
+from snorkel.mtl.loggers import logging_default_config
```
henryre: e.g. `from .checkpointer import ...` is fine here

snorkel/types/logger.py
```diff
@@ -0,0 +1,5 @@
+from typing import Any, Dict
+
+# A config dict is a nested dict with a max depth of 3
+Config = Dict[str, Any]
+Path = str
```
henryre: this might be confusing to people since `Path` objects exist elsewhere. can we just use str in-place?

henryre:
Also needs to be rebased onto redux

snorkel/mtl/loggers/log_manager.py
bhancock8: PEP8 recommends absolute imports ("Absolute imports are recommended, as they are usually more readable and tend to be better behaved (or at least give better error messages) if the import system is incorrectly configured"). Will change here for now though to be consistent with the files you've already moved over. Can discuss and update later if need be.

snorkel/types/logger.py
bhancock8: Yep, will remove the `Path` type for now, though in a future commit I suggest we call-out where strings are paths explicitly.

PR #1135, opened 2019-06-18
Add MTL data classes

bhancock8:
Port over from MeTaL/mmtl and Emmental

**Test plan**
- Port over and update existing tests for Dataset, DataLoader, and utility functions

### DIFF
--- a/snorkel/mtl/data.py
+++ b/snorkel/mtl/data.py
@@ -4,7 +4,7 @@ from collections import defaultdict
 import torch
 from torch.utils.data import DataLoader, Dataset
 
-from snorkel.mtl.utils import list_to_tensor
+from .utils import list_to_tensor
 
 
 class MultitaskDataset(Dataset):
--- a/snorkel/mtl/data.py
+++ b/snorkel/mtl/data.py
@@ -1,0 +1,166 @@
+import logging
+from collections import defaultdict
+
+import torch
+from torch.utils.data import DataLoader, Dataset
+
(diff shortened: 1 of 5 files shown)


snorkel/mtl/data.py
```diff
@@ -0,0 +1,166 @@
+import logging
+from collections import defaultdict
+
+import torch
+from torch.utils.data import DataLoader, Dataset
+
+from snorkel.mtl.utils import list_to_tensor
```
henryre: nit: `from .utils import ...`

test/mtl/test_mtl_utils.py
```diff
@@ -0,0 +1,112 @@
+import unittest
```
henryre: nit: rename `test_utils.py`

bhancock8: I agree, but fun fact: tox won't let us have two tests with the same name:
```
import file mismatch:
imported module 'test_utils' has this __file__ attribute:
  /Users/braden/repos/snorkel/test/model/test_utils.py
which is not the same as the test file we want to collect:
  /Users/braden/repos/snorkel/test/mtl/test_utils.py
HINT: remove __pycache__ / .pyc files and/or use a unique basename for your test file modules
```

bhancock8: It's actually a pytest issue:
https://github.com/pytest-dev/pytest/issues/529

PR #1136, opened 2019-06-19
Add LambdaPreprocessor and decorator

henryre:
* Adds LambdaPreprocessor (and associated decorator) that makes it easier to define a simple preprocessor that just executes a function without setup
* Next diff: factoring out common functionality (DataPoint reconstruction) for use with transformation functions

**Test plan**
* Incorporated changes into unit tests

### DIFF
--- a/snorkel/labeling/preprocess/core.py
+++ b/snorkel/labeling/preprocess/core.py
@@ -1,6 +1,6 @@
 from enum import Enum, auto
 from types import SimpleNamespace
-from typing import Any, Mapping, NamedTuple, Union
+from typing import Any, Callable, Mapping, NamedTuple, Union
 
 from snorkel.types import DataPoint, FieldMap
 
@@ -82,3 +82,38 @@ class Preprocessor:
             raise ValueError(
                 f"Preprocessor mode {self.mode} not recognized. Options: {PreprocessorMode}."
             )
+
+
+class LambdaPreprocessor(Preprocessor):
+    def __init__(
+        self,
+        f: Callable[..., FieldMap],
+        field_names: Mapping[str, str],
+        preprocessed_field_names: Mapping[str, str],
+        mode: PreprocessorMode = PreprocessorMode.NONE,
+    ) -> None:
+        super().__init__(field_names, preprocessed_field_names, mode)
+        self._f = f
+
+    def preprocess(self, **kwargs: Any) -> FieldMap:
+        return self._f(**kwargs)
+
+
+class preprocessor:
+    def __init__(
+        self,
+        field_names: Mapping[str, str],
+        preprocessed_field_names: Mapping[str, str],
+        mode: PreprocessorMode = PreprocessorMode.NONE,
+    ) -> None:
+        self.field_names = field_names
+        self.preprocessed_field_names = preprocessed_field_names
+        self.mode = mode
+
+    def __call__(self, f: Callable[..., FieldMap]) -> LambdaPreprocessor:
+        return LambdaPreprocessor(
+            f=f,
+            field_names=self.field_names,
+            preprocessed_field_names=self.preprocessed_field_names,
+            mode=self.mode,
+        )
--- a/snorkel/labeling/preprocess/core.py
+++ b/snorkel/labeling/preprocess/core.py
@@ -1,6 +1,7 @@
+import inspect
 from enum import Enum, auto
 from types import SimpleNamespace
-from typing import Any, Callable, Mapping, NamedTuple, Union
+from typing import Any, Callable, Mapping, NamedTuple, Optional, Union
 
 from snorkel.types import DataPoint, FieldMap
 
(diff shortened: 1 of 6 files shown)


ajratner:
@henryre Lm take a look too?  Next 5-10 min

test/labeling/apply/test_lf_applier.py
```diff
@@ -9,20 +9,16 @@
...
 from snorkel.labeling.preprocess.nlp import SpacyPreprocessor
 from snorkel.types import DataPoint, FieldMap
 
 
-class SquarePreprocessor(Preprocessor):
-    def __init__(self, x_field: str, squared_x_field: str) -> None:
-        super().__init__(dict(x=x_field), dict(x=squared_x_field))
+@preprocessor(dict(x="a"), dict(x="a"))
+def square(x: float) -> FieldMap:
+    return dict(x=x ** 2)
```
ajratner: Two requests:

1. Make the tests more readable with (a) comments and (b) readable attribute names (e.g. this should be `@preprocessor(dict(x="num"), dict(x="num_squared"))`.  I think improving test readability is *very* critical to team development speed

2. It would be really nice if you could provide only one attribute, and this would default to modifying the attribute in place

test/labeling/preprocess/test_core.py
```diff
@@ -15,20 +15,16 @@ def preprocess(self, text: str) -> FieldMap:  # type: ignore
...
 
-class SquarePreprocessor(Preprocessor):
-    def __init__(self, x_field: str, squared_x_field: str) -> None:
-        super().__init__(dict(x=x_field), dict(x=squared_x_field))
-
-    def preprocess(self, x: float) -> FieldMap:  # type: ignore
-        return dict(x=x ** 2)
+@preprocessor(dict(x="a"), dict(x="c"))
+def square(x: float) -> FieldMap:
+    return dict(x=x ** 2)
```
ajratner: Again just add a comment here, otherwise is cryptic

ajratner: + readable attrib names

tutorials/drybell/nlp_lf.py
```diff
@@ -39,17 +39,11 @@
...
 
 
-class CombineTextPreprocessor(Preprocessor):
-    def __init__(self, title_field, body_field, article_field):
-        super().__init__(
-            dict(title=title_field, body=body_field), dict(article=article_field)
-        )
+@preprocessor(dict(title="title", body="body"), dict(article="article"))
+def combine_text_preprocessor(title: str, body: str) -> FieldMap:
+    return dict(article=f"{title} {body}")
```
ajratner: Okay here's what's bugging me: this decorator is still super clunky.  Why not [get the input fields from the preprocess args](https://stackoverflow.com/questions/218616/getting-method-parameter-names-in-python) and just let the user implicitly define the output fields by the return statement?  Then you'd have a super nice interface where you only would need to add a `@preprocessor` decorator here!

Currently, there's just so much redundant info being written down for a simple op...

henryre: Good call, I'll implement this for LambdaPreprocessor. Might take more thought for Preprocessor, and out-of-scope for this diff.

henryre:
@ajratner @bhancock8 changes made, take another look

snorkel/labeling/preprocess/core.py
```diff
@@ -58,9 +56,11 @@ def preprocess(self, **kwargs: Any) -> FieldMap:
...
         field_map = {k: getattr(x, v) for k, v in self.field_names.items()}
         preprocessed_fields = self.preprocess(**field_map)
-        preprocessed_fields = {
-            v: preprocessed_fields[k] for k, v in self.preprocessed_field_names.items()
-        }
+        if self.preprocessed_field_names is not None:
+            preprocessed_fields = {
+                v: preprocessed_fields[k]
+                for k, v in self.preprocessed_field_names.items()
+            }
```
ajratner: Some of my comments are redundant with the other PR- I think there should be a check, since this depends on user returning a dict-type?

snorkel/labeling/preprocess/core.py
```diff
@@ -85,35 +85,33 @@ def __call__(self, x: DataPoint) -> DataPoint:
...
+        """Convenience class for Preprocessors that execute a simple
+        function with no set up. The function arguments are parsed
+        to determine the input field names of the data points.
+
+        Args:
+            * f: the function executing the preprocessing operation
+        """
         self._f = f
+        field_names = {k: k for k in inspect.getfullargspec(f)[0]}
+        super().__init__(field_names=field_names, preprocessed_field_names=None)
```
ajratner: LGTM!

snorkel/labeling/preprocess/core.py
```diff
@@ -85,35 +85,33 @@ def __call__(self, x: DataPoint) -> DataPoint:
...
-        return LambdaPreprocessor(
-            f=f,
-            field_names=self.field_names,
-            preprocessed_field_names=self.preprocessed_field_names,
-            mode=self.mode,
-        )
+        ```
+        @preprocessor()
+        def concatenate_text(title: str, body: str) -> FieldMap:
+            return f"{title} {body}"
```
ajratner: I guess I commented on PRs in wrong order... this example should return a dict right?

henryre: Yes correct, will update

henryre:
Going to land this then address comments on other PR. Git didn't track the move so changes would be made twice.

PR #1140, opened 2019-06-19
[EASY] Remove test artifacts

bhancock8:
LogManager tests currently leave a logs/ directory after running. This creates and removes that directory during setup/cleanup.

**Test plan**:
Passes existing tests and no logs/ directory exists when it finishes.


henryre:
re: test plan, did you confirm that it's writing to the expected location?

bhancock8:
Yes, temp logs file gets written to expected location.

PR #1141, opened 2019-06-19
Synthetic data generation

ajratner:
This PR adds code to core Snorkel lib (under `snorkel/synthetic`) for simple synthetic data generation; specifically:
- Generates a mixture of Gaussians dataset
- Generates a set of LFs that each fire according to one data feature

Also includes a basic tutorial in VSC Jupyter notebook format.

Follow-on TODOs:
- Hook in the `LabelModel` to the tutorial
- Hook in the `EndModel` to the tutorial
- _TFs:_ Add noise components to data + basic TFs that permute these elements
- Add utils for more complex covariance matrix in dataset generation (which will lead to LFs with correlations)

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -51,5 +51,8 @@ ignore_missing_imports = True
 [mypy-pyspark]
 ignore_missing_imports = True
 
+[mypy-spacy]
+ignore_missing_imports = True
+
 [mypy-tqdm]
 ignore_missing_imports = True
--- a/setup.cfg
+++ b/setup.cfg
@@ -1,6 +1,7 @@
 [flake8]
 ignore = 
     E203,
+    E265, # Throws errors for '#%%' delimiter in VSCode jupyter notebook syntax
     E266,
     E501,
     E731,
--- a/snorkel/synthetic/synthetic_data.py
+++ b/snorkel/synthetic/synthetic_data.py
@@ -1,0 +1,69 @@
+from functools import partial
+from types import SimpleNamespace
+from typing import List, Optional
+
+import numpy as np
+import pandas as pd
+
+from snorkel.labeling.lf import LabelingFunction
+from snorkel.types import DataPoint
+
+
+def generate_mog_dataset(
+    n: int, d: int, cov: Optional[np.ndarray] = None
+) -> pd.DataFrame:
+    """
+    Generates a simple mixture-of-gaussians (MOG) dataset consisting of
+    d-dim vectors x \in \mathbb{R}^d, and binary labels y \in {-1,1}.
+    Returns as a pandas DataFrame
+    """
+    ones = np.ones(d)
+    nh = int(np.floor(n / 2))
+    if cov is None:
+        cov = np.diag(np.random.random(d))
+    X_pos = np.random.multivariate_normal(ones, cov, nh)
+    X_neg = np.random.multivariate_normal(-1 * ones, cov, n - nh)
+
+    # Combine and shuffle
+    X = np.vstack([X_pos, X_neg])
+    Y = np.concatenate([np.ones(nh), -1 * np.ones(n - nh)]).astype(int)
+    order = list(range(n))
+    np.random.shuffle(order)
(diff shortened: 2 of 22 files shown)


setup.cfg
```diff
@@ -1,6 +1,7 @@
 [flake8]
 ignore = 
     E203,
+    E265, # Throws errors for '#%%' delimiter in VSCode jupyter notebook syntax
```
henryre: Let's remove this and add `# noqa: E265` on the errored lines

snorkel/synthetic/synthetic_data.py
```diff
@@ -0,0 +1,69 @@
...
+import numpy as np
+import pandas as pd
+
+from snorkel.labeling.lf import LabelingFunction
+from snorkel.types import DataPoint
+
+
+def generate_mog_dataset(
+    n: int, d: int, cov: Optional[np.ndarray] = None
+) -> pd.DataFrame:
```
henryre: This returns List[DataPoint]

snorkel/synthetic/synthetic_data.py
```diff
@@ -0,0 +1,69 @@
...
+    """
+    Takes as input a DataPoint x with attribute x \in \mathbb{R}^d, and
+    outputs based on the `index`th entry, with probability `abstain_rate` of outputting 0 (abstain).
+    """
+    if np.random.random() < abstain_rate:
+        return 0
+    elif x.x[index] > 0:
+        return 1
+    else:
+        return -1
```
henryre: The label model ported from MeTaL takes -1, 1? Or 0, 1, 2?

test/synthetic/test_synthetic_data.py
```diff
@@ -0,0 +1,75 @@
...
+    def test_output_types(self) -> None:
+        # Test x
+        x = self.data[0].x
+        self.assertIsInstance(x, np.ndarray)
+        self.assertEqual(x.shape[0], self.d)
+
+        # Test y
+        self.assertIsInstance(self.data[0].y, np.integer)
+
+    def test_dataset_means(self) -> None:
```
henryre: Fine for now, but we probably want to run this for multiple setUp calls since its rand

test/synthetic/test_synthetic_data.py
```diff
@@ -0,0 +1,75 @@
...
+        cov[0, 0] = 0.0
+        data = generate_mog_dataset(self.n, self.d, cov=cov)
+
+        # Generate a single-feature LF and apply to the data
+        LFs = generate_single_feature_LFs(1)
+        lf_applier = LFApplier(LFs)
+        L = lf_applier.apply(data)
+
+        # Get accuracies
+        Ld = np.array(L.todense())
```
henryre: redundant?

setup.cfg
ajratner: That would be super annoying, as this delimiter is the equivalent of starting a new cell in Jupyter...

snorkel/synthetic/synthetic_data.py
ajratner: Good catch

snorkel/synthetic/synthetic_data.py
ajratner: It can take {-1,1}; either way I'll leave that for the next PR (integrating label and end models into this example)

setup.cfg
ajratner: The problem is these are still just `.py` files, so can't do any kind of per-file ignore (which is an option in flake).  Unless there is a *strong* objection I'm going to leave this in- I think it's a really nice function in VSC, but more importantly, potentially a really nice way to store and version our jupyter notebooks going forwards

test/synthetic/test_synthetic_data.py
ajratner: For this simple example, should always work- but down the road sure

setup.cfg
henryre: Ok that's fine

PR #1143, opened 2019-06-19
Add MTL model and task objects

bhancock8:
Add MTL Model and Task objects.

Each Task defines:
- a pool of modules
- a flow through those modules (a sequence of modules to go through, with a map of what fields to pull out of the intermediate outputs dictionary to use as input for each module). Importantly, the flow can reuse the same module multiple times (e.g., as in multi-hop frameworks or multiple-choice setups where multiple candidates are passed through the network before making predictions)
- a loss function
- an output function (e.g., a final softmax)
- a scorer (a bundle of metrics that are defined and meaningful for that task)
A Task doesn't do much by itself.

Each Model is built from a collection of Tasks. The model pools all modules from the tasks (merging shared modules based on their keys), and defines a forward pass that respects the flows defined in the tasks, with appropriate caching of intermediate results.

**Test plan**
- `test_task.py` confirms that a Task object can be initialized; there's not much more to test in isolation
- `test_model.py` confirms that a Model can be created from a single task, or two two tasks that overlap completely, not at all, or partially. In each case, we confirm that the model contains the proper number of tasks and modules (i.e., is sharing modules where it should).
- Additional tests will be included when the `Trainer` object is added

### DIFF
--- a/snorkel/mtl/model.py
+++ b/snorkel/mtl/model.py
@@ -1,0 +1,402 @@
+import logging
+import os
+from collections import defaultdict
+from collections.abc import Iterable
+
+import numpy as np
+import torch
+import torch.nn as nn
+
+from snorkel.mtl.task import Task
+from snorkel.mtl.utils import move_to_device, prob_to_pred, recursive_merge_dicts
+
+model_default_config = {
+    "model_path": None,  # the path to a saved checkpoint to initialize with
+    "device": 0,  # gpu id (int) or -1 for cpu
+    "dataparallel": True,
+}
+
+
+class MultitaskModel(nn.Module):
+    """A class to build multi-task model.
+
+    :param name: Name of the model
+    :type name: str
+    :param tasks: a list of Task that trains jointly
+    :type tasks: list of Task project
+    """
+
+    def __init__(self, name=None, tasks=None, **kwargs):
+        super().__init__()
+        self.config = recursive_merge_dicts(
+            model_default_config, kwargs, misses="insert"
+        )
+        self.name = name if name is not None else type(self).__name__
+
+        # Initiate the model attributes
+        self.module_pool = nn.ModuleDict()
+        self.task_names = set()
+        self.task_flows = dict()
+        self.loss_funcs = dict()
+        self.output_funcs = dict()
+        self.scorers = dict()
+
+        # Build network with given tasks
+        if tasks is not None:
+            self._build_network(tasks)
+
+        logging.info(
+            f"Created multi-task model {self.name} that contains "
+            f"task {self.task_names}."
+        )
+
+        # Move model to specified device
+        self._move_to_device()
+
+    def _move_to_device(self):
+        """Move model to specified device."""
+
+        if self.config["device"] >= 0:
+            if torch.cuda.is_available():
+                logging.info(f"Moving model to GPU " f"(cuda:{self.config['device']}).")
+                self.to(torch.device(f"cuda:{self.config['device']}"))
+            else:
+                logging.info("No cuda device available. Switch to cpu instead.")
+
+    def _build_network(self, tasks):
+        """Build the MTL network using all tasks"""
+
+        if not isinstance(tasks, Iterable):
+            tasks = [tasks]
+        for task in tasks:
+            if task.name in self.task_names:
+                raise ValueError(
+                    f"Found duplicate task {task.name}, different task should use "
+                    f"different task name."
+                )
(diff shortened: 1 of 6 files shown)


bhancock8:
Note: I believe that the specification of the `task_flow` parameter of Tasks can be simplified or clarified substantially, but I propose that this be done in a subsequent PR.

ajratner:
Note: This will become decreasingly relevant, but I would mark which PRs / pieces or PRs are ports from existing codebases (where presumably reviews have been done = lower standard of review) or new code

snorkel/mtl/task.py
```diff
@@ -0,0 +1,36 @@
...
+
+from torch import nn
+
+
+class Task(object):
+    """A single Task in a multi-task problem
+
+    :param name: The name of the task (Primary key).
+    :type name: str
+    :param module_pool: A dict of all modules that uses in the task.
```
ajratner: Grammar

snorkel/mtl/model.py
```diff
@@ -0,0 +1,402 @@
...
+        super().__init__()
+        self.config = recursive_merge_dicts(
+            model_default_config, kwargs, misses="insert"
+        )
+        self.name = name if name is not None else type(self).__name__
+
+        # Initiate the model attributes
+        self.module_pool = nn.ModuleDict()
+        self.task_names = set()
+        self.task_flows = dict()
```
ajratner: How often are task flows used in full generality, versus following simple templates?  Am assuming this is part of what you meant when you said this could be simplified?  To discuss later!

snorkel/mtl/model.py
```diff
@@ -0,0 +1,402 @@
...
+                raise ValueError(f"Unrecognized task type {task}.")
+            self.add_task(task)
+
+    def add_task(self, task):
+        """Add a single task into MTL network"""
+
+        # Combine module_pool from all tasks
+        for key in task.module_pool.keys():
+            if key in self.module_pool.keys():
+                if self.config["dataparallel"]:
```
ajratner: Should already be wrapped with `nn.DataParallel` given the below lines, right?  This is redundant?

bhancock8:
This is part 4 of a 5 part porting of existing mmtl pieces:
- mtl logging (merged)
- mtl data (merged)
- mtl metrics (merged)
- mtl model (this)
- mtl trainer (upcoming)
All of these have been modified slightly in the port, but their general logic has been used quite a bit in their previous repo. I think they could all still use code review for how we feel about the design, but I move that they be added first off a basic sanity check and then refactored so that we go from working to working, rather than making major updates as they're introduced.

snorkel/mtl/model.py
```diff
@@ -0,0 +1,402 @@
...
+        self.loss_funcs[task.name] = task.loss_func
+        # Update output functions
+        self.output_funcs[task.name] = task.output_func
+        # Collect scorers
+        self.scorers[task.name] = task.scorer
+
+        # Move model to specified device
+        self._move_to_device()
+
+    def remove_task(self, task_name):
```
ajratner: How often is it necessary to update or remove tasks...?  If answer is close to never, I would remove these pieces of code

snorkel/mtl/model.py
```diff
@@ -0,0 +1,402 @@
...
+        del self.scorers[task_name]
+        # TODO: remove the modules only associate with that task
+
+    def __repr__(self):
+        cls_name = type(self).__name__
+        return f"{cls_name}(name={self.name})"
+
+    def forward(self, X_dict, task_names):
+        """Forward based on input and task
+        Note: We assume that all shared the modules from all tasks are based on the
```
ajratner: Grammar

snorkel/mtl/model.py
```diff
@@ -0,0 +1,402 @@
...
+        X_dict = move_to_device(X_dict, self.config["device"])
+
+        outputs = dict()
+        outputs["_input_"] = X_dict
+
+        # Call forward for each task
+        for task_name in task_names:
+            task_flow = self.task_flows[task_name]
+
+            for action in task_flow:
```
ajratner: A little cryptic what this action object is

snorkel/mtl/model.py
bhancock8: We actually did use some of the fancier logic for the SuperGLUE tasks (e.g., when you have a multiple choice question and you need to run all four options through the network separately first before making a prediction). I think the interface definitely could use some work though, and we should be able to make special simplified syntax for the most common cases at the very least.

PR #1145, opened 2019-06-20
Add MTL Trainer

bhancock8:
This PR includes:
- (Primary) Adds the Trainer class which operates on a model and dataloaders
- A little cleanup of other pieces that touch training (see commit messages). This includes adjusting the defaults so that by default no logs or checkpoints are written (so users' workspaces aren't spammed with logs if they don't want them)

**Test plan**
Unit tests build a single-task model and multi-task model and confirms that for both, the initial accuracy is below a threshold and final accuracy is above a threshold, confirming both that the model trained and that we're successfully evaluating/returning scores. We set random seeds so that each run is deterministic (i.e., there won't be rare outlier runs that spontaneously fail the unit test).

### DIFF
--- a/snorkel/mtl/modules/utils.py
+++ b/snorkel/mtl/modules/utils.py
@@ -1,6 +1,6 @@
 """
-For compatibility with SuperGLUE tutorials, we we use these wrappers around
-standard pytorch functional operators.
+Commonly used loss and output functions for Task objects
+# TODO: move filtering of actives and pulling out modules before these functions
 """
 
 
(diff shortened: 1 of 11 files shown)


ajratner:
@bhancock8 tag me again when tests pass

bhancock8:
Ah, one sec. The commit with the schedulers didn't make. Will update and re-ping.

bhancock8:
Good to go now @ajratner!

snorkel/mtl/modules/utils.py
```diff
@@ -1,15 +1,15 @@
...
-F.softmax can be used directly.
 """
 
 
 import torch.nn.functional as F
 
 
 def ce_loss(module_name, outputs, Y, active):
+    # Subtract 1 from Ys to account for Snorkel reserving the label 0 for abstains
+    # (but F.cross_entropy() expect 0-indexed labels)
```
ajratner: Be clear that this is for hard (non-prob) labels

snorkel/mtl/trainer.py
```diff
@@ -0,0 +1,380 @@
...
+            train_split = [train_split]
+
+        train_dataloaders = [
+            dataloader for dataloader in dataloaders if dataloader.split in train_split
+        ]
+
+        if not train_dataloaders:
+            raise ValueError(
+                f"Cannot find the specified train_split "
+                f'{self.config["train_split"]} in datloaders.'
```
ajratner: Typo

PR #1146, opened 2019-06-20
Synthetic Tutorial: LabelModel

ajratner:
This PR puts the LabelModel into the synthetic data tutorial, and adds a helper function to `LabelModel` for this

### DIFF
--- a/.pylintrc
+++ b/.pylintrc
@@ -1,0 +1,6 @@
+[MASTER]
+extension-pkg-whitelist=numpy,torch
+
+[TYPECHECK]
+ignored-modules=numpy,torch
+ignored-classes=numpy,torch
--- a/test/synthetic/test_synthetic_data.py
+++ b/test/synthetic/test_synthetic_data.py
@@ -27,9 +27,9 @@ class TestGenerateMOGDataset(unittest.TestCase):
     def test_dataset_means(self) -> None:
         # Compute the means of the two gaussians and confirm that they
         # are different
-        m0 = np.mean([d.x for d in self.data if d.y == -1])
         m1 = np.mean([d.x for d in self.data if d.y == 1])
-        dist = np.linalg.norm(m1 - m0)
+        m2 = np.mean([d.x for d in self.data if d.y == 2])
+        dist = np.linalg.norm(m1 - m2)
         self.assertGreater(dist, 1)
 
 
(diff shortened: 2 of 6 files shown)


tutorials/synthetic/synthetic.py
```diff
@@ -47,14 +47,29 @@
...
+label_model = LabelModel()
+label_model.train_model(L)
+accs_est = label_model.get_accuracies()
+print(f"Avg. LF accuracy estimation error: {np.mean(np.abs(accs_emp - accs_est))}")
+
+#%%
+# Get probabilistic training labels
+Y_prob = label_model.predict_proba(L)
+
+# Get the hard-thresholded predictions and compute the accuracy
```
bhancock8: These utility function (prob_to_pred() and pred_to_prob()) are in snorkel/mtl/utils.py at the moment, and can be moved up a level, since they're not mtl-specific.

test/synthetic/test_synthetic_data.py
```diff
@@ -50,10 +50,10 @@ def test_perfect_LF(self) -> None:
         L = lf_applier.apply(data)
 
         # Get accuracies
-        Ld = L.todense()
+        # TODO: Replace this with a generic utility function
         Y = np.array([d.y for d in data])
-        accs = Ld.T @ Y / self.n
-        self.assertEqual(accs[0], 1.0)
+        acc = np.where(L.getcol(0).toarray().reshape(-1) == Y, 1, 0).sum() / self.n
```
bhancock8: How does getcol() differ from just slicing out that column?

.pylintrc
```diff
@@ -0,0 +1,6 @@
+[MASTER]
+extension-pkg-whitelist=numpy,torch
```
bhancock8: Thank you for this. :)

tutorials/synthetic/synthetic.py
ajratner: Will add utils in

ajratner: Replaced all the generic steps with utils.

Side note: I almost started cleaning up and merging all the util files... but left that for another PR.

.pylintrc
```diff
@@ -0,0 +1,6 @@
+[MASTER]
```
henryre: We don't use pylint anywhere, so we can either add it (on top of flake8), or this should go outside of git (https://stackoverflow.com/a/22449845)

test/synthetic/test_synthetic_data.py
```diff
@@ -50,10 +50,10 @@ def test_perfect_LF(self) -> None:
         L = lf_applier.apply(data)
 
         # Get accuracies
-        Ld = L.todense()
+        # TODO: Replace this with a generic utility function
         Y = np.array([d.y for d in data])
-        accs = Ld.T @ Y / self.n
-        self.assertEqual(accs[0], 1.0)
+        acc = np.where(L[:, 0].toarray().reshape(-1) == Y, 1, 0).sum() / self.n
```
henryre: can we just use https://scikit-learn.org/stable/modules/generated/sklearn.metrics.accuracy_score.html?

PR #1147, opened 2019-06-20
Added in end model to synthetic tutorial

ajratner:
[DRAFT PR] Adds in the basic `EndModel` from metal to the synthetic tutorial

@henryre @bhancock8 @paroma My goal is to have a basic, single-task end model working.  Is it worth merging this in, or should I use the MTL code, or...?  (Note: I know could also import external lib, but while we still support some end model capability, want to have it here)

Let's discuss the above and then can finish & clean this up, discard, or modify.

### DIFF
--- a/snorkel/end_model/__init__.py
+++ b/snorkel/end_model/__init__.py
@@ -1,0 +1,6 @@
+from .end_model import EndModel
+from .identity_module import IdentityModule
+from .logreg import LogisticRegression
+from .loss import SoftCrossEntropyLoss
+
+__all__ = ["EndModel", "IdentityModule", "LogisticRegression", "SoftCrossEntropyLoss"]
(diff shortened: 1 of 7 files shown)


henryre:
Waiting to comment until MTL vs. not MTL is resolved.

In general: there are a bunch of references to Metal (some e.g. MetalDataset that I didn't catch). Lets strip those out

bhancock8:
In the interest of going from working to working, I say go ahead with porting the EndModel for now. Can merge in a follow-up PR if we have time. It should certainly be merged down the road, but let's get a complete library first.

ajratner:
@bhancock8 seems like per our convo just now, we will kill this PR and just create a simple subclass of `MultitaskModel` that does single task.  I'll leave this open till that route is confirmed though.

PR #1144, opened 2019-06-19
Update setup config and instructions

henryre:
* Removes pre-commit, rely on tox only
* Strips out README, adds latest install instructions
* Updates CONTRIBUTING for correct dev env setup
* Updates tox dev setup to install an expected virtualenv in `.env`

**Test plan**
Manually validated all listed commands on a fresh EC2 Ubuntu instance

PR #1149, opened 2019-06-21
[EASY] Fix .venv -> .env refactor

bhancock8:
Fix a bug introduced in #1144 where tox creates '.env' but black and isort ignore '.venv'

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ exclude = '''
   | \.git
   | \.mypy_cache
   | \.tox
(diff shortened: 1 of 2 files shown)


henryre:
@bhancock8 should we just ignore both? Also related: I had a `logs/` directory show up after running tox and git wanted to commit it. Any ideas?

bhancock8:
Re: the `logs/` file, that's the default log directory and logging was on by default. That's fixed in #1150.

PR #1150, opened 2019-06-21
Refactor configs/loggers to enable SuperGLUE

bhancock8:
All refactoring, no new material:

1. All configs (model_config, trainer_config, logger_config, checkpointer_config, etc.) are now found in `snorkel_config.py` for simplicity/cleanliness

2. Untangle Checkpointer, LogManager, and LogWriter for compatibility with SuperGLUE run scripts and general modularity. The LogManager dictates when logging or checkpointing trigger, but the LogWriter and Checkpointer can be now be initialized and used without first instantiating the LogManager.

3. Replace vanilla `assert` statements with unittest `assertTrue` and `assertFalse` for slightly more informative error messages.

**Test plan**
No new material was added. After the refactor, all tests still pass. 
Also, the SuperGLUE application code functions properly after this refactor.

### DIFF
--- a/snorkel/mtl/loggers/checkpointer.py
+++ b/snorkel/mtl/loggers/checkpointer.py
@@ -85,7 +85,7 @@ class Checkpointer(object):
         checkpoint_dir = f"{self.checkpoint_dir}/checkpoint_{iteration}.pth"
         torch.save(state_dict, checkpoint_dir)
         logging.info(
-            f"Save checkpoint of {iteration} {self.checkpoint_unit} "
+            f"Save checkpoint at {iteration} {self.checkpoint_unit} "
             f"at {checkpoint_dir}."
         )
 
--- a/snorkel/mtl/loggers/checkpointer.py
+++ b/snorkel/mtl/loggers/checkpointer.py
@@ -6,23 +6,33 @@ from typing import Dict, List
 
 import torch
 
-from snorkel.types import Config
+from snorkel.mtl.snorkel_config import default_config
+from snorkel.mtl.utils import recursive_merge_dicts
 
 
 class Checkpointer(object):
     """Checkpointing Training logging class to log train infomation"""
 
-    def __init__(self, config: Config):
+    def __init__(self, **kwargs):
 
-        self.logging_config = config
-        self.config = config["checkpointer_config"]
+        # Checkpointer requires both checkpointer_config and log_manager_config
+        checkpointer_config = recursive_merge_dicts(
+            default_config["checkpointer_config"],
+            default_config["log_manager_config"],
+            misses="insert",
+        )
+        self.config = recursive_merge_dicts(checkpointer_config, kwargs)
 
         # Pull out checkpoint settings
-        self.checkpoint_unit = self.logging_config["counter_unit"]
         self.checkpoint_dir = self.config["checkpoint_dir"]
+        self.checkpoint_unit = self.config["counter_unit"]
         self.checkpoint_clear = self.config["checkpoint_clear"]
         self.checkpoint_runway = self.config["checkpoint_runway"]
+        self.checkpoint_factor = self.config["checkpoint_factor"]
         self.checkpoint_condition_met = False
+
+        if self.checkpoint_dir is None:
+            raise Exception("Checkpointing is on but no checkpoint_dir was specified.")
 
         # Collect all metrics to checkpoint
         self.checkpoint_metric = self._make_metric_map(
@@ -39,7 +49,7 @@ class Checkpointer(object):
 
         # Set checkpoint frequency
         self.checkpoint_freq = (
-            self.logging_config["evaluation_freq"] * self.config["checkpoint_factor"]
+            self.config["evaluation_freq"] * self.config["checkpoint_factor"]
         )
         if self.checkpoint_freq <= 0:
             raise ValueError(
--- a/snorkel/mtl/loggers/checkpointer.py
+++ b/snorkel/mtl/loggers/checkpointer.py
@@ -16,11 +16,8 @@ class Checkpointer(object):
     def __init__(self, **kwargs):
 
         # Checkpointer requires both checkpointer_config and log_manager_config
-        checkpointer_config = recursive_merge_dicts(
-            default_config["checkpointer_config"],
-            default_config["log_manager_config"],
-            misses="insert",
-        )
+        checkpointer_config = default_config["checkpointer_config"]
+        checkpointer_config.update(default_config["log_manager_config"])
         self.config = recursive_merge_dicts(checkpointer_config, kwargs)
 
         # Pull out checkpoint settings
@@ -32,7 +29,7 @@ class Checkpointer(object):
         self.checkpoint_condition_met = False
 
         if self.checkpoint_dir is None:
-            raise Exception("Checkpointing is on but no checkpoint_dir was specified.")
+            raise ValueError("Checkpointing is on but no checkpoint_dir was specified.")
 
         # Collect all metrics to checkpoint
         self.checkpoint_metric = self._make_metric_map(
@@ -48,9 +45,7 @@ class Checkpointer(object):
             os.makedirs(self.checkpoint_dir)
 
         # Set checkpoint frequency
-        self.checkpoint_freq = (
-            self.config["evaluation_freq"] * self.config["checkpoint_factor"]
-        )
+        self.checkpoint_freq = self.config["evaluation_freq"] * self.checkpoint_factor
         if self.checkpoint_freq <= 0:
             raise ValueError(
                 f"Invalid checkpoint freq {self.checkpoint_freq}, "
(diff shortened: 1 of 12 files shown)


snorkel/mtl/loggers/checkpointer.py
```diff
@@ -6,24 +6,34 @@
...
-    def __init__(self, config: Config):
+    def __init__(self, **kwargs):
 
-        self.logging_config = config
-        self.config = config["checkpointer_config"]
+        # Checkpointer requires both checkpointer_config and log_manager_config
+        checkpointer_config = recursive_merge_dicts(
+            default_config["checkpointer_config"],
+            default_config["log_manager_config"],
+            misses="insert",
```
henryre: Note: this should be an enum

snorkel/mtl/loggers/checkpointer.py
```diff
@@ -6,24 +6,34 @@
...
 class Checkpointer(object):
     """Checkpointing Training logging class to log train infomation"""
 
-    def __init__(self, config: Config):
+    def __init__(self, **kwargs):
 
-        self.logging_config = config
-        self.config = config["checkpointer_config"]
+        # Checkpointer requires both checkpointer_config and log_manager_config
+        checkpointer_config = recursive_merge_dicts(
```
henryre: Let's compute this elsewhere since it's deterministic

snorkel/mtl/loggers/checkpointer.py
```diff
@@ -6,24 +6,34 @@
...
-        self.checkpoint_unit = self.logging_config["counter_unit"]
         self.checkpoint_dir = self.config["checkpoint_dir"]
+        self.checkpoint_unit = self.config["counter_unit"]
         self.checkpoint_clear = self.config["checkpoint_clear"]
         self.checkpoint_runway = self.config["checkpoint_runway"]
+        self.checkpoint_factor = self.config["checkpoint_factor"]
         self.checkpoint_condition_met = False
 
+        if self.checkpoint_dir is None:
+            raise Exception("Checkpointing is on but no checkpoint_dir was specified.")
```
henryre: ValueError

snorkel/mtl/loggers/log_manager.py
```diff
@@ -37,17 +13,17 @@ class LogManager(object):
     :type verbose: bool
     """
 
-    def __init__(self, n_batches_per_epoch: int, **kwargs) -> None:
+    def __init__(
+        self, n_batches_per_epoch: int, log_writer=None, checkpointer=None, **kwargs
```
henryre: **kwargs: Any

snorkel/mtl/loggers/checkpointer.py
```diff
@@ -39,7 +49,7 @@ def __init__(self, config: Config):
 
         # Set checkpoint frequency
         self.checkpoint_freq = (
-            self.logging_config["evaluation_freq"] * self.config["checkpoint_factor"]
+            self.config["evaluation_freq"] * self.config["checkpoint_factor"]
```
henryre: Nit: Use `self.` properties

snorkel/mtl/loggers/log_manager.py
```diff
@@ -114,9 +86,9 @@ def trigger_evaluation(self) -> bool:
 
     def trigger_checkpointing(self) -> bool:
         """Check if triggers the checkpointing"""
-        if not self.checkpointing:
+        if self.checkpointer is None:
             return False
-        satisfied = self.trigger_count >= self.checkpointing_freq
+        satisfied = self.trigger_count >= self.checkpointer.checkpoint_factor
```
henryre: Why is this changing now?

snorkel/mtl/loggers/log_writer.py
```diff
@@ -10,25 +14,48 @@ class LogWriter:
     :type object: [type]
```
henryre: Get rid of this plz

snorkel/mtl/loggers/log_writer.py
```diff
@@ -10,25 +14,48 @@ class LogWriter:
...
+        """Dump user-provided text to filename (e.g., the launch command)"""
+        text_path = os.path.join(self.log_dir, filename)
+        with open(text_path, "w") as f:
+            f.write(text)
 
-    def write_log(self, log_filename="log.json"):
+    def write_json(self, dict_to_write, filename):
         """Dump the log to file"""
-        log_path = os.path.join(self.log_dir, log_filename)
+        if not filename.endswith(".json"):
```
henryre: Is this needed? This kind of unexpected behavior to offer a very mild convenience always scares me.

snorkel/mtl/loggers/checkpointer.py
```diff
@@ -6,24 +6,34 @@
...
-from snorkel.types import Config
+from snorkel.mtl.snorkel_config import default_config
+from snorkel.mtl.utils import recursive_merge_dicts
 
 
 class Checkpointer(object):
     """Checkpointing Training logging class to log train infomation"""
 
-    def __init__(self, config: Config):
+    def __init__(self, **kwargs):
```
henryre: What do we gain here by all of these `__init__(self, **kwargs)` vs. an `__init__` with properly types parameters, and collecting them into a dictionary? It seems like a convenience for developers at the expense of allowing users to pass in incorrect named or typed parameters. And without a docstring, I have no clue what could possible be passed in.

snorkel/mtl/loggers/tensorboard_writer.py
```diff
@@ -19,7 +19,7 @@ def add_scalar(self, name, value, step):
     def write_config(self, config: Config, config_filename: str = "config.json"):
         """Dump the config to file"""
         super().write_config(config, config_filename)
-        self.writer.add_text(tag="config", text_string=config)
+        self.writer.add_text(tag="config", text_string=f"{config}")
```
henryre: `str(config)`?

snorkel/mtl/model.py
```diff
@@ -25,12 +20,12 @@ class MultitaskModel(nn.Module):
     :param tasks: a list of Tasks to be trained jointly
     """
 
-    def __init__(self, name: str, tasks: List[Task], **kwargs) -> None:
+    def __init__(self, tasks: List[Task], name=None, **kwargs) -> None:
```
henryre: `name: Optional[str], **kwargs: Any`

No partial typing plz!

snorkel/mtl/snorkel_config.py
```diff
@@ -0,0 +1,50 @@
+default_config = {
```
henryre: Sooooo why isn't this a JSON?

bhancock8:
Regarding config/__init__:

Current Design:
- Defaults are in master config
- Any kwargs you pass overwrite the defaults and get merged into self.config 
- All kwarg properties used in the class are accessed from self.config
- All config entries are automatically converted into flag options for run script

Pros:
- All settings in one place (read/write single config to define/log the whole run)
- Defaults are only in one place (not in config and function signature and possibly intermediate steps)
- You can easily pass bundles of subconfigs (i.e., Trainer initializes `optimizer`, `checkpointer`, etc. and can pass each one its corresponding nested dict--`optimizer_config`, `checkpointer_config`--from the main config)
- With recursive_merge_dicts, the user doesn’t need to pass in nested flags (e.g., to update momentum, they can just pass in `momentum=0.99` as a kwarg, rather than something like `optimizer_config={"sgd_config": {"momentum": 0.001}}`

Cons:
- You can’t see what the kwargs are for a given class (you're right--easier for developers, harder for users)
- Potential for users to pass in invalid kwargs (which currently yield warning statements, but don’t fail)

Suggestion:
- Leave this for now since it touches a _lot_ of things, but then post-workshop rework how we handle configs to be the best of both worlds (already noted in Asana).
- Classes have their kwargs and types defined in their own signature, but we automatically generate complete config by pulling out those values from the list of classes
- Can think some more on how to maintain the ability of not having to pass nested dicts as kwargs

snorkel/mtl/loggers/log_manager.py
bhancock8: Previously log_manager pulled out `checkpoint_factor` and called it `checkpoint_freq` (a mistake). Now it just pulls the value directly from the checkpointer. So the value is the same as before, it's just being called the right thing now.

snorkel/mtl/snorkel_config.py
bhancock8: So that it can be directly imported easily instead of having to use a json reader each time, and so there's no question about how types are being converted.

snorkel/mtl/loggers/log_writer.py
bhancock8: sg. Replaced it with a warning instead.

snorkel/mtl/loggers/checkpointer.py
bhancock8: See comment below. Discussed offline.

bhancock8:
Comments addressed! Ready for another look @henryre.

PR #1151, opened 2019-06-21
Add SimpleModel

bhancock8:
This PR demos how we could create a simple end model as a special case of our more powerful multi-task model. 

Example construction:
```
from snorkel.mtl.simple_model import SimpleModel
modules = [
    nn.Linear(2, 10), 
    nn.Linear(10, 2),
]
model = SimpleModel(modules)
```

The SimpleModelDemo.ipynb demonstrates its usage through training.
Note that if the user is using vanilla Datasets/DataLoaders, we'll need to upgrade those too, though that can be done automatically (probably with a check in the trainer?).

The only thing the model requires for construction is a list of modules. These are stitched together with nonlinearities and (optional) dropout between them. We also require that the data and labels both be only a single field.

The user can also change the loss_func (default cross-entropy) or output_func (default softmax), turn on dropout, and change the default metrics that are calculated (default "accuracy").

If we keep this design, I vote we name them something like `EndModel` and `FlexEndModel` (since the other model is more powerful in a number of ways---more fields, weirder flows, multiple tasks, etc.).


ajratner:
@bhancock8 can review soon but note github has a feature where you can submit a "draft" PR :)!

snorkel/mtl/simple_model.py
```diff
@@ -0,0 +1,70 @@
+"""A proof-of-concept that we can create a simple single-task model interface while
+primarily using the MultitaskModel class.
+"""
+
+from functools import partial
+from typing import Callable, List
+
```
paroma: `ERROR: /home/travis/build/HazyResearch/snorkel/snorkel/mtl/simple_model.py Imports are incorrectly sorted.`

PR #1152, opened 2019-06-21
Add slicing functionality

bhancock8:
Add slicing functionality.

SF class, decorator, and appliers are all mostly aliases of their LF counterparts (with the exception that they restrict the label space to 0, 1).

To use slicing, users need to:
1. Generate a slice_labels matrix with an applier
2. Update their dataloaders with the slice labels using add_slice_labels()
3. Convert their task to a set of slice tasks using convert_to_slice_tasks()

**Test plan**
test_slicing.py confirms that the slice model can be built and the overall task head as well as the slice indicator and slice predictor heads learn.


bhancock8:
Pause on review for a moment. Will ping again when ready.

bhancock8:
Ready to go again.

snorkel/slicing/sf.py
```diff
@@ -0,0 +1,22 @@
+"""Define SF appliers as simple alias of LF counterparts for now"""
+from typing import Callable
+
+from snorkel.labeling.lf import LabelingFunction, labeling_function
+
+
+class SlicingFunction(LabelingFunction):
```
henryre: label_space is unused right now, would suggest a direct alias for both again. just for simplicity

test/mtl/test_trainer.py
```diff
@@ -30,21 +29,22 @@ def test_trainer_onetask(self):
...
         self.assertLess(scores["task1/TestData/test/accuracy"], 0.7)
-        trainer = Trainer(**self.trainer_config, **self.logger_config)
+        trainer = Trainer(**self.trainer_config)
         trainer.train_model(model, dataloaders)
         scores = model.score(dataloaders)
         self.assertGreater(scores["task1/TestData/test/accuracy"], 0.9)
 
     def test_trainer_twotask(self):
         """Train a model with overlapping modules and flows"""
+        set_seed(SEED)
```
henryre: Does this not pass if the seed isn't the same? Probably not a valid unit test in that case.

test/slicing/test_slicing.py
```diff
@@ -0,0 +1,172 @@
+import random
+import unittest
+from functools import partial
+
+import numpy as np
+import pandas as pd
+import torch
+import torch.nn as nn
+
+from snorkel.mtl.data import MultitaskDataLoader, MultitaskDataset
```
henryre: Let's do some thinking later re: simplification (add a task please). Too many intra-snorkel deps for a unit test

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+
+from .modules.slice_combiner import SliceCombinerModule
+
+
+def add_slice_labels(
+    base_task: Task,
+    dataloader: MultitaskDataLoader,
+    slice_labels: csr_matrix,
+    slice_names: List[str],
+):
```
henryre: Return type?

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    slice_labels = slice_labels.toarray()
+    slice_labels, slice_names = _add_base_slice(slice_labels, slice_names)
```
henryre: assert slice_labels and slice_names are the same len

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+        pred_labels = indicators * labels
+
+        ind_task_name = f"{base_task.name}_slice:{slice_name}_ind"
+        pred_task_name = f"{base_task.name}_slice:{slice_name}_pred"
+
+        # Update dataloaders
+        dataloader.dataset.Y_dict[ind_task_name] = ind_labels
+        dataloader.dataset.Y_dict[pred_task_name] = pred_labels
+
+        dataloader.task_to_label_dict[ind_task_name] = ind_task_name
```
vincentschen: Open q: could we simplify? Is there ever a case where we _don't_ want to map a task to a labelset of the same name? Maybe the base task?

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+        dataloader.task_to_label_dict[pred_task_name] = pred_task_name
+    return dataloader
+
+
+def convert_to_slice_tasks(base_task: Task, slice_names: List[str]):
```
henryre: Bit hard for me to parse. Maybe @vincentschen is a better reviewer here.

Also good to get an isolated unit test for it. But can add later

snorkel/slicing/utils.py
henryre: It looks like it's different for the base task right now

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    """Adds slice labels to dataloader and creates new slice tasks (including base slice)
+
+    Each slice will get two slice-specific heads:
+    - an indicator head that learns to identify when DataPoints are in that slice
+    - a predictor head that is trained on only members of that slice
+
+    The base task's head is replaced by a master head that makes predictions based on
+    a combination of the predictor heads' predictions that are weighted by the
```
vincentschen: nit: to be more precise, "prediction heads' predictions" >> "prediction heads' prediction confidences"

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    indicator heads' predictions.
+
+    NOTE: The current implementation pollutes the module_pool---the indicator's task's
```
vincentschen: typo: indicator task's

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    extra copies of (potentially very large) modules that will be merged in a moment
+    away in the model since they have the same name. We leave resolution of this issue
+    for a future release.
+    """
+    if "base" not in slice_names:
+        slice_names = slice_names + ["base"]
+
+    tasks = []
+
+    # NOTE: We assume here that the last operation uses the head module
```
vincentschen: is there a way to validate this? with naming or a dimension check?

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    del base_task.module_pool[head_module_op.module_name]
+    base_task.task_flow = base_task.task_flow[:-1]
+
+    # Create slice indicator tasks
+    for slice_name in slice_names:
+
+        ind_task_name = f"{base_task.name}_slice:{slice_name}_ind"
+
+        ind_head_module_name = f"{ind_task_name}_ind_head"
+        ind_head_module = nn.Linear(neck_size, 2)
```
vincentschen: warning here? `ASSUMPTION: binary classification`

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+    head_module_op = base_task.task_flow[-1]
+    head_module = base_task.module_pool[head_module_op.module_name]
+
+    if isinstance(head_module, nn.DataParallel):
+        head_module = head_module.module
+
+    neck_size = head_module.in_features
+    base_task_cardinality = head_module.out_features
+
+    # Remove the slice-unaware head module from module pool and task flow
```
vincentschen: I got confused about this when I read line 101—a quick glance made think the final prediction head was still intact. 

could we rename the base task flow? maybe as `body_flow` or something to indicate we have decapitated the original task

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+        ind_task_flow = base_task.task_flow + [ind_head_op]
+
+        # Create ind task
+        ind_task = Task(
+            name=ind_task_name,
+            module_pool=ind_module_pool,
+            task_flow=ind_task_flow,
+            loss_func=partial(ce_loss, ind_head_op.name),
+            output_func=partial(softmax, ind_head_op.name),
+            scorer=Scorer(metrics=["accuracy"]),
```
vincentschen: let's add `f1` by default here—usually indicators have class balance issues

snorkel/slicing/utils.py
```diff
@@ -0,0 +1,202 @@
...
+        pred_module_pool[pred_head_module_name] = shared_pred_head_module
+
+        # Create task_flow
+        pred_transform_op = Operation(
+            module_name=pred_transform_module_name, inputs=head_module_op.inputs
+        )
+        pred_head_op = Operation(
+            module_name=pred_head_module_name, inputs=[(pred_transform_op.name, 0)]
+        )
+        pred_task_flow = base_task.task_flow + [pred_transform_op, pred_head_op]
```
vincentschen: same as above—temporarily confusing that we're still calling the decapitated base_task flow by the same name

test/slicing/test_slicing.py
```diff
@@ -0,0 +1,172 @@
...
+        self.assertGreater(scores["task1_slice:f_pred/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task1_slice:g_ind/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task1_slice:g_pred/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task1_slice:base_ind/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task1_slice:base_pred/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task2/TestData/valid/accuracy"], 0.9)
+        self.assertGreater(scores["task2/TestData/valid/accuracy"], 0.9)
+
+
+def set_seed(seed):
```
vincentschen: can call `mtl.utils.set_seed` now!

snorkel/slicing/apply/sf_applier.py
```diff
@@ -0,0 +1,7 @@
+from snorkel.labeling.apply import LFApplier
+
+
+class SFApplier(LFApplier):
```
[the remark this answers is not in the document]
bhancock8: But we want it to have a different name, yes? Like when you print it and it reports its class, we don't want it report that it's an LFApplier object.

snorkel/mtl/task.py
```diff
@@ -1,12 +1,27 @@
...
-from typing import Any, Callable, Dict, List
+from typing import Callable, List
 
 import torch
 from torch import nn
 
 from snorkel.mtl.scorer import Scorer
 
 
+class Operation:
```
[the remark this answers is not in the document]
bhancock8: Can't use a namedtuple because of the logic where the name defaults to the module_name. And @dataclass didn't come until Python 3.7.

snorkel/slicing/utils.py
bhancock8: Good Q. We don't currently place any restrictions on what modules are named (i.e., not required to include "head" in them).

snorkel/slicing/utils.py
bhancock8: Yeah, I was hesitant because the metrics dict quickly balloons in size, but from a practical standpoint, it is nice to have it. Adding.

vincentschen: if we want to reduce size, i think f1 is preferable to accuracy, fwiw

snorkel/slicing/utils.py
bhancock8: Yeah, you can have gold and silver standard labels, for example. We could make some nice defaults, but will save for a future PR.

test/slicing/test_slicing.py
bhancock8: Yeah, this is more of an integration test. Just converted the minimal script I was using. Agreed that we should add some finer grained tests. Added to the tracker.

snorkel/slicing/utils.py
bhancock8: Done!

test/mtl/test_trainer.py
bhancock8: Yeah, there's variation. Will address when we make add real unit tests for slicing.

PR #1153, opened 2019-06-21
Add augmentation module

henryre:
Primary changes:
* TF aliases for mappers
* RandomAugmentationPolicy (DataPoint -> DataPoint) which takes in TFs and applies them at random (for a seq length of k) to a DataPoint
* TFApplier (List[SimpleNamespace] -> List[SimpleNamespace]) and PandasTFApplier (pd.DF -> pd.DF) that takes in a policy, a number of transformed data points per original, and keep_original as a flag

Example usage:

```python
import pandas as pd
from snorkel.augmentation.apply import PandasTFApplier
from snorkel.augmentation.policy import RandomAugmentationPolicy
from snorkel.augmentation.tf import transformation_function
from snorkel.types import FieldMap

df = pd.DataFrame(dict(num=[1, 2, 3]))

@transformation_function
def square(num: int) -> FieldMap:
    return dict(num=num ** 2)

@transformation_function
def double(num: int) -> FieldMap:
    return dict(num=2 * num)

policy = RandomAugmentationPolicy([square, double], sequence_length=4)

applier = PandasTFApplier(policy, k=2, keep_original=True)
df_augmented = applier.apply(df)
```

Other changes:
* Add `__init__.py` files to test dir so that they don't have to be imported in the global namespace by pytest
* Fixes a bug for Mapper in NAMESPACE mode where the underlying `SimpleNamespace.__dict__` was getting modified in-place
* Allow people to pass in positional args to pytest when calling tox, e.g. `tox -e py36 -- -k test_tf_applier_pandas`

**Test plan**
Unit tests for TFAppliers and RandomAugmentationPolicy

### DIFF
--- a/snorkel/augmentation/apply/tf_applier.py
+++ b/snorkel/augmentation/apply/tf_applier.py
@@ -1,0 +1,39 @@
+from typing import Any, List
+
+from tqdm import tqdm
+
+from snorkel.augmentation.policy import AugmentationPolicy
+from snorkel.augmentation.tf import TransformationFunctionMode
+from snorkel.types import DataPoint
+
+
+class BaseTFApplier:
+    def __init__(
+        self, policy: AugmentationPolicy, k: int = 1, keep_original: bool = True
+    ) -> None:
+        self._policy = policy
+        self._k = k
+        self._keep_original = keep_original
+
+    def _set_tf_mode(self, mode: TransformationFunctionMode) -> None:
+        self._policy.set_tf_mode(mode)
+
+    def _apply_policy_to_data_point(self, x: DataPoint) -> List[DataPoint]:
+        x_transformed = []
+        if self._keep_original:
+            x_transformed.append(x)
+        for _ in range(self._k):
+            x_transformed.append(self._policy.apply(x))
+        return x_transformed
+
+    def apply(self, data_points: Any, *args: Any, **kwargs: Any) -> Any:
+        raise NotImplementedError
+
+
+class TFApplier(BaseTFApplier):
+    def apply(self, data_points: List[DataPoint]) -> List[DataPoint]:  # type: ignore
+        self._set_tf_mode(TransformationFunctionMode.NAMESPACE)
+        x_transformed = []
+        for x in tqdm(data_points):
+            x_transformed.extend(self._apply_policy_to_data_point(x))
+        return x_transformed
--- a/snorkel/augmentation/apply/tf_applier.py
+++ b/snorkel/augmentation/apply/tf_applier.py
@@ -3,27 +3,37 @@ from typing import Any, List
 from tqdm import tqdm
 
 from snorkel.augmentation.policy import AugmentationPolicy
-from snorkel.augmentation.tf import TransformationFunctionMode
+from snorkel.augmentation.tf import TransformationFunction, TransformationFunctionMode
 from snorkel.types import DataPoint
 
 
 class BaseTFApplier:
     def __init__(
-        self, policy: AugmentationPolicy, k: int = 1, keep_original: bool = True
+        self,
+        tfs: List[TransformationFunction],
+        policy: AugmentationPolicy,
+        k: int = 1,
+        keep_original: bool = True,
     ) -> None:
+        self._tfs = tfs
         self._policy = policy
         self._k = k
         self._keep_original = keep_original
 
     def _set_tf_mode(self, mode: TransformationFunctionMode) -> None:
-        self._policy.set_tf_mode(mode)
+        for tf in self._tfs:
+            tf.set_mode(mode)
 
     def _apply_policy_to_data_point(self, x: DataPoint) -> List[DataPoint]:
         x_transformed = []
         if self._keep_original:
             x_transformed.append(x)
         for _ in range(self._k):
-            x_transformed.append(self._policy.apply(x))
+            x_t = x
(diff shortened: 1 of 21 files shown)


henryre:
Offline:

* Updating signature to 
```
@lambda_mapper
def my_function(x: DataPoint) -> DataPoint:
   ...
```
in follow up PR
* Subclassing Mapper so that TFs can yield multiple transformations

ajratner:
@henryre thoughts on making the `TFApplier` return a generator, or at least having this option?  This is more in keeping with how data aug usually applied

ajratner:
Or: add a kwarg that determines whether the TF applier returns a dataset or a generator, e.g. `materialize=true|false`

henryre:
@ajratner agreed, adding in a followup

snorkel/augmentation/apply/tf_applier.py
```diff
@@ -0,0 +1,39 @@
...
+from snorkel.augmentation.policy import AugmentationPolicy
+from snorkel.augmentation.tf import TransformationFunctionMode
+from snorkel.types import DataPoint
+
+
+class BaseTFApplier:
+    def __init__(
+        self, policy: AugmentationPolicy, k: int = 1, keep_original: bool = True
+    ) -> None:
+        self._policy = policy
```
ajratner: Any thoughts on accepting either an `AugmentationPolicy` or a `TransformationFunction`, and in latter case defaulting to a trivial policy?

snorkel/augmentation/policy/core.py
```diff
@@ -0,0 +1,31 @@
...
+        self, tfs: List[TransformationFunction], sequence_length: int = 1
+    ) -> None:
+        self._k = sequence_length
+        super().__init__(tfs)
+
+    def apply(self, x: DataPoint) -> DataPoint:
+        for _ in range(self._k):
+            tf = random.choice(self._tfs)
+            x = tf(x)
+        return x
```
ajratner: Mainly cosmetic at this point but: can we add in the mean field policy (with user-provided weights) just to show what this would look like if someone peeks (ie hint in the direction of TANDA)?  Happy to add this in

snorkel/augmentation/policy/core.py
```diff
@@ -0,0 +1,31 @@
...
+class AugmentationPolicy:
+    def __init__(self, tfs: List[TransformationFunction]):
+        self._tfs = tfs
+
+    def set_tf_mode(self, mode: TransformationFunctionMode) -> None:
+        for tf in self._tfs:
+            tf.set_mode(mode)
+
+    def apply(self, x: DataPoint) -> DataPoint:
+        raise NotImplementedError
```
ajratner: Okay how I was originally thinking about this, and based on what "policy" usually means: thoughts on having the `AugmentationPolicy` class just yield an ordering / generator of TF ids to be applied, but leave the actual application of TFs to the TF applier (so e.g. it can be optimized, and factored out better)?

snorkel/map/core.py
```diff
@@ -74,6 +68,7 @@ def run(self, **kwargs: Any) -> FieldMap:
         raise NotImplementedError
 
     def __call__(self, x: DataPoint) -> DataPoint:
+        x = dill.loads(dill.dumps(x))
```
ajratner: Huh?

ajratner: At least add insider-focused comments for this sort of cryptic thing...

snorkel/augmentation/apply/tf_applier.py
henryre: A single TF is a fringe use-case, but I can add an ApplyAllPolicy which applies all TFs in order for an input list. Either way, let's just take a Policy here. Would trade convenience for clarity.

snorkel/augmentation/policy/core.py
henryre: All yours!

snorkel/augmentation/policy/core.py
henryre: Sounds good

snorkel/map/core.py
henryre: Will do. Just a more reliable way to deepcopy (e.g. calling deepcopy on a pd.Series or SimpleNamespace with a dictionary attribute won't dereference the dictionary)

snorkel/augmentation/apply/tf_applier.py
ajratner: LGTM; last thing- what about setting `RandomAugmentationPolicy` if that arg is `None`?

ajratner: Another reason for this- having to pass in `RandomAugmentationPolicy(n_tfs)` seems kind of extra clunky, bc specifying n_tfs (implicitly) twice...

(On the other hand, when you're instead providing a parameter to the policy class, this is obviously fine and no longer seems clunky)

ajratner:
@henryre Happy to finish up this PR including requested change if you want

snorkel/augmentation/apply/tf_applier.py
henryre: Offline: avoid guessing what user wants here, but come up with a less clunky way to define a simple policy over a list of TFs

paroma:
Add matplotlib into requirements or this `synthetic.py` does not run

PR #1157, opened 2019-06-23
[EASY] Standardize set_seed()

bhancock8:
Method set_seed() was added to mtl.utils and was redefined as a class method in a test and Classifier(). Now all use the version in snorkel.model.utils.


bhancock8:
Good call, formal type check now.

PR #1162, opened 2019-06-23
Add Multitask Tutorial

bhancock8:
The Multitask Tutorial walks through the construction and training of a multi-task network using the MultitaskModel.

**Test plan**
Notebook is saved with output from a fresh "restart and run all". 
Notebook runs to completion.


snorkel/mtl/snorkel_config.py
```diff
@@ -18,7 +18,7 @@
     "checkpointer_config": {  # Note that checkpointer behavior also depends on log_manager_config
         "checkpoint_dir": None,  # Trainer will set this to log_dir if None
         "checkpoint_factor": 1,  # Checkpoint every this many evaluations
-        "checkpoint_metric": "model/train/all/loss:min",
```
vincentschen: Hmm.. nervous that this seemed to work for me before. Maybe the defaults were ok?

PR #1167, opened 2019-06-27
Set up pydocstyle

henryre:
See title. We won't land until current docs pass.


henryre:
Ok I changed this so that we add a tox command and record pydocstyle settings without running in tox. This makes it easier to use VSCode pydocstyle tooling in the interim.

PR #1168, opened 2019-06-27
Add testing and typing to analysis.py

henryre:
See title. Primarily adding unit tests for analysis.py
Also:
* Handling of NaN outputs in `lf_empirical_accuracies`
* Trim poorly defined columns in `single_lf_summary`
* Replace `issparse` for type-ability
* Removed that goofball `global` since it's unused

**Test plan**
Added tests...obviously

### DIFF
--- a/test/labeling/test_analysis.py
+++ b/test/labeling/test_analysis.py
@@ -1,0 +1,171 @@
+import unittest
+
+import numpy as np
+import pandas as pd
+import scipy.sparse as sparse
+
+from snorkel.labeling.analysis import (
+    label_coverage,
+    label_overlap,
+    label_conflict,
+    lf_conflicts,
+    lf_coverages,
+    lf_empirical_accuracies,
+    lf_overlaps,
+    lf_polarities,
+    error_buckets,
+    lf_summary,
+    single_lf_summary,
+    confusion_matrix,
+)
+
+
+L = [
+    [0, 0, 1, 0, 0, 1],
+    [0, 0, 0, 3, 0, 0],
+    [3, 0, 0, 0, 0, 1],
+    [2, 0, 3, 0, 1, 1],
+    [0, 0, 0, 0, 0, 0],
+    [2, 0, 1, 3, 2, 1],
+]
+
+Y = [1, 2, 3, 1, 2, 3]
+
+
+class TestAnalysis(unittest.TestCase):
+    def setUp(self) -> None:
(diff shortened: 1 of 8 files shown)


tox.ini
```diff
@@ -31,7 +31,7 @@ commands = mypy -p snorkel
 description = run coverage checks
 basepython = python3
 usedevelop = True
-commands = pytest --cov=snorkel test/
+commands = python -m pytest --cov=snorkel test/
```
bhancock8: Why was this necessary to add?

test/labeling/test_analysis.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def test_label_overlap(self) -> None:
+        self.assertEqual(label_overlap(self.L), 4 / 6)
+
+    def test_label_conflict(self) -> None:
+        self.assertEqual(label_conflict(self.L), 3 / 6)
+
+    def test_lf_polarities(self) -> None:
+        polarities = lf_polarities(self.L)
+        self.assertEqual(polarities, [[2, 3], [], [1, 3], 3, [1, 2], 1])
```
bhancock8: lf_polarities should return a consistent type--probably always a list?

tox.ini
henryre: It wasn't strictly but just standardized the pytest call (see the py env above) when debugging the PyArrow+PyTorch issue

test/labeling/test_analysis.py
henryre: https://github.com/HazyResearch/snorkel/pull/1168/files#diff-5839d69014187bdf1a2f60ccee964363R73

bhancock8: Right, I see that the code _allows_ for both right now. But I think we should change it so that it only returns a List of Lists.

henryre: Ah I see, you're saying the lone ints are unexpected. I agree. Can change now.

henryre: Or add a flag?

bhancock8: The flag would only be applicable some of the time, since multi-polar LFs would still need to be a collection of some sort. I say just make it always a list of lists.

henryre: Cool on it

henryre: Will do in a stacked PR so as not to complicate rebases. Not really in scope for this diff either.

PR #1169, opened 2019-06-27
Remove PyArrow-based test

henryre:
Grrrr https://issues.apache.org/jira/browse/ARROW-3346

tl;dr importing PyTorch before PyArrow (depending on build) results in a segfault

Currently only using PyArrow in one completely unnecessary test, so we'll remove it for now, then figure out how to deal with Parquet once we have core tooling.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,6 @@ matplotlib==3.0.1
 networkx==2.2
 numpy==1.16.0
 pandas==0.24.0
-pyarrow==0.12.0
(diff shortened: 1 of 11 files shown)


henryre:
#1168 adds another torch import, which is the first one that could be randomly executed before `test_lf_applier.py`. So would randomly get a core dump when running `tox`. Took a bit of digging and https://github.com/pytest-dev/pytest-faulthandler

PR #1174, opened 2019-06-28
Speed up class balance test a bit

henryre:
Cut class balance test time from 24s -> 8s via some vectorization and cutting the number of test examples. This is very much a bandaid, we need to rethink these tests from the ground up.

**Test plan**
New unit test results. Spot checked output probabilities.

### DIFF
--- a/test/labeling/model/test_class_balance.py
+++ b/test/labeling/model/test_class_balance.py
@@ -7,73 +7,80 @@ from snorkel.labeling.model.class_balance import ClassBalanceModel
 from snorkel.model.utils import set_seed
 
 
+def generate_class_balance(k: int) -> np.ndarray:
+    """Generate class balance"""
+    p_Y = np.random.random(k)
+    p_Y /= p_Y.sum()
+    return p_Y
+
+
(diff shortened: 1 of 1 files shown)


test/labeling/model/test_class_balance.py
```diff
@@ -7,73 +7,80 @@
...
+
+def generate_cond_probs(
+    k: int, m: int, bias_diag: bool = True, abstains: bool = False
+) -> np.ndarray:
+    """Generate conditional probability tables for the m conditionally ind.
+    LFs, such that:
+        cpts[i, y1, y2] = P(\lambda_i = y1 | Y = y2)
+
+    Args:
+        k: (int) Number of classes
```
paroma: nit: k = number of classes vs. cardinality throughout documentation

bhancock8: Isn't the cardinality always just the number of classes though, so they're the same?

henryre:
@paroma lmk if I should merge this. Can also just wait for your rewrite.

PR #1176, opened 2019-06-28
Add docstrings for map module

henryre:
See title.

**Test plan**
`pydoctest snorkel/map`

### DIFF
--- a/snorkel/map/core.py
+++ b/snorkel/map/core.py
@@ -100,11 +100,10 @@ class Mapper(BaseMapper):
         A map from attribute names of the incoming data points
         to the input argument names of the `run` method. If None,
         the parameter names in the function signature are used.
-        By default None.
     mapped_field_names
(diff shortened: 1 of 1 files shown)


snorkel/map/core.py
```diff
@@ -41,33 +76,58 @@ def __call__(self, x: DataPoint) -> DataPoint:
...
 class Mapper(BaseMapper):
+    """Base class for any data point to data point mapping in the pipeline.
+
+    Map data points to new data points by transforming, adding
+    additional information, or decomposing into primitives. This module
+    provides base classes for other operators like `TransformationFunction`
+    and `Preprocessor`. We don't expect people to construct `Mapper`
+    objects directly.
+
+    A Mapper maps an data point to a new data point, possibly with
```
bhancock8: nit: an data point

henryre: Fixing in next diff

PR #1177, opened 2019-06-28
Add docstrings for augmentation module

henryre:
See title.

**Test plan**
`pydocstyle snorkel/augmentation`


henryre:
**Changes planned**: going to add some color on what a TF actually is

PR #1182, opened 2019-06-28
Add BaselineModel and refactor LabelModel tests

paroma:
Add tests for baseline label models
Refactor tests for LabelModel as checking intermediate steps


test/labeling/model/test_label_model.py
```diff
@@ -2,13 +2,12 @@
 
 import numpy as np
 
-from snorkel.labeling.model.baselines import MajorityLabelVoter
+# from snorkel.labeling.model.baselines import MajorityLabelVoter
```
henryre: No no

test/labeling/model/test_label_model.py
```diff
@@ -88,11 +91,42 @@ def test_augmented_L_construction(self):
         self.assertEqual(L_aug[1, j], 1)
         self.assertEqual(L_aug[2, j], 1)
 
-    def test_with_deps(self):
-        for seed in range(self.n_iters):
-            np.random.seed(seed)
-            data = SingleTaskTreeDepsGenerator(self.n, self.m, k=self.k, edge_prob=1.0)
-            self._test_label_model(data, test_acc=False)
+    # def test_no_deps(self):
```
henryre: No no no

test/labeling/model/test_baseline_model.py
```diff
@@ -0,0 +1,41 @@
...
+from snorkel.labeling.model.baselines import (
+    MajorityClassVoter,
+    MajorityLabelVoter,
+    RandomVoter,
+)
+
+
+class BaselineModelTest(unittest.TestCase):
+    def test_random_vote(self):
+        L = np.array([[1, 1, 3, 1], [2, 4, 1, 2], [1, 3, 1, 2]]).T
```
henryre: Probably want some 0s in all the cases (know this one doesn't matter)

henryre: Why are all the L's transposed?

test/labeling/model/test_baseline_model.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertGreaterEqual(Y_p.min(), 0.0)
+
+    def test_majority_class_vote(self):
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        mc_voter = MajorityClassVoter()
+        mc_voter.train_model(balance=[0.8, 0.2])
+        Y_p = mc_voter.predict_proba(L)
+
+        Y_p_true = np.array([[1.0, 0.0], [1.0, 0.0], [1.0, 0.0], [1.0, 0.0]])
+        self.assertTrue((Y_p == Y_p_true).all())
```
henryre: Use `np.testing.assertArrayAlmostEquals`

test/labeling/model/test_baseline_model.py
```diff
@@ -0,0 +1,41 @@
...
+    def test_random_vote(self):
+        L = np.array([[1, 1, 3, 1], [2, 4, 1, 2], [1, 3, 1, 2]]).T
+        rand_voter = RandomVoter()
+        rand_voter.train_model()
+        Y_p = rand_voter.predict_proba(L)
+        self.assertLessEqual(Y_p.max(), 1.0)
+        self.assertGreaterEqual(Y_p.min(), 0.0)
+
+    def test_majority_class_vote(self):
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
```
henryre: Check an example where `balance` and `L` have different majorities.

test/labeling/model/test_baseline_model.py
```diff
@@ -0,0 +1,41 @@
...
+        self.assertTrue((Y_p == Y_p_true).all())
+
+    def test_majority_label_vote(self):
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        ml_voter = MajorityLabelVoter()
+        ml_voter.train_model(balance=[0.8, 0.2])
+        Y_p = ml_voter.predict_proba(L)
+
+        Y_p_true = np.array([[1.0, 0.0], [1.0, 0.0], [1.0, 0.0], [0.0, 1.0]])
+        self.assertTrue((Y_p == Y_p_true).all())
```
henryre: Same

test/labeling/model/test_baseline_model.py
```diff
@@ -0,0 +1,41 @@
...
+        mc_voter.train_model(balance=[0.8, 0.2])
+        Y_p = mc_voter.predict_proba(L)
+
+        Y_p_true = np.array([[1.0, 0.0], [1.0, 0.0], [1.0, 0.0], [1.0, 0.0]])
+        self.assertTrue((Y_p == Y_p_true).all())
+
+    def test_majority_label_vote(self):
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        ml_voter = MajorityLabelVoter()
+        ml_voter.train_model(balance=[0.8, 0.2])
```
henryre: Unused

henryre: A little out-of-scope but baselines.py is a mess. Should be

```python
import numpy as np

from .label_model import LabelModel

class BaselineVoter(LabelModel):
    def train_model(self, *args, **kwargs):
        pass

class RandomVoter(BaselineVoter):
    """
    A class that votes randomly among the available labels
    """

    def predict_proba(self, L):
        """
        Args:
            L: An [n, m] scipy.sparse matrix of labels
        Returns:
            output: A [n, k] np.ndarray of probabilistic labels
        """
        n = L.shape[0]
        Y_p = np.random.rand(n, self.k)
        Y_p /= Y_p.sum(axis=1).reshape(-1, 1)
        return Y_p


class MajorityClassVoter(LabelModel):
    """
    A class that places all probability on the majority class based on class
    balance (and ignoring the label matrix).

    Note that in the case of ties, non-integer probabilities are possible.
    """

    def train_model(self, balance, *args, **kwargs):
        """
        Args:
            balance: A 1d arraylike that sums to 1, corresponding to the
                (possibly estimated) class balance.
        """
        self.balance = np.array(balance)

    def predict_proba(self, L):
        n = L.shape[0]
        Y_p = np.zeros((n, self.k))
        max_classes = np.where(self.balance == max(self.balance))
        for c in max_classes:
            Y_p[:, c] = 1.0
        Y_p /= Y_p.sum(axis=1).reshape(-1, 1)
        return Y_p


class MajorityLabelVoter(BaselineVoter):
    """
    A class that places all probability on the majority label from all
    non-abstaining LFs for that task.

    Note that in the case of ties, non-integer probabilities are possible.
    """

    def predict_proba(self, L):
        L = self._to_numpy(L).astype(int)
        n, m = L.shape
        Y_p = np.zeros((n, self.k))
        for i in range(n):
            counts = np.zeros(self.k)
            for j in range(m):
                if L[i, j]:
                    counts[L[i, j] - 1] += 1
            Y_p[i, :] = np.where(counts == max(counts), 1, 0)
        Y_p /= Y_p.sum(axis=1).reshape(-1, 1)
        return Y_p
```

henryre: MajorityClassVoter and MajorityLabelVoter are not "RandomVoters". And `train_model` inheritance made no sense.

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
-            deps=data.E,
-            class_balance=data.p,
-            n_epochs=1000,
-            log_train_every=200,
+    def test_model_init(self):
+        label_model = LabelModel(k=2, verbose=False)
+
+        # Test class balance
+        Y_dev = np.array([1, 1, 2, 2, 1, 1, 1, 1, 2, 2])
+        label_model._set_class_balance(class_balance=None, Y_dev=Y_dev)
```
henryre: Not to pontificate, but unit testing internal state via private methods is contentious: https://stackoverflow.com/questions/34571/how-do-i-test-a-private-function-or-a-class-that-has-private-methods-fields-or

It might be ok to do this here. Honestly, these methods can likely be made public

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
+        self.assertTrue((label_model.p == np.array([0.6, 0.4])).all())
+
+        # Test dimension constants
```
henryre: Why is this in the same test? Do we need the class balance set already?

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
-            data.L,
-            deps=data.E,
-            class_balance=data.p,
-            n_epochs=1000,
-            log_train_every=200,
+    def test_model_init(self):
+        label_model = LabelModel(k=2, verbose=False)
+
+        # Test class balance
+        Y_dev = np.array([1, 1, 2, 2, 1, 1, 1, 1, 2, 2])
```
henryre: zeros plz!!

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
-            mv = MajorityLabelVoter()
-            mv_score = mv.score((data.L, data.Y), verbose=False)
-            self.assertGreater(score, mv_score)
-
-    def test_no_deps(self):
-        for seed in range(self.n_iters):
-            np.random.seed(seed)
-            data = SingleTaskTreeDepsGenerator(self.n, self.m, k=self.k, edge_prob=0.0)
-            self._test_label_model(data)
+        self.assertTrue((label_model.O.numpy() == true_O).all())
```
henryre: `numpy.testing`

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
+        # Test dimension constants
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        label_model._set_constants(L)
+        self.assertEqual(label_model.n, 4)
+        self.assertEqual(label_model.m, 3)
+
+    def test_generate_O(self):
+        label_model = LabelModel(k=2, verbose=False)
+
+        # Test inverse) from L
```
henryre: nit: typo

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        label_model._set_constants(L)
+        label_model._set_dependencies(deps=[])
+        label_model._generate_O(L)
+
+        true_O = np.array(
```
henryre: How did we compute this? Fractions (e.g. 3 / 4) might be more legible if you computed them by hand.

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +16,40 @@ def setUpClass(cls):
...
+        self.assertEqual(label_model.m, 3)
+
+    def test_generate_O(self):
+        label_model = LabelModel(k=2, verbose=False)
+
+        # Test inverse) from L
+        L = np.array([[1, 1, 2, 1], [2, 2, 1, 2], [1, 1, 1, 2]]).T
+        label_model._set_constants(L)
+        label_model._set_dependencies(deps=[])
+        label_model._generate_O(L)
```
henryre: Should `generate_O()` execute this chain itself?

paroma: chain as in the set constant and set dependencies? nope

test/labeling/model/test_label_model.py
paroma: this is GT labels

test/labeling/model/test_label_model.py
paroma: I think making these methods public is the way to go - everything is private except train, which makes it hard to test

test/labeling/model/test_label_model.py
paroma: good catch, separating

test/labeling/model/test_label_model.py
```diff
@@ -17,36 +15,43 @@ def setUpClass(cls):
...
+        label_model = LabelModel(k=2, verbose=False)
+
+        # Test inverse from L
+        L = np.array([[1, 2, 1], [1, 2, 1], [2, 1, 1], [1, 2, 2]])
+        label_model._set_constants(L)
+        label_model._set_dependencies(deps=[])
+        label_model._generate_O(L)
+
+        true_O = np.array(
+            [
```
henryre: Nit: don't need to be floats

test/labeling/model/test_label_model.py
```diff
@@ -88,11 +93,141 @@ def test_augmented_L_construction(self):
...
+    def test_conditional_probs(self):
+        label_model = LabelModel(k=2, verbose=False)
+
+        L = np.array([[1, 2, 1], [1, 2, 1]])
+        label_model._set_constants(L)
+        label_model._set_dependencies(deps=[])
+        label_model._generate_O(L)
+        label_model.inv_form = False
+        label_model._set_class_balance(class_balance=[0.6, 0.4], Y_dev=None)
+        label_model._init_params()
```
henryre:
```python
        label_model._set_constants(L)
        label_model._set_dependencies(deps=[])
        label_model._generate_O(L)
        label_model.inv_form = False
        label_model._set_class_balance(class_balance=[0.6, 0.4], Y_dev=None)
        label_model._init_params()
```

is repeated quite a bit (or some form of it). Can we abstract that out to an internal function?

bhancock8:
We (@paroma, @vincentschen , @henryre, and I) discussed something related to this last week; we wanted to make sure our tutorials stay in a working state, but running them all on every commit would be too slow. We landed on having a snorkel-tutorials repo (like pytorch) that can be pinned to a particular commit and then when changes are made runs those slower tutorials. Sounds like you want some fatter tests for the core code though.

paroma:
@ajratner we're trying to test every function that `train()` and `predict_proba()` would call and like @bhancock8 said, the snorkel-tutorials repo can have the end to end label model testing grounded in the tutorial examples

also some of these will change with the LabelModel refactor, so we can discuss how to lay this out then!

snorkel/labeling/model/label_model.py
```diff
@@ -320,19 +307,11 @@ def loss_l2(self, l2=0):
         if isinstance(l2, (int, float)):
             D = l2 * torch.eye(self.d)
         else:
-            D = torch.diag(torch.from_numpy(l2))
+            D = torch.diag(torch.from_numpy(l2)).type(torch.FloatTensor)
```
henryre: shouldn't this be a type? i.e. torch.float32 here?

test/labeling/model/test_label_model.py
```diff
@@ -88,11 +100,155 @@ def test_augmented_L_construction(self):
...
+
+    def test_inv_loss(self):
+        L = np.array([[1, 0, 1], [1, 2, 1]])
+        label_model = self._set_up_model(L, deps=[(1, 2)])
+        label_model.inv_form = True
+
+        label_model.Z = nn.Parameter(torch.ones(label_model.d, label_model.k)).float()
+
+        Q = label_model.get_Q()
+        self.assertAlmostEqual(Q[0, 0], 0.89285714)
```
henryre: Fewer decimals here? Did you compute this by hand or just see what an execution of `get_Q()` spat out?

test/labeling/model/test_label_model.py
```diff
@@ -88,11 +100,155 @@ def test_augmented_L_construction(self):
...
+        L = np.array([[1, 0, 1], [1, 2, 1]])
+        label_model = self._set_up_model(L)
+
+        label_model.train_model(L, n_epochs=1)
+        init_loss = label_model.loss_mu().detach().numpy().ravel()[0]
+
+        label_model.train_model(L, n_epochs=10)
+        next_loss = label_model.loss_mu().detach().numpy().ravel()[0]
+
+        self.assertLessEqual(next_loss, init_loss)
```
henryre: Couldn't this depend on e.g. default learning rate? Why do we have to see a loss decrease on this particular L? (we might have to, idk)

test/labeling/model/test_label_model.py
```diff
@@ -88,11 +100,155 @@ def test_augmented_L_construction(self):
...
+
+        label_model.mu = nn.Parameter(label_model.mu_init.clone() + 0.05)
+        self.assertAlmostEqual(
+            label_model.loss_l2(l2=1.0).detach().numpy().ravel()[0], 0.03
+        )
+        self.assertAlmostEqual(
+            label_model.loss_l2(l2=np.ones(6)).detach().numpy().ravel()[0], 0.03
+        )
+        self.assertAlmostEqual(
+            label_model.loss_mu().detach().numpy().ravel()[0], 0.6751751
```
henryre: Same here

test/labeling/model/test_label_model.py
```diff
@@ -1,52 +1,64 @@
...
-            n_epochs=1000,
-            log_train_every=200,
+    def _set_up_model(self, L, deps=[], class_balance=[0.5, 0.5]):
+        label_model = LabelModel(k=2, verbose=False)
+        label_model._set_constants(L)
+        label_model._set_dependencies(deps=deps)
+        label_model._generate_O(L)
+        label_model._build_mask()
+        label_model._get_augmented_label_matrix(L, higher_order=True)
+        label_model.inv_form = len(deps) > 0
```
henryre: Seems like this doesn't do anything anymore? Same below

snorkel/labeling/model/label_model.py
```diff
@@ -50,10 +50,6 @@ def _create_L_ind(self, L):
 
         Note that no column is required for 0 (abstain) labels.
         """
-        # TODO: Update LabelModel to keep L variants as sparse matrices
-        # throughout and remove this line.
-        if issparse(L):
```
henryre: So we're requiring a manual cast now?

henryre: The LF appliers spit out a CSR. Should we change that?

henryre:
High level comment: ideally this should be 3 diffs.

1. Small refactor and test update for baselines
2. Removing unused functionality from LabelModel and any relevant tests
3. Refactoring LabelModel tests

test/labeling/model/test_label_model.py
paroma: I got it from Mathematica after doing the needful operations on the label matrix and substituting variables. I'll change to check 2 decimal places

test/labeling/model/test_label_model.py
paroma: With this small L, it should be dropping given the default LR

test/labeling/model/test_label_model.py
paroma: Will check to 2 decimal places

snorkel/labeling/model/label_model.py
paroma: this was a repeat of line 32 in the same file (`_check_L`) which already converts it to a dense matrix if it's sparse

paroma:
Closing PR as suggested. Changes shifted to stacked PRs (#1209, #1210) and #1202

PR #1183, opened 2019-06-28
Move metrics and shared analysis tools into snorkel/analysis

bhancock8:
- Mostly moving around files/directories. 
- Metrics and error analysis tools (e.g., confusion matrix) are now under `snorkel/analysis/`, which all other directories can pull from
- Label-model-specific analysis tools (e.g., lf_summary()) remain in `snorkel/labeling/analysis.py`
- Some typing added along the way, but majority of typing will be added in a future PR

Test plan:
After refactor, all tests continue to pass.


henryre:
Ping when Travis is fixed

bhancock8:
Ready for review

snorkel/analysis/metrics.py
```diff
@@ -208,7 +178,6 @@ def _preprocess(gold, pred, prob, ignore_in_gold, ignore_in_pred):
     "recall": recall_score,
     "f1": f1_score,
     "fbeta": fbeta_score,
-    "roc-auc": roc_auc_score,
```
henryre: Why removing?

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        A [num_datapoints, num_classes] array of probabilistic labels such that each
+        row sums to 1.
+
+    Returns
+    -------
+    np.ndarray
+        A [num_datapoints, 1] array of predictions (integers in [1, ..., num_classes])
+    """
+    assert prob.ndim == 2
+    assert (prob <= 1).all()
```
henryre: This is expensive, and if we do this, we should also check 0 <= p <= 1, p.sum(1) = 1. My vote is to remove and rely on contract in the docstring.

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+
+    pred = np.argmax(prob, axis=1) + 1
+
+    n, k = prob.shape
+    assert (pred >= 1).all()
```
henryre: Also remove these. Should be a unit test instead.

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+    np.ndarray
+        A [num_datapoints, num_classes] array of probabilistic labels with probability
+        of 1.0 in the column corresponding to the prediction
+    """
+    pred = pred.reshape(-1, 1)
+    n = pred.shape[0]
+
+    prob = np.zeros((n, num_classes))
+
+    for idx, class_idx in enumerate(pred):
```
henryre: Should be vectorizable no?

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+    elif isinstance(array_like, list):
+        array_like = np.array(array_like)
+    elif isinstance(array_like, sparse.spmatrix):
+        array_like = array_like.toarray()
+    elif isinstance(array_like, torch.Tensor):
+        array_like = array_like.numpy()
+    elif not isinstance(array_like, np.ndarray):
+        array_like = np.array(array_like)
+    else:
+        msg = f"Input of type {orig_type} could not be converted to 1d " "np.ndarray"
```
henryre: what's with the cat?

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        raise ValueError(msg)
+
+    # Correct shape
+    if (array_like.ndim > 1) and (1 in array_like.shape):
```
henryre: If we just use squeeze, won't this work without the check?

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,130 @@
...
+        array_like = array_like.flatten()
+    if array_like.ndim != 1:
```
henryre: How about checking sum(dim > 1 for dim in shape), then just flattening?

henryre:
Love the -200 differential

snorkel/analysis/metrics.py
bhancock8: Restored in a future commit when we wrap scikit-learn

snorkel/analysis/utils.py
bhancock8: I was bummed that numpy didn't have a torch.scatter like op, but behold a numpy one-liner hack: 
```return np.eye(num_classes)[pred.squeeze() - 1]```

snorkel/analysis/utils.py
bhancock8: Addressed in a subsequent PR in the pipeline.

bhancock8:
Comments addressed. Merging in.

PR #1185, opened 2019-06-29
Add Mapper memoization

henryre:
Adds memoization to `BaseMapper` classes (preprocessor, TF, etc.).

* BaseMapper objects just maintain a simple dictionary cache which is populated/read from if memoization is turned on
* Only complexity is in `get_hashable`, which recursively generates a hashable value type for common unhashable data types we encounter (SimpleNamespace, pd.Series, np.ndarray, list, dict). It's a bit tricky, so it's extensively tested in both `test/labeling/apply/test_lf_applier.py` and `test/map/test_core.py`
* Considered using `functools.lru_cache`, but it doesn't handle unhashable types
* `lambda_mapper` is now a class so it has parens (@bhancock8 @vincentschen)

There are times we could manually reset the preprocessor caches (e.g. LFApplier.__call__), but not sure we always want to. Opinions welcome. In a follow-up diff, will introduce nested mappers. These two diffs together mean we avoid re-execution of preprocessors when they're used by multiple LFs.

**Test plan**
* Lots of added unit tests for `get_hashable` and BaseMapper caching
* Speed-up test results below. Ran on 2000k English movie review sentences with 4 LFs that use SpaCy. Speed-up is close to 4x (corresponding to reduction in SpaCy calls).

Screen Shot 2019-06-28 at 7 23 19 PM

### DIFF
--- a/snorkel/map/core.py
+++ b/snorkel/map/core.py
@@ -94,6 +94,19 @@ class BaseMapper:
 
     Implements mode setting, memoization, and deep copy functionality.
 
+    Parameters
+    ----------
+    memoize
+        Memoize mapper outputs?
+
+    Attributes
+    ----------
+    mode
+        Mapper mode, corresponding to input data point format.
+        See `MapperMode`.
+    memoize
+        Memoize mapper outputs?
+
     Raises
     ------
     NotImplementedError
@@ -101,32 +114,13 @@ class BaseMapper:
     """
 
     def __init__(self, memoize: bool) -> None:
-        self.memoize_outputs(memoize)
+        self.mode = MapperMode.NONE
+        self.memoize = memoize
         self.reset_cache()
-
-    def memoize_outputs(self, mode: bool) -> None:
-        """Set memoization mode.
-
-        Parameters
-        ----------
-        mode
-            Memoize mapper outputs?
-        """
-        self._memoize_outputs = mode
 
     def reset_cache(self) -> None:
         """Reset the memoization cache."""
         self._cache: Dict[DataPoint, DataPoint] = {}
-
-    def set_mode(self, mode: MapperMode) -> None:
-        """Change mapper mode, depending on data point format.
-
-        Parameters
-        ----------
-        mode
-            Mode to set mapper to
-        """
-        self.mode = mode
 
     def _generate_mapped_data_point(self, x: DataPoint) -> Optional[DataPoint]:
         raise NotImplementedError
@@ -135,7 +129,10 @@ class BaseMapper:
         """Run mapping function on input data point.
 
         Deep copies the data point first so as not to make
-        accidental in-place changes.
+        accidental in-place changes. If `memoize` is set to
+        `True`, an internal cache is checked for results. If
+        no cached results are found, the computed results are
+        added to the cache.
 
         Parameters
         ----------
@@ -147,7 +144,7 @@ class BaseMapper:
         DataPoint
             Mapped data point of same format but possibly different fields
         """
-        if self._memoize_outputs:
+        if self.memoize:
             # NB: don't do `self._cache.get(...)` first in case cached value is `None`
             x_hashable = get_hashable(x)
             if x_hashable in self._cache:
@@ -157,7 +154,7 @@ class BaseMapper:
         # with a dictionary attribute won't create a copy of the dictionary
         x = pickle.loads(pickle.dumps(x))
         x_mapped = self._generate_mapped_data_point(x)
-        if self._memoize_outputs:
+        if self.memoize:
             self._cache[x_hashable] = x_mapped
         return x_mapped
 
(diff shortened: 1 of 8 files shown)


snorkel/map/core.py
```diff
@@ -28,19 +33,90 @@ def get_parameters(
     return params[0]
 
 
+def is_hashable(obj: Any) -> bool:
+    """Test if object is hashable via duck typing.
+
+    NB: not using `collections.Hashable` as some objects
```
bhancock8: NB?

bhancock8: By usage, I assume this just means "Note:", but the latter seems clearer?

bhancock8: (Whatever we decide, update all instances of NB)

test/map/test_core.py
```diff
@@ -36,29 +41,41 @@ def run(self, text: str, **kwargs: Any) -> Optional[FieldMap]:  # type: ignore
...
+    def _get_x(self, num=8, text="Henry has fun") -> SimpleNamespace:
+        return SimpleNamespace(num=num, text=text)
 
     def _get_x_dict(self) -> SimpleNamespace:
         return SimpleNamespace(num=8, d=dict(my_key=1))
 
     def test_numeric_mapper(self) -> None:
         square.set_mode(MapperMode.NAMESPACE)
         x_mapped = square(self._get_x())
+        # NB: not using `self.assertIsNotNone` due to mypy
```
bhancock8: Could we do `self.assertTrue(x_mapped is not None)` instead?

snorkel/map/core.py
henryre: This is going to be so pretentious. https://en.wikipedia.org/wiki/Nota_bene. It means "Important note". Pretty common in code comments.

test/map/test_core.py
henryre: Hmmm, I'll try it

snorkel/map/core.py
ajratner: Hahahaha I literally read Braden's first comment, and was like- I can't wait for Henry's response it's going to be great ("Well, on the Yale Daily News, we used it for...")

ajratner: However I also have seen it a lot fwiw, although I also default to 'Note:'

snorkel/map/core.py
```diff
@@ -28,19 +33,90 @@ def get_parameters(
...
+    a more specific exception.
+    """
+    try:
+        hash(obj)
+        return True
+    except Exception:
+        return False
+
+
+def get_hashable(obj: Any) -> Hashable:
```
ajratner: FWIW: If we did the approach where we converted every data type to a common IR, we could just make sure that said IR was hashable with `hash` and skip this complexity, which makes me nervous...

snorkel/map/core.py
```diff
@@ -28,19 +33,90 @@ def get_parameters(
...
+
+    def memoize_outputs(self, mode: bool) -> None:
+        """Set memoization mode.
+
+        Parameters
+        ----------
+        mode
+            Memoize mapper outputs?
+        """
+        self._memoize_outputs = mode
```
ajratner: Is this a little overkill?  Can't we just set the `self._memoize_outputs` attrib in init...?  I *really* want us to avoid code complexity that isn't strictly needed... it is going to *kill* us at this stage.

Regardless, if we move to only using getters / setters a la Java, I think we should be consistent throughout, and I don't think we're anywhere near there...

snorkel/map/core.py
```diff
@@ -52,7 +128,10 @@ def set_mode(self, mode: MapperMode) -> None:
         """
         self.mode = mode
```
ajratner: Ditto here re: comment above

snorkel/map/core.py
ajratner: (this is not a minor comment, but not an actionable one in this PR...)

henryre: Not necessarily...the values in certain fields of the IR might not be hashable (e.g. a dictionary). I agree this would help with the top-layer, but we'd still need this functionality. Will keep it on the roadmap.

snorkel/map/core.py
henryre: ☠️

Of course we're not going towards setters and getters, but I do see people preferring this type of mode setting (e.g. `torch.nn.Module.train(...)`) for things that actual users interact with (not just other classes). Open to just making `memoize` a public attrib for now.

snorkel/map/core.py
henryre: Ooooh you shouldn't have highlighted this one...now I want to leave these methods for proper type checking.

henryre: Scratch that, mypy will yell if we type the attribute.

snorkel/map/core.py
henryre: Although `train(...)` sets train mode recursively, so that's a bad example. Ok let's make it a public attrib

test/map/test_core.py
henryre: Answer: nope

henryre:
Travis failure due to non-deterministic slicing test, merging

PR #1186, opened 2019-06-29
Add basic deterministic policies

henryre:
* Adds distribution option for random policy (e.g. sticking in a distribution learned by a TANDA mean-field model)
* Adds a policy that just generates [0, 1, ..., k]
* Adds a policy that just generates [0]

**Test plan**
* Added unit test for distribution
* Other two don't really need testing, but assured basic functionality by using ApplyOne in a test

### DIFF
--- a/snorkel/augmentation/apply/tf_applier.py
+++ b/snorkel/augmentation/apply/tf_applier.py
@@ -2,7 +2,7 @@ from typing import Any, List
 
 from tqdm import tqdm
 
-from snorkel.augmentation.policy import AugmentationPolicy
+from snorkel.augmentation.policy import Policy
 from snorkel.augmentation.tf import (
     BaseTransformationFunction,
     TransformationFunctionMode,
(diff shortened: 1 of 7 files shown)


henryre:
Broken connection. None of the travis_retry's worked. Might up the number of retries. :/ 

I did `tox -e coverage` manually and diff coverage was 100%

henryre:
@ajratner can you take a look as well?

henryre:
@ajratner you can pass in probabilities to the RandomAugmentationPolicy now, which is sufficient to reproduce given a learned distribution. Updated doc string points to TANDA. Did you mean policy learning code?

Also how about just `*AugmentationPolicy` -> `*Policy`?

ajratner:
@henryre no that sounds great (though maybe should name it a different class so it stands out? `Random` is very generic)

`*Policy` is okay although a bit generic, when this is a sequence policy?

PR #1188, opened 2019-06-30
Add generator mode to TF applier

henryre:
Add a method to TFApplier that takes a batch size and yields lists/DataFrames of transformed examples for batches. Implemented as separate `apply` method for stronger typing and for alignment with e.g. https://keras.io/preprocessing/image/.

**Test plan**
Added unit tests

### DIFF
--- a/test/augmentation/apply/test_tf_applier.py
+++ b/test/augmentation/apply/test_tf_applier.py
@@ -40,6 +40,10 @@ DATA_IN_PLACE_EXPECTED = [
 ]
 
 
+def make_df(values: list, index: list, key: str = "num") -> pd.DataFrame:
+    return pd.DataFrame({key: values}, index=index)
+
+
 # NB: reconstruct each time to avoid inplace updates
 def get_data_dict():
     return [dict(my_key=num) for num in DATA]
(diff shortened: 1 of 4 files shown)


test/augmentation/apply/test_tf_applier.py
```diff
@@ -94,6 +99,53 @@ def test_tf_applier_keep_original_modify_in_place(self):
...
+    def test_tf_applier_returns_none_generator(self) -> None:
+        data = self._get_x_namespace()
+        applier = TFApplier([square_returns_none], policy, k=2, keep_original=True)
+        batches_expected = [[1, 1, 1, 2], [3, 81, 81]]
+        gen = applier.apply_generator(data, batch_size=2)
+        for batch, batch_expected in zip(gen, batches_expected):
+            self.assertEqual([x.num for x in batch], batch_expected)
+        self.assertEqual([x.num for x in data], DATA)
+
+    def test_tf_applier_keep_original_modify_in_place_generator(self) -> None:
```
bhancock8: It's a little hard to follow what this test is doing, both since "modify in place _and_ keep original" is a bit of a paradox, and because both the assert test below and the DATA_IN_PLACE_EXPECTED are generated programmatically, so for a user to understand the test they need to hold a few things in their heads at once.

henryre: Ok sg

henryre:
@paroma yeah that's the plan. was holding off for now since parallel data aug isn't really expected by most users (esp. for generator mode). Will explicitly add to roadmap though.

PR #1195, opened 2019-07-01
Update analysis tools (docstrings, types, standard names, wrap sklearn) 

bhancock8:
This PR updates our analysis tools:

- Adds types
- Adds docstrings
- Moves shared analysis tools (e.g., metrics) to `analysis/`
- Standardizes names to (golds, probs, preds) from a mixture of singular/plural versions
- Replaces metrics with a single method metric_score that wraps scikit-learn implementations

**Test plan:**
Existing tests for each metric and confusion matrix continue to pass


snorkel/analysis/utils.py
```diff
@@ -0,0 +1,251 @@
...
+        array_like = array_like.toarray()
+    elif isinstance(array_like, torch.Tensor):
+        array_like = array_like.numpy()
+    elif not isinstance(array_like, np.ndarray):
+        array_like = np.array(array_like)
+    else:
+        msg = f"Input of type {orig_type} could not be converted to an np.ndarray"
+        raise ValueError(msg)
+
+    # Correct shape
```
bhancock8: In a previous PR, @henryre proposed replacing these with a single `squeeze()` command. The issue is that there are places where we use this to convert golds/preds/probs to numpy before applying metrics to them, and we don't want to flatten and cast the probs.

snorkel/analysis/error_analysis.py
```diff
@@ -0,0 +1,207 @@
...
+    if normalize:
+        mat = mat / len(golds)
+
+    if pretty_print:
+        conf.display(normalize=normalize)
+
+    return mat
+
+
+class ConfusionMatrix(object):
```
paroma: I guess we've had this since metal days, but when do we use the iterative confusion matrix?

henryre:
@bhancock8 there's no real way to review this diff. can't pick out the moves + already-revieweds from the new stuff. current plan is to just look at the new sklearn-based metrics.

bhancock8:
It's true--it's a squashed monster. Assume types/docstrings are new where they exist, `metrics.py` is new, and everything else is a move.

snorkel/analysis/error_analysis.py
bhancock8: True, we don't use the iterative building nature anymore, and it probably introduces unnecessary complexity to have it. I'll make a note to merge the class into the method down the road.

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,251 @@
...
+    """Convert a matrix from one label type to another
+
+    Args:
+        Y: A np.ndarray or torch.Tensor of labels (ints) using source convention
+        source: The convention the labels are currently expressed in
+        target: The convention to convert the labels to
+    Returns:
+        Y: an np.ndarray or torch.Tensor of labels (ints) using the target convention
+
+    Conventions:
```
paroma: Note for later: we should probably remove the `plus minus` notation completely (including tutorials, that's on me)

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,251 @@
...
+
+    Returns
+    -------
+    ArrayLike
+        Integer labels in categorical form categorical: (0 = abstain, 1 = pos, 2 = neg)
+    """
+    return convert_labels(Y, "plusminus", "categorical")
+
+
+def categorical_to_plusminus(Y: ArrayLike) -> ArrayLike:
```
paroma: do we need these separate functions given they're a simple return statement? the `convert_labels` function is clear enough

snorkel/analysis/utils.py
bhancock8: I think we definitely shouldn't use it anywhere, but I think there's a good chance that datasets people bring in will have one-zero or plus-minus notation, and it's nice to have an easy way for them to convert into the form we accept.

snorkel/analysis/utils.py
bhancock8: I vote we drop them too. We'll just need to pull out where they're used in the tutorials. I'll make a note for a subsequent PR.

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,251 @@
...
+    Returns
+    -------
+    np.ndarray
+        A [num_datapoints, 1] array of predictions (integers in [1, ..., num_classes])
+    """
+    return np.argmax(probs, axis=1) + 1
+
+
+def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
+    """Convert an array of probabilistic labels into an array of predictions
```
paroma: docstring describes `probs_to_preds`

test/analysis/test_utils.py
```diff
@@ -0,0 +1,21 @@
+import unittest
+
+import numpy as np
+
+from snorkel.analysis.utils import preds_to_probs, probs_to_preds
+
+PROBS = np.array([[0.1, 0.9], [0.7, 0.3]])
+PREDS = np.array([2, 1])
+PREDS_SOFT = np.array([[0, 1], [1, 0]])
```
paroma: nit: PROBS_ROUND?

snorkel/analysis/metrics.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+class Metric(NamedTuple):
+    """Specifies a metric function and the subset of [golds, preds, probs] it expects"""
+
+    func: Callable[..., float]
+    inputs: List[str] = ["golds", "preds"]
+
+
+def metric_score(
```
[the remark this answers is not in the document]
bhancock8: Yep! Mostly for internal use. Like Keras (https://keras.io/metrics/), it gives us a standard interface to a bunch of metrics so you can call them out by name only.

snorkel/analysis/utils.py
```diff
@@ -0,0 +1,251 @@
...
+        'plusminus': [0: abstain, 1: positive, -1: negative]
+        'onezero': [0: negative, 1: positive]
+
+    Note that converting to 'onezero' will combine abstain and negative labels.
+    """
+    if Y is None:
+        return Y
+    if isinstance(Y, np.ndarray):
+        Y = Y.copy()
+        assert Y.dtype == np.int64
```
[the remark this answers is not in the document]
bhancock8: At one point the pytorch cross entropy loss functions required longs, I believe. It seems to work fine now without though, so dropping the type assertions/ValueErrors.

PR #1200, opened 2019-07-02
Refactor task_schedulers as batch_schedulers

bhancock8:
- Refactors `task_scheduler` as `batch_schedulers` (since they return a schedule of batches, not tasks...)
- Pull out scheduler by name from `batch_scheduler` module so that `trainer.py` does not need to import every supported Scheduler type.

This one of a sequence of PRs that will be made to clean up the trainer/schedulers. 
Docstrings and typing will be added in a subsequent PR.

**Test plan:**
- Tests pass post-refactor
- Scheduler signatures are about to change and will receive unit tests then


snorkel/end_model/batch_schedulers/__init__.py
```diff
@@ -0,0 +1,4 @@
+from .sequential_scheduler import SequentialScheduler
+from .shuffled_scheduler import ShuffledScheduler
+
+batch_schedulers = {"sequential": SequentialScheduler, "shuffled": ShuffledScheduler}
```
henryre: Can we define this in trainer.py instead?

bhancock8: The point of moving it here was to get it out of trainer.py. If it's in trainer.py, then trainer.py needs to import all scheduler types and whenever a new scheduler is added, trainer.py needs to change.

henryre: k fine

PR #1207, opened 2019-07-03
Add PySpark unit testing

henryre:
On the way to adding integration/complex testing, I found a good way to add PySpark tests. We don't want to run them with everything else all the time because we need Java and the PySpark install is huge. We do however want to run them in Travis.

You can also see that I'm marking things with `pytest.mark.complex`. In a future diff, we'll add integration/complex system testing support by just running `pytest.mark.complex` on master.

It's important to note that single node Spark tests won't catch everything. But it's the best we can do with current CI setup.

* Add PySpark unit tests, decorated with `pytest.mark.spark`
* Separate Spark tests from other tests in pytest invocation in tox.ini
* Update Travis to install JDK8 so that we can run Spark tests
* Add documentation on PySpark testing

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -85,6 +85,17 @@ Note that we use PEP 484 type hints, so parameter types should be removed from t
 }
 ```
 
+### PySpark tests
+PySpark tests are invoked separately from the rest since they require
+installing Java and the large PySpark package.
+They are executed on Travis, but not by default for a local `tox` command.
+If you're making changes to Spark-based operators, make sure you have
+Java 8 installed locally and then run `tox -e spark`.
+If you add a test that imports PySpark mark it with the 
+`@pytest.mark.spark` decorator.
+Add the `@pytest.mark.complex` decorator as well if it runs a Spark
+action (e.g. `.collect()`).
+
 
 ## PRs
 
--- a/requirements.txt
+++ b/requirements.txt
@@ -28,7 +28,7 @@ networkx>=2.2
 
 # Notebook usage (may be ported to tutorials repo)
 jupyter>=1.0.0
-matplotlib>=3.0.1
+matplotlib==3.0.1
 
 # NLP applications
 spacy>=2.1.3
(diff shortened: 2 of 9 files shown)


CONTRIBUTING.md
```diff
@@ -85,6 +85,17 @@ Note that we use PEP 484 type hints, so parameter types should be removed from t
 }
 ```
 
+### PySpark tests
+PySpark tests are invoked separately from the rest since they require
+installing Java and the large PySpark package.
+They are executed on Travis, but not by default for a local `tox` command.
```
paroma: repeating discussion we had earlier, TBD a way to not install packages every time a PR is created/updates are pushed to Travis

requirements.txt
```diff
@@ -40,10 +40,9 @@ tensorboardX>=1.6
 # Scale-up/out
 dask[complete]>=1.2.1
 
-# Note: you may want to remove PySpark from the 
-# required installs. Installing a new version may
-# overwrite your existing system install.
-pyspark>=2.4.3
+# Note: we don't include PySpark in the required installs.
+# Installing a new version may overwrite your existing system install.
+# pyspark>=2.4.3
```
paroma: we don't want  == some version that works for the tests?

CONTRIBUTING.md
henryre: You mean the `tox` setup steps on Travis are taking too long?

requirements.txt
henryre: yeah I could make a requirements-spark.txt file to pin it just for tests

CONTRIBUTING.md
henryre: yeah...seems like caching pip (which we already do) is the best we can do. i can keep digging.

paroma:
In general, the `lf_applier` and `map` tests are repeated per data type within the applier file and then again with Spark, is there a way to get around this somehow?

henryre:
@paroma good point. kind of trading off readability for maintainability (maybe) here. as in, programmatically generated test cases are more difficult to reason with manually. any ideas coming to mind for a clean up?

CONTRIBUTING.md
henryre: Opened a task in Asana, pushing on this now.

PR #1208, opened 2019-07-03
[EASY] Refactor Multitask{DataLoader, Dataset} to Snorkel{...}

bhancock8:
As simple as it looks: just change MultitaskDataLoader to SnorkelDataLoader and MultitaskDataset to SnorkelDataset. 

With recent refactorings, all end model tooling can be used by both single-task and multi-task end models except the models themselves, so the dataset/dataloader classes are getting more general names.

### DIFF
--- a/snorkel/end_model/data.py
+++ b/snorkel/end_model/data.py
@@ -8,7 +8,7 @@ from torch.utils.data import DataLoader, Dataset
 from .utils import list_to_tensor
 
 
-class MultitaskDataset(Dataset):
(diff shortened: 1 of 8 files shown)


bhancock8:
Following up on Slack discussion, I'm going to close this for now as we consider moving instead to the `EndModel` becoming `Predictor` and then these can be called something like `PredictorDataset` and `PredictorDataLoader`---datasets that were intended for consumption by a `Predictor`

PR #1209, opened 2019-07-03
Ind. LabelModel Part 1: Conditionally Independent Model

paroma:
Part 1 of stacked PR stemming form #1182 

* Remove inverse form of solving for `mu` associated with dependencies

* Remove `deps`

__Test Plan__

Small refactor to `label_model_tests` to match conditionally independent model


snorkel/labeling/model/label_model.py
```diff
@@ -269,39 +252,13 @@ def predict_proba(self, L):
 
         L_aug = self._get_augmented_label_matrix(L)
```
henryre: Do we need to update to reflect changed default?

snorkel/labeling/model/label_model.py
```diff
@@ -188,13 +174,11 @@ def _init_params(self):
...
 
-        # Handle single or per-LF values
+        # Handle single values
         if isinstance(train_config["prec_init"], (int, float)):
-            prec_init = train_config["prec_init"] * torch.ones(self.m)
-        else:
-            prec_init = torch.from_numpy(train_config["prec_init"])
-            if prec_init.shape[0] != self.m:
-                raise ValueError(f"prec_init must have shape {self.m}.")
+            self.prec_init = train_config["prec_init"] * torch.ones(self.m)
```
henryre: Why is this a public attribute now? Name is too unintuitive for that.

snorkel/labeling/model/label_model.py
paroma: just reverted default, I was setting it in the test explicitly

PR #1212, opened 2019-07-03
[EASY] Consistent file naming conventions

henryre:
Simplifies naming scheme to make things more clear and avoid repeating the full tree in the file names. Also moves aliases to import statements to simplicity.

**Test plan**
All unit tests still passing.


henryre:
@bhancock8 ready for another look! Also moved Dask out of module __init__.py imports since it's not a core dependency.

PR #1213, opened 2019-07-03
Reuse tox dependency installs in Travis

henryre:
See title. A little (well documented) tox fanciness to avoid slow reinstalls. Removed develop mode where it wasn't necessary to avoid reinstalls. See

* https://tox.readthedocs.io/en/latest/config.html
* https://blog.ionelmc.ro/2015/04/14/tox-tricks-and-patterns/#environment-reuse

Note: for some reason, we reinstall when calling `tox` after `tox --notest`. We could get a couple minutes back by removing `tox --notest` but that would make it extremely hard to debug install issues. Leaving for now.

**Test plan:**
* Travis speed-up 18m -> 11m. Manually validated outputs.
* Ran the following series of typical developer commands and validated correct outputs (e.g. things fixed by `-e fix` and correct coverage from `-e coverage`

```bash
rm -rf .tox
rm -rf .env
tox -e dev
source .env/bin/activate
tox
# Messed some formatting up
tox
tox -e fix
tox
tox -e check
tox -e spark
tox -e coverage
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -11,7 +11,7 @@ cache: pip
 dist: xenial
 
 env:
-  TOXENV=py,spark,check,coverage JAVA_HOME=/usr/lib/jvm/java-8-openjdk-amd64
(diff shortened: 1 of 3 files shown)


henryre:
Measuring now!

PR #1218, opened 2019-07-03
Update types in classifier/

bhancock8:
Update type comments to reflect recent changes.

**Test plan:**
tox -e check passes and tox -e type only complains on files outside of classifier

### DIFF
--- a/snorkel/classification/models/advanced/advanced_classifier.py
+++ b/snorkel/classification/models/advanced/advanced_classifier.py
@@ -107,8 +107,8 @@ class AdvancedClassifier(nn.Module):
         self._move_to_device()
 
     def forward(  # type: ignore
-        self, X_dict: Mapping[Union[str, int], Any], task_names: Iterable[str]
-    ) -> Dict[str, Mapping[Union[str, int], Any]]:
+        self, X_dict: Mapping[str, Any], task_names: Iterable[str]
+    ) -> Any:
         """Forward pass through the network
 
         :param X_dict: The input data
@@ -119,7 +119,7 @@ class AdvancedClassifier(nn.Module):
         X_dict = move_to_device(X_dict, self.config["device"])
 
         outputs: Dict[str, Mapping[Union[str, int], Any]] = {}
-        outputs["_input_"] = X_dict
+        outputs["_input_"] = X_dict  # type: ignore
 
         # Call forward for each task, using cached result if available
         # Each task flow consists of one or more operations that are executed in order
@@ -149,11 +149,14 @@ class AdvancedClassifier(nn.Module):
                         output = [output]
                     outputs[operation.name] = output
 
+        # Note: We return all outputs to enable advanced workflows such as multi-task
+        # learning (where we want to calculate loss from multiple head modules on some
+        # forward passes).
         return outputs
 
     def calculate_loss(
         self,
-        X_dict: Mapping[Union[str, int], torch.Tensor],
+        X_dict: Mapping[str, Any],
         Y_dict: Dict[str, torch.Tensor],
         task_to_label_dict: Dict[str, str],
     ):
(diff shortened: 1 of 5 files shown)


snorkel/classification/models/advanced/advanced_classifier.py
```diff
@@ -119,7 +119,7 @@ def forward(  # type: ignore
         X_dict = move_to_device(X_dict, self.config["device"])
 
         outputs: Dict[str, Mapping[Union[str, int], Any]] = {}
-        outputs["_input_"] = X_dict
+        outputs["_input_"] = X_dict  # type: ignore
```
henryre: What's the issue here?

snorkel/classification/models/advanced/advanced_classifier.py
```diff
@@ -196,9 +199,7 @@ def calculate_loss(
         return loss_dict, count_dict
 
     @torch.no_grad()
-    def _calculate_probs(
-        self, X_dict: Mapping[Union[str, int], torch.Tensor], task_names: Iterable[str]
-    ):
+    def _calculate_probs(self, X_dict: Mapping[str, Any], task_names: Iterable[str]):
```
henryre: Not all Tensors any more?

snorkel/classification/scorer.py
```diff
@@ -32,12 +34,14 @@ def __init__(
...
-    def score(self, golds, preds, probs):
+    def score(
+        self, golds: ArrayLike, preds: ArrayLike, probs: ArrayLike
+    ) -> Dict[str, float]:
         metric_dict = dict()
 
         for metric_name, metric in self.metrics.items():
             # Handle no examples
-            if len(golds) == 0:
+            if len(golds) == 0:  # type: ignore
```
henryre: Which ArrayLike has no `len`?

snorkel/classification/training/schedulers/scheduler.py
```diff
@@ -14,7 +16,9 @@ def __init__(self):
     @abstractmethod
     def get_batches(
         self, dataloaders: Sequence[ClassifierDataLoader]
-    ) -> Tuple[Tuple[Dict[str, Any], Dict[str, Any]], ClassifierDataLoader]:
+    ) -> Iterator[
+        Tuple[Tuple[Dict[str, Any], Dict[str, Tensor]], ClassifierDataLoader]
```
henryre: Used below as well, let's alias this (e.g. `BatchIterator = Iterator[...]`)

snorkel/classification/models/advanced/advanced_classifier.py
bhancock8: The type of outputs is: Dict[str, Mapping[Union[str, int], Any]]
The type of X_dict is: Mapping[str, Any]
 
outputs["__input__": X_dict] is unhappy that X_dict cannot possibly be Mapping[int, Any].

snorkel/classification/models/advanced/advanced_classifier.py
bhancock8: Y will always be tensors. X doesn't have to be. e.g., you can pass string-based uids or other tags through as well that a module takes advantage of.

snorkel/classification/scorer.py
bhancock8: It doesn't know what to do with scipy.sparse.

bhancock8:
Comments addressed

snorkel/classification/models/advanced/advanced_classifier.py
henryre: Try Dict[str, Union[Mapping[int, Any], Mapping[str, Any]]]

henryre:
But see one above comment

bhancock8:
Yeah, tried your typing suggestion @henryre, still no dice I think the issue is that we actually can't guarantee type safety here. The output of each operation in the task flow is either a dict of fields or a sequence of fields (including the trivial case of a singleton list if they just output a single tensor), and the user specifies for each operation what its inputs should be by pointing to a previously run module and either an integer index or a string key. It's possible for the user to specify an input like `(my_module_that_outputs_dicts, 0)', which would in fact fail when the task flow tries to access the 0th element of a dict. Leaving as is for now, can always revisit.

PR #1220, opened 2019-07-05
Add more typing to classification/

bhancock8:
- Mostly adds more typing to classification/ pieces.
- Also slightly simplifies a unit test and fixes updates a path in a unit test to reflect the new directory structure.

**Test plan:**
`tox` and `tox -e type` pass


snorkel/classification/data.py
```diff
@@ -1,12 +1,18 @@
...
 
 import numpy as np
 import torch
+from torch import Tensor
 from torch.utils.data import DataLoader, Dataset
 
 from .utils import list_to_tensor
 
+XDict = Dict[str, Any]
+YDict = Dict[str, torch.Tensor]
```
ajratner: Nit: just `Tensor`?

PR #1221, opened 2019-07-05
Simplify/merge save and load methods

bhancock8:
Removes redundant save/load methods, keeping save() and load() and dropping the *_state_dict() versions of these. Note, however, that save()/load() do what they do by working with the state dicts.

**Test plan**
- Current tests pass.
- New tests for save/load are in the follow-up PR #1222 (they require some of the pieces that that PR introduces).

### DIFF
--- a/snorkel/classification/models/advanced/advanced_classifier.py
+++ b/snorkel/classification/models/advanced/advanced_classifier.py
@@ -332,69 +332,28 @@ class AdvancedClassifier(nn.Module):
 
         return metric_score_dict
 
-    def save(self, model_path: str) -> None:
-        """Save the current model
-        :param model_path: Saved model path.
-        :type model_path: str
-        """
-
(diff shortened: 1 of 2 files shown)


snorkel/classification/models/advanced/advanced_classifier.py
```diff
@@ -332,69 +332,28 @@ def score(self, dataloaders: List[ClassifierDataLoader]) -> Dict[str, float]:
...
         return metric_score_dict
 
-    def save(self, model_path: str) -> None:
-        """Save the current model
-        :param model_path: Saved model path.
-        :type model_path: str
-        """
-
-        # Check existence of model saving directory and create if does not exist.
+    def save(self, model_path: str):
```
ajratner: Why are we dropping the docstring and output typing here?

bhancock8:
Subsumed by #1234

PR #1222, opened 2019-07-05
Merge SimpleClassifier and AdvancedClassifier

bhancock8:
After some thought, it's just too confusing having multiple Classifier classes offered by our repo, especially when we'd like to rely more on third party integrations. Instead of creating a separate SimpleClassifier class, there is now an option to instantiate an AdvancedClassifier with a one-liner, such as:
```
model = AdvancedClassifier.from_modules([nn.Linear(2,4), nn.Linear(4,2)])
```
This should only be used in cases where certain criteria are met (it's a single task, the data has a single field, etc.) but these conditions are clearly laid out in the docstring for that method. In a follow-up PR once this is merged, I'll fix the directory structure (no need for an `advanced/` directory anymore). 

For our Classifier strategy moving forward, I propose:
- most of our tutorials use other 3rd party tools with simple interfaces (such as Ludwig) and our general recommendation will be to use those.
- The only Classifier in our repo will be the current `AdvancedClassifier`, which I propose we rename as `TaskFlowClassifier`, and we can use that for the slicing and MTL tutorials (situations where the user needs fine-grained control over exactly what fields from what tasks flow through the network in what order).
- Then ideally down the road we're able to find ways to support even these workflows with other tools and the hold `classifier` can just contain wrappers and integration tools for working with other libraries.


snorkel/classification/models/__init__.py
```diff
@@ -1,2 +1 @@
 from .advanced import AdvancedClassifier, Operation, Task  # noqa: F401
```
ajratner: `AdvancedClassifier` is a bad name here- it doesn't mean anything, and even more so when there's no `SimpleClassifier`.  So I think this needs to be changed in this PR.  As for alternatives- `TaskFlowClassifier` also sounds a bit wonky to me.  I don't think "task flow" will mean anything to anyone other than a few people in the lab...

More broadly than just naming: I am still a little confused as to why we don't just have `SingleTaskClassifier` (perhaps shortened or aliased to `Classifier`) and `MultiTaskClassifier`?  This is a straightforward distinction that anyone will understand and seems cleaner

snorkel/classification/models/advanced/advanced_classifier.py
```diff
@@ -357,3 +360,77 @@ def load(self, model_path: str):
...
+        """Instantiates a simple AdvancedClassifier from `Module`s instead of `Task`s
+        This is intended as a convenience method for quickly creating an
+        AdvancedClassifier in certain situations where many defaults can be inferred.
+        Only use this method of instantiation if all of the following properties apply:
+        - Your problem has only one task
+            - That task name is the only key in dataloader.task_to_label_dict
+        - Your data has only one field
+            - The `data_name` you pass here and that field name in the X_dict match
+        - Your architecture is a simple sequential path using each Module exactly once
+        - You don't require passing any custom metric functions to the Scorer
```
ajratner: Even after reading this doc string I have no idea what's going on here, when this method is applicable, how to use it... way to complex for a simple single task classifier.

Of course, I haven't reviewed this new MTL architecture- but that's *exactly* why I think my take is valuable here. Because if it takes intense insider knowledge of some MTL model "task flow" thing to understand how to use a basic single task classifier class in our framework, then we're done- no one's going to use it.

bhancock8:
Yep, conversation required. There are just a lot of interconnected decisions here; will be much easier to discuss in person.

bhancock8:
Subsumed by #1234

PR #1227, opened 2019-07-06
Simplify and align LF and Mapper interfaces

henryre:
Removes some currently unused fields from LF, adds name to Mapper/TF/Preprocessor to align with LF definition.

**Test plan**
Unit tests

### DIFF
--- a/snorkel/labeling/preprocess/nlp.py
+++ b/snorkel/labeling/preprocess/nlp.py
@@ -39,7 +39,8 @@ class SpacyPreprocessor(Preprocessor):
         language: str = EN_CORE_WEB_SM,
         disable: List[str] = DEFAULT_DISABLE,
     ) -> None:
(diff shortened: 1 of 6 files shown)


snorkel/labeling/preprocess/nlp.py
```diff
@@ -39,7 +39,8 @@ def __init__(
         language: str = EN_CORE_WEB_SM,
         disable: List[str] = DEFAULT_DISABLE,
     ) -> None:
-        super().__init__(dict(text=text_field), dict(doc=doc_field))
+        name = type(self).__name__
```
ajratner: Why not just set this as the default for `name`?

henryre: You mean allow people to pass in their own name with this as the default if None is passed in?

PR #1228, opened 2019-07-07
Add NLP LF helper

henryre:
Add `NLPLabelingFunction` and `nlp_labeling_function` to make it easier to use `SpacyPreprocessor` and write LFs over parsed SpaCy docs. Constructs an LF with a single preprocessor, a `SpacyPreprocessor` instance which might have other nested preprocessors. All instances of NLPLabelingFunction share the same `SpacyPreprocessor` instance to take advantage of memoization. This makes the assumption that in any basic application, only one configuration of `SpacyPreprocessor` will exist. More advanced Snorkel applications can just manually define multiple `SpacyPreprocessor` objects if needed.

* Adds `NLPLabelingFunction` and `nlp_labeling_function`
* Adds relevant unit tests
* Adds updated `Mapper` kwargs to `SpacyPreprocessor`

```python
@nlp_labeling_function()
def has_person_mention(x: DataPoint) -> int:
    person_ents = [ent for ent in x.doc.ents if ent.label_ == "PERSON"]
    return 1 if len(person_ents) > 0 else 0

x = SimpleNamespace(text="The movie was good.")
has_person_mention(x)  # 0
```

**Test plan**
Added unit tests


henryre:
Offline: adding option to spit doc fields into top fields

ajratner:
Talked offline- tag when ready for review!

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,202 @@
+from typing import Any, Callable, List, Mapping, NamedTuple, Optional
+
+from snorkel.labeling.preprocess import BasePreprocessor
+from snorkel.labeling.preprocess.nlp import EN_CORE_WEB_SM, SpacyPreprocessor
+
+from .core import LabelingFunction
+
+
+class SpacyPreprocessorParameters(NamedTuple):
+    """Parameters need to construct a SpacyPreprocessor."""
```
vincentschen: should there be an explanation of the params here? or maybe a reference to the class where we use them?

snorkel/augmentation/tf.py
```diff
@@ -10,13 +9,7 @@
...
 
 
-class BaseTransformationFunction(BaseMapper):
-    """Base class for `TransformationFunction` and `transformation_function`.
-
-    See `snorkel.map.core.BaseMapper` for details.
-    """
-
-    pass
+BaseTransformationFunction = BaseMapper
```
vincentschen: The inheritance above is a pattern we've used elsewhere to "alias" classes—we should be consistent?

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,202 @@
...
+
+    def __init__(
+        self,
+        name: Optional[str] = None,
+        resources: Optional[Mapping[str, Any]] = None,
+        preprocessors: Optional[List[BasePreprocessor]] = None,
+        fault_tolerant: bool = False,
+        text_field: str = "text",
+        doc_field: str = "doc",
+        language: str = EN_CORE_WEB_SM,
```
vincentschen: can we add a small comment about what this refers to?

snorkel/labeling/preprocess/nlp.py
```diff
@@ -1,12 +1,11 @@
...
 
 import spacy
 
 from snorkel.types import FieldMap
 
-from .core import Preprocessor
+from .core import BasePreprocessor, Preprocessor
 
-DEFAULT_DISABLE = ["parser", "ner"]
 EN_CORE_WEB_SM = "en_core_web_sm"
```
vincentschen: use a more instructive variable name (e.g. `SPACY_CORPUS`?)

snorkel/augmentation/tf.py
ajratner: It's also unclear where this is used, given there's a base class for TFs, `TransformationFunction`, right below... why do we need both?  (A) there should at least be a comment here explaining, but more importantly, this seems *very* heavyweight if we really do need both...

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,202 @@
...
+    Parameters
+    ----------
+    name
+        Name of the LF
+    f
+        Function that implements the core LF logic
+    resources
+        Labeling resources passed in to `f` via `kwargs`
+    preprocessors
+        Preprocessors to run before SpacyPreprocessor is executed
```
ajratner: What about if you want to run a preprocessor after?  This seems equally likely

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,202 @@
...
+    disable: Optional[List[str]]
+    preprocessors: List[BasePreprocessor]
+    memoize: bool
+
+
+class SpacyPreprocessorConfig(NamedTuple):
+    """Tuple of SpacyPreprocessor and the parameters used to construct it."""
+
+    nlp: SpacyPreprocessor
+    parameters: SpacyPreprocessorParameters
```
ajratner: Seems like defining this isn't really necessary?  Seems a little overkill?

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,202 @@
...
+    ```
+    @nlp_labeling_function()
+    def has_person_mention(x: DataPoint) -> int:
+        person_ents = [ent for ent in x.doc.ents if ent.label_ == "PERSON"]
+        return 1 if len(person_ents) > 0 else 0
+    print(f)  # "NLPLabelingFunction has_person_mention"
+
+    x = SimpleNamespace(text="The movie was good.")
+    has_person_mention(x)  # 0
+    ```
```
ajratner: Should (a) include this example with the LF above, (b) give the standard list of Spacy attributes, and (c) be more clear in the docstring that everything is accessible via the `doc` attribute by default!

snorkel/augmentation/tf.py
henryre: This is only used for type annotations, and mypy balks if you subclass like we do above. Having this alias IMO makes the TF files more legible.

snorkel/labeling/lf/nlp.py
henryre: One could argue this isn't basic usage any more, so might be better served to roll your own using the SpaCy preprocessor. Thoughts?

snorkel/labeling/lf/nlp.py
henryre: The tuple? It's a cleaner way to enforce mutual inclusivity and other properties of the parameters than a bunch of logical checks.

snorkel/augmentation/tf.py
ajratner: Okay well at least comment explaining this!

snorkel/labeling/lf/nlp.py
henryre: included in `NLPLabelingFunction`, which is referenced by this decorator. is that enough?

snorkel/labeling/preprocess/nlp.py
henryre: going to leave this as is. don't want to bias people towards thinking this is a default spacy corpus (since it isn't, might want bigger or non-english model)

snorkel/labeling/lf/nlp.py
henryre: What do you mean by "give the standard list of Spacy attributes"?

snorkel/labeling/lf/nlp.py
vincentschen: sg!

snorkel/labeling/preprocess/nlp.py
vincentschen: why define a global variable in the first place, then?

henryre: It's reused in NLPLabelingFunction

snorkel/labeling/lf/nlp.py
ajratner: Sorry, coming probably too late, but I meant copy-paste some of the important attributes from spacy people might want (e.g. NER tags, tokens, etc etc)

snorkel/labeling/lf/nlp.py
```diff
@@ -0,0 +1,210 @@
...
+    with all other ``NLPLabelingFunction`` instances. This makes
+    it easy to define LFs that have a text input field and have
+    logic written over SpaCy ``Doc`` objects. Examples passed
+    into an ``NLPLabelingFunction`` will have a new field which
+    can be accessed which contains a SpaCy ``Doc``. By default,
+    this field is called ``doc``. For details of SpaCy ``Doc``
+    objects and a full attribute listing, see https://spacy.io/api/doc.
+
+    Simple ``NLPLabelingFunction``s can be defined via a
+    decorator. See ``nlp_labeling_function``.
```
ajratner: Could we have (a) a quick example here (just copy-paste the one from the PR description?) and (b) a partial listing of the top 5/10 spacy attributes (people will be lazy re: following the link)

snorkel/labeling/lf/nlp.py
ajratner: This is not major but looking down below, I still think it's confusing that the "config" object contains the actual preprocessor

snorkel/labeling/lf/nlp.py
ajratner: Or just copy from decorator version below?

PR #1230, opened 2019-07-08
Merge snorkel/model into snorkel/labeling/model

henryre:
* Removes snorkel/model
* Merges currently-used classifier functionality into LabelModel, strips away everything else
* Removes log writers, checkpointers, and a lot of logging for LabelModel

**Test plan**
All unit tests run, manually verified logging output for correctness
+9% repo coverage!


snorkel/labeling/model/utils.py
```diff
@@ -96,37 +74,6 @@ def recurse(x, y, misses="report", verbose=1):
...
-    """
-    for k, v in x.items():
-        if test_func(v):
-            x[k] = transform(v)
-        if isinstance(v, dict):
-            recursive_transform(v, test_func, transform)
-    return x
-
-
 def set_seed(seed: int):
```
bhancock8: Use set_seed() from `snorkel/analysis/utils.py`and delete this.

snorkel/labeling/model/utils.py
```diff
@@ -3,28 +3,6 @@
...
-        assert len(X) == len(Y)
-
-    def __getitem__(self, index):
-        return tuple([self.X[index], self.Y[index]])
-
-    def __len__(self):
-        return len(self.X)
 
 
 def recursive_merge_dicts(x, y, misses="report", verbose=None):
```
bhancock8: This also exists in `snorkel.classification.utils`, but the signature of LabelModel is simple enough that I think we should just move the defaults into the model definition and drop this extra copy.

snorkel/labeling/model/logger.py
```diff
@@ -0,0 +1,49 @@
+from collections import defaultdict
+
+
+class Logger:
```
bhancock8: This is fine for this PR, since it's just porting. But we should replace this with the LogManager (identifies when it's time to log) and LogWriter (writes logs to file or tensorboard), since they do the same thing.

snorkel/labeling/model/label_model.py
```diff
@@ -1,33 +1,42 @@
...
 
-    # This class variable is explained in the Classifier class
-    implements_l2 = True
-
     def __init__(self, k=2, **kwargs):
-        config = recursive_merge_dicts(lm_default_config, kwargs)
-        super().__init__(k, config)
+        super().__init__()
+        self.config = recursive_merge_dicts(lm_default_config, kwargs)
+        self.k = k
```
bhancock8: How often does k occur in an equation? Would renaming it to "cardinality" cause any lines to blow up in length?

snorkel/labeling/model/baselines.py
```diff
@@ -1,8 +1,31 @@
 import numpy as np
+import torch
+from scipy.sparse import issparse
 
 from .label_model import LabelModel
 
 
+def to_numpy(Z):
```
bhancock8: Smells that we have to_numpy() here and arraylike_to_numpy() in `analysis/utils.py`. Can they be combined?

snorkel/labeling/model/utils.py
henryre: More involved that one might imagine. Leaving for now.

snorkel/labeling/model/logger.py
henryre: Noted. cc @ajratner @paroma for new label model effort

henryre:
@bhancock8 changes made! @paroma would be great to get comments from you

PR #1233, opened 2019-07-09
PySpark monkey patcher

henryre:
`SparkMapper` worked fine, but you'd have to define a separate preprocessor to create a Spark version. For example, I'd have to do some weird MRO and/or mixin madness to get a Spark version of `SpacyPreprocessor`. But now, I can just do

```python
spacy_preprocessor = make_spark_preprocessor(SpacyPreprocessor())
```

**Test plan**
Shifted all Spark mapper unit tests to new version

### DIFF
--- a/snorkel/map/spark.py
+++ b/snorkel/map/spark.py
@@ -5,7 +5,7 @@ from snorkel.types import FieldMap
 from .core import Mapper
 
 
-def _update_fields(self, x: Row, mapped_fields: FieldMap) -> Row:
+def _update_fields(x: Row, mapped_fields: FieldMap) -> Row:
     # `pyspark.sql.Row` objects are not mutable, so need to
     # reconstruct
     all_fields = x.asDict()
--- a/snorkel/map/spark.py
+++ b/snorkel/map/spark.py
@@ -5,42 +5,21 @@ from snorkel.types import FieldMap
 from .core import Mapper
 
 
-class SparkMapper(Mapper):
-    """Base class for any `Mapper` that runs on PySpark `Row` objects.
+def _update_fields(self, x: Row, mapped_fields: FieldMap) -> Row:
+    # `pyspark.sql.Row` objects are not mutable, so need to
+    # reconstruct
+    all_fields = x.asDict()
+    all_fields.update(mapped_fields)
+    return Row(**all_fields)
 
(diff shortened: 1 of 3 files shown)


snorkel/labeling/preprocess/core.py
```diff
@@ -11,13 +10,7 @@
...
 
 
-class BasePreprocessor(BaseMapper):
-    """Base class for `Preprocessor` and `preprocessor`.
-
-    See `snorkel.map.core.Mapper` for details.
-    """
-
-    pass
+BasePreprocessor = BaseMapper
```
bhancock8: Why did we drop the docstring for just this class and not the other tree trivial wrappers in this file? Also, as a browser of the code, if I see `BasePreprocessor`, `Preprocessor` and `LambdaPreprocessor`, all three of those sound like reasonable classes for me to use for a simple use case. And `BasePreprocessor` is currently only used as a type annotation. Can we clear it up somehow?

snorkel/map/spark.py
```diff
@@ -5,42 +5,21 @@
...
-        A map from attribute names of the incoming data points
-        to the input argument names of the `run` method. If None,
-        the parameter names in the function signature are used.
-    mapped_field_names
-        A map from output keys of the `run` method to attribute
-        names of the output data points. If None, the original
-        output keys are used.
-    memoize
-        Memoize mapper outputs?
+def make_spark_mapper(mapper: Mapper) -> Mapper:
```
bhancock8: If the only thing you need to do to make a Mapper a SparkMapper is to change the logic of _update_fields, why can't we subclass it and just overwrite that method? (I trust there's a good reason--otherwise this PR wouldn't exist--but it's not obvious from review)

snorkel/labeling/preprocess/core.py
henryre: mypy has trouble with this: some of the Mapper base classes use BaseMapper as the type of parameters and since this is a new class, it doesn't read BasePreprocessor as fulfilling that type correctly. As you saw, currently only used a type annotation, so documentation probably isn't needed.

snorkel/map/spark.py
henryre: See description. We want to write a Mapper/Preprocessor/etc. once, then be able to use it with Spark without having to make a new subclass.

PR #1239, opened 2019-07-09
Add unit tests for analysis

paroma:
See title

### DIFF
--- a/test/analysis/test_metrics.py
+++ b/test/analysis/test_metrics.py
@@ -27,6 +27,18 @@ class MetricsTest(unittest.TestCase):
         self.assertRaises(
             ValueError, metric_score, golds, pred3, probs=None, metric="accuracy"
         )
+        self.assertRaises(
+            ValueError, metric_score, golds, pred3, probs=None, metric="bad_metric"
+        )
+        self.assertRaises(
+            ValueError,
+            metric_score,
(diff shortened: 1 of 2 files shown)


test/analysis/test_metrics.py
```diff
@@ -27,6 +27,18 @@ def test_bad_inputs(self):
         self.assertRaises(
             ValueError, metric_score, golds, pred3, probs=None, metric="accuracy"
         )
+        self.assertRaises(
```
henryre: I know these aren't all yours, but can we switch these to

```python
with self.assertRaisesRegex(ValueError, "not currently implemented"):
    metric_score(golds, ...)
```

test/analysis/test_utils.py
```diff
@@ -2,14 +2,39 @@
...
+            ValueError,
+            convert_labels,
+            Y=not_array,
+            source="categorical",
+            target="plusminus",
+        )
+        self.assertIsNone(
+            convert_labels(Y=None, source="categorical", target="plusminus")
+        )
+        np.testing.assert_array_equal(
```
bhancock8: It looks like the last test is a working test, not a bad input.

paroma: changing name to`test_input`

PR #1241, opened 2019-07-09
Support only one string per split name

bhancock8:
Changes:
- Previously, the user could specify a list of split names to map to the "train" split used by the `Trainer` (e.g., `train_set_1` and `train_set_2` could belong to `train`). In practice, this is a very rarely (if ever) used feature. Stripping it out for simplicity.
- Simplifies (and speeds up) trainer unit tests by making a small deterministic dataset rather than a randomly generated one.

**Test plan**:
New unit test confirms that errors are raised when no training dataset is provided to the `Trainer` or when an unrecognized split name is used.

### DIFF
--- a/test/classification/training/test_trainer.py
+++ b/test/classification/training/test_trainer.py
@@ -1,27 +1,33 @@
 import copy
+import tempfile
 import unittest
 
 import torch
 import torch.nn as nn
+import torch.optim as optim
 
 from snorkel.classification.data import DictDataLoader, DictDataset
 from snorkel.classification.scorer import Scorer
 from snorkel.classification.snorkel_classifier import Operation, SnorkelClassifier, Task
 from snorkel.classification.training import Trainer
+from snorkel.classification.training.loggers import LogWriter, TensorBoardWriter
 
 TASK_NAMES = ["task1", "task2"]
-trainer_config = {"n_epochs": 2, "progress_bar": False}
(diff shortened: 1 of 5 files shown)


test/classification/training/test_trainer.py
```diff
@@ -90,5 +51,41 @@ def create_task(task_name, module_suffixes):
...
+        trainer.train_model(model, dataloaders)
+
+    def test_trainer_errors(self):
+        model = SnorkelClassifier([tasks[0]])
+        dataloader = copy.deepcopy(dataloaders[0])
+
+        # No train split
+        trainer = Trainer(**trainer_config)
+        dataloader.dataset.split = "valid"
+        with self.assertRaises(ValueError):
```
paroma: include ValueError message with `assertRaisesRegex`

test/classification/training/test_trainer.py
```diff
@@ -90,5 +51,41 @@ def create_task(task_name, module_suffixes):
...
+            trainer.train_model(model, [dataloader])
+
+        # Unused split
+        trainer = Trainer(**trainer_config, valid_split="val")
+        with self.assertRaises(ValueError):
```
paroma: same as above

bhancock8:
@paroma These `assertRaise` statements are fixed already in #1243 which merges into this. Leaving alone here for simpler cascading merge.

PR #1243, opened 2019-07-09
Add a bunch of unit tests for trainer.py

bhancock8:
Primarily adds unit tests for trainer.py. Also:
- Drop unsupported optimizer/scheduler options (ones that were in the code that was ported over, but didn't have tests to show that they wouldn't actually be usable in their current state)
- Update "batch" -> "batches" and "epoch" -> "epochs" for consistent usage between lr_scheduler and LogManager.

**Test plan**:
New unit tests! These focus primarily on whether or not the appropriate optimizer/scheduler/checkpointer/etc. is found and initialized appropriately, not whether it functions as expected (which would be tested in the unit tests for the individual classes).

### DIFF
--- a/test/classification/training/test_trainer.py
+++ b/test/classification/training/test_trainer.py
@@ -1,30 +1,33 @@
 import copy
+import tempfile
 import unittest
 
 import torch
 import torch.nn as nn
+import torch.optim as optim
 
 from snorkel.classification.data import DictDataLoader, DictDataset
 from snorkel.classification.scorer import Scorer
 from snorkel.classification.snorkel_classifier import Operation, SnorkelClassifier, Task
 from snorkel.classification.training import Trainer
+from snorkel.classification.training.loggers import LogWriter, TensorBoardWriter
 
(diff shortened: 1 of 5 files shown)


test/classification/training/test_trainer.py
```diff
@@ -90,6 +91,94 @@ def test_trainer_errors(self):
...
+
+        trainer = Trainer(**trainer_config, optimizer="adam")
+        trainer.train_model(model, [dataloaders[0]])
+        self.assertIsInstance(trainer.optimizer, optim.Adam)
+
+        trainer = Trainer(**trainer_config, optimizer="adamax")
+        trainer.train_model(model, [dataloaders[0]])
+        self.assertIsInstance(trainer.optimizer, optim.Adamax)
+
+        with self.assertRaises(ValueError):
```
paroma: For the `ValueErrors`, can we include checking the error message as a check for consistency
```python
with self.assertRaisesRegex(ValueError, "not currently implemented"):
```

PR #1252, opened 2019-07-10
Label model: complex tests

ajratner:
Just adding a simple conditionally independent labelers (with non-symmetric class accuracies) test which is tagged as `complex`, along with a new supporting analysis function, `lf_empirical_probs`, that computes the full conditional probability table given a labeled dataset, i.e. `L` and `Y`, rather than just the accuracies.

Note that this is just an initial stub- will extend these tests to be (a) more thorough and (b) include dependencies once the new LabelModel code is in.

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -9,7 +9,6 @@ from scipy.sparse import issparse
 
 from snorkel.analysis.utils import set_seed
 from snorkel.classification.utils import recursive_merge_dicts
-
 from snorkel.labeling.model.graph_utils import get_clique_tree
 from snorkel.labeling.model.lm_defaults import lm_default_config
 from snorkel.labeling.model.logger import Logger
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -10,9 +10,9 @@ from scipy.sparse import issparse
 from snorkel.analysis.utils import set_seed
 from snorkel.classification.utils import recursive_merge_dicts
 
-from .graph_utils import get_clique_tree
-from .lm_defaults import lm_default_config
-from .logger import Logger
+from snorkel.labeling.model.graph_utils import get_clique_tree
+from snorkel.labeling.model.lm_defaults import lm_default_config
+from snorkel.labeling.model.logger import Logger
 
 
 class LabelModel(nn.Module):
--- a/test/labeling/model/test_label_model.py
+++ b/test/labeling/model/test_label_model.py
@@ -296,15 +296,25 @@ class TestLabelModelAdvanced(unittest.TestCase):
         P, Y, L = self._generate_data()
         P_emp = lf_empirical_probs(L, Y, k=self.k)
         np.testing.assert_array_almost_equal(P, P_emp, decimal=2)
-
+    
     def test_label_model(self) -> None:
         """Test the LabelModel's estimate of P and Y."""
         np.random.seed(123)
         P, Y, L = self._generate_data()
+
+        # Train LabelModel
         label_model = LabelModel(cardinality=self.k, verbose=False)
         label_model.train_model(L, lr=0.01, l2=0.0, n_epochs=100)
+
+        # Test estimated LF conditional probabilities
         P_lm = label_model.get_conditional_probs().reshape((self.m, self.k + 1, -1))
         np.testing.assert_array_almost_equal(P, P_lm, decimal=2)
+
+        # Test predicted labels
+        L = L.todense()  # Note: Remove once Label Model L typing is cleaned up!
+        Y_lm = label_model.predict_proba(L).argmax(axis=1) + 1
+        err = np.where(Y != Y_lm, 1, 0).sum() / self.n
+        self.assertLess(err, 0.1)
 
 
 if __name__ == "__main__":
--- a/test/labeling/model/test_label_model.py
+++ b/test/labeling/model/test_label_model.py
@@ -1,13 +1,11 @@
 import unittest
-from typing import Tuple
 
 import numpy as np
 import pytest
 import torch
 import torch.nn as nn
-from scipy.sparse import csr_matrix, lil_matrix
+from scipy.sparse import csr_matrix
 
-from snorkel.labeling.analysis import lf_empirical_probs
 from snorkel.labeling.model.label_model import LabelModel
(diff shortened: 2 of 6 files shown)


test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
+        """Test the generated dataset for consistency."""
+        np.random.seed(123)
+        P, Y, L = self._generate_data()
+        P_emp = lf_empirical_probs(L, Y, k=self.k)
+        np.testing.assert_array_almost_equal(P, P_emp, decimal=2)
+
+    def test_label_model(self) -> None:
+        """Test the LabelModel's estimate of P and Y."""
+        np.random.seed(123)
+        P, Y, L = self._generate_data()
```
henryre: `P, _, L` to make it clear we aren't using `Y`

test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
+        label_model = LabelModel(k=self.k, verbose=False)
+        label_model.train_model(L)
```
henryre: Can you hardcode at least LR here in case defaults change?

test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
+class LabelModelTestAdvanced(unittest.TestCase):
+    """Advanced (marked complex) tests for the LabelModel."""
+
+    def setUp(self) -> None:
+        """Set constants for the tests."""
+        self.m = 10  # Number of LFs
+        self.n = 10000  # Number of data points
+        self.k = 2  # Number of classes
+
+    def _generate_data(self) -> Tuple[np.ndarray, np.ndarray, csr_matrix]:
```
henryre: I'm a little torn on this. A bit complicated to put in `test` and we want to make sure it works. What's the current usefulness of `snorkel/synthetics`? Should we burn that and replace it with simpler stuff like this>

test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
 
+@pytest.mark.complex
+class LabelModelTestAdvanced(unittest.TestCase):
+    """Advanced (marked complex) tests for the LabelModel."""
+
+    def setUp(self) -> None:
+        """Set constants for the tests."""
+        self.m = 10  # Number of LFs
+        self.n = 10000  # Number of data points
+        self.k = 2  # Number of classes
```
paroma: `self.k = self.cardinality`

test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
+        # Generate the true datapoint labels
+        # Note: Assuming balanced classes to start
+        Y = np.random.choice(self.k, self.n) + 1  # Note y \in {1,...,self.k}
+
+        # Generate the label matrix L
+        L = lil_matrix((self.n, self.m))
+        for i in range(self.n):
+            for j in range(self.m):
+                L[i, j] = np.random.choice(self.k + 1, p=P[j, :, Y[i] - 1])
+        L = L.tocsr()
```
paroma: any reason to sparsify? we convert to dense in the LabelModel each time

test/labeling/model/test_label_model.py
```diff
@@ -247,5 +250,63 @@ def test_train_model(self):
...
+        np.random.seed(123)
+        P, Y, L = self._generate_data()
+        P_emp = lf_empirical_probs(L, Y, k=self.k)
+        np.testing.assert_array_almost_equal(P, P_emp, decimal=2)
+
+    def test_label_model(self) -> None:
+        """Test the LabelModel's estimate of P and Y."""
+        np.random.seed(123)
+        P, Y, L = self._generate_data()
+        label_model = LabelModel(k=self.k, verbose=False)
```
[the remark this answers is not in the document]
ajratner: Fixed, added broader issue to asana

test/labeling/model/test_label_model.py
ajratner: Actually I meant to put in a test for the predicted labels against `Y`; putting in now.

test/labeling/model/test_label_model.py
ajratner: That's just temporary- we want to operate over sparse `L` in `LabelModel` soon enough, and so I think this is more representative input here!

test/labeling/model/test_label_model.py
ajratner: Sure makes sense- different generators for different things, but I'll at least split this (and the corresponding test) into `snorkel/synthetics`

ajratner:
@henryre @paroma ready for re-review!

henryre:
@ajratner looking good! can you just comment with the output of `tox -e py36 -- -m 'not complex and not spark' --durations 5?

snorkel/labeling/model/label_model.py
```diff
@@ -9,10 +9,9 @@
 
 from snorkel.analysis.utils import set_seed
 from snorkel.classification.utils import recursive_merge_dicts
-
-from .graph_utils import get_clique_tree
-from .lm_defaults import lm_default_config
-from .logger import Logger
+from snorkel.labeling.model.graph_utils import get_clique_tree
```
henryre: Why? we're allowing `.` imports, just not `..` ones

ajratner: VSC debugger throws an error with these kinds of relative imports if for example you're trying to debug a sub-function called from tests.  If we're encouraging using VSC debugger, nice to have these be full import statements

ajratner:
```
3.16s call     test/labeling/apply/test_lf_applier.py::TestDaskApplier::test_lf_applier_pandas_spacy_preprocessor_memoized
3.12s call     test/labeling/apply/test_lf_applier.py::TestDaskApplier::test_lf_applier_dask_spacy_preprocessor
2.97s call     test/slicing/test_slicing.py::SlicingTest::test_slicing
2.54s call     test/synthetic/test_synthetic_data.py::TestGenerateSimpleLabelMatrix::test_generate_L
2.46s call     test/synthetic/test_synthetic_data.py::TestGenerateSimpleLabelMatrix::test_generate_L_multiclass
=============================== 135 passed, 13 deselected, 31 warnings in 18.17 seconds ===============================
```

I'll cut these last two down

PR #1256, opened 2019-07-10
Remove application logic from TFApplier

henryre:
The idea here is that we don't want to have to serialize appliers to replicate application performance/outputs. That should be wrapped into the TFs and policy. Also adds an ApplyEach policy, since generating sets of sequences in the policy lets us include that type of logic.

**Test plan**
* Reran unit tests, including added
* Manually inspected augmentation coverage


snorkel/augmentation/policy/core.py
```diff
@@ -57,11 +94,48 @@ def generate(self) -> List[int]:
...
+
+    Parameters
+    ----------
+    n_tfs
+        Total number of TFs
+    keep_original
+        Keep untransformed data point in augmented data set? Note that
+        even if in-place modifications are made to the original data
+        point by the TFs being applied, the original data point will
+        remain unchanged.
```
vincentschen: Can we give examples? That will help differentiate between the other given policies—a quick glance over the class name is a bit confusing (`ApplyEachPolicy` vs. `ApplyOnePolicy`)

snorkel/augmentation/policy/sampling.py
```diff
@@ -22,14 +22,28 @@ class MeanFieldPolicy(Policy):
     p
         Probability distribution from which to sample TF indices.
         Must have length `n_tfs` and be a valid distribution.
+    n_per_original
+        Number of transformed data points per original
```
vincentschen: see clarification above

test/augmentation/policy/test_core.py
```diff
@@ -1,25 +1,19 @@
...
-        a_ct = samples.count([0, 0])
-        b_ct = samples.count([0, 1])
-        c_ct = samples.count([1, 0])
-        d_ct = samples.count([1, 1])
-        self.assertGreater(a_ct, 0)
-        self.assertGreater(b_ct, 0)
-        self.assertGreater(c_ct, 0)
-        self.assertGreater(d_ct, 0)
-        self.assertEqual(a_ct + b_ct + c_ct + d_ct, n_samples)
+class TestSamplingPolicy(unittest.TestCase):
```
vincentschen: these tests are clear examples — could include them in docstrings!

snorkel/augmentation/apply/core.py
```diff
@@ -21,101 +21,36 @@ class BaseTFApplier:
...
-            transform_applied = False
-            for tf_idx in self._policy.generate():
+            # Handle empty sequence for `keep_original`
+            transform_applied = len(seq) == 0
+            # Apply TFs
+            for tf_idx in seq:
                 tf = self._tfs[tf_idx]
                 x_t_raw = tf(x_t)
+                # Update if transformation was applied
                 if x_t_raw is not None:
```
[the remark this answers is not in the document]
henryre: if not able to transform due to bad type check. e.g. day of week example if no day of week in input text, or if a necessary field of input is missing, etc.

henryre:
@vincentschen ready for another look!

PR #1257, opened 2019-07-10
Add unit tests for SnorkelClassifier.predict(), score()

bhancock8:
- Adds basic tests for the predict() and score() methods
- Also fixes a bug discovered by the unit tests (yay unit tests!) when calculating abstain mask
- Moves filtering of examples with unknown gold labels to the metrics instead of the predict() method so that predict returns a label for every example in the dataset.

**Test plan:**
Adds unit test confirming that predict() and score() yield outputs of the right types and shapes.

### DIFF
--- a/snorkel/analysis/metrics.py
+++ b/snorkel/analysis/metrics.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Dict, List, NamedTuple
+from typing import Any, Callable, Dict, List, NamedTuple, Optional
 
 import numpy as np
 import sklearn.metrics as skmetrics
@@ -18,7 +18,7 @@ def metric_score(
     preds: np.ndarray,
     probs: np.ndarray,
     metric: str,
-    filter_dict: Dict[str, List[int]] = {"golds": [0]},
+    filter_dict: Optional[Dict[str, List[int]]] = None,
     **kwargs: Any,
 ) -> float:
     """A method for evaluating a standard metric on a set of predictions/probabilities
@@ -52,6 +52,9 @@ def metric_score(
     if metric not in METRICS:
         msg = f"The metric you provided ({metric}) is not currently implemented."
         raise ValueError(msg)
+
+    if filter_dict is None:
+        filter_dict = {"golds": [0]}
 
     # Convert to numpy
     golds = arraylike_to_numpy(golds) if golds is not None else None
--- a/snorkel/analysis/metrics.py
+++ b/snorkel/analysis/metrics.py
@@ -1,4 +1,4 @@
-from typing import Any, Callable, Dict, List, NamedTuple, Optional
+from typing import Any, Callable, Dict, List, NamedTuple
 
 import numpy as np
 import sklearn.metrics as skmetrics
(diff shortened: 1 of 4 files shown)


test/classification/test_snorkel_classifier.py
```diff
@@ -2,38 +2,83 @@
...
+        name=task_name,
+        module_pool=module_pool,
+        task_flow=task_flow,
+        scorer=Scorer(metrics=["accuracy"]),
+    )
+
+    return task
+
+
+task1 = create_task("task1", module_suffixes=["A", "A"])
```
henryre: Nit: caps in global scope

henryre: or make these attributes in setUp

test/classification/test_snorkel_classifier.py
```diff
@@ -45,12 +90,37 @@ def test_twotask_partial_overlap_model(self):
...
+        self.assertEqual(sorted(list(results.keys())), ["golds", "preds", "probs"])
+        self.assertEqual(results["preds"]["task1"].shape, (NUM_EXAMPLES,))
+
+    def test_empty_batch(self):
+        # Make the first BATCH_SIZE labels 0 so that one batch will have no labels
+        dataloader = create_dataloader("task1", shuffle=False)
+        for i in range(BATCH_SIZE):
+            dataloader.dataset.Y_dict["task1"][i] = 0
+        model = SnorkelClassifier([task1])
+        model.predict(dataloader)
```
henryre: What do we expect this to be?

test/classification/test_snorkel_classifier.py
```diff
@@ -45,12 +90,37 @@ def test_twotask_partial_overlap_model(self):
...
+
+    def test_score(self):
+        model = SnorkelClassifier([task1])
+        metrics = model.score([dataloader])
+        self.assertIsInstance(metrics["task1/dataset/train/accuracy"], float)
```
henryre: What should the accuracy be?

test/classification/test_snorkel_classifier.py
```diff
@@ -2,38 +2,83 @@
...
+    return dataloader
+
+
+def create_task(task_name, module_suffixes=("", "")):
+    module1_name = f"linear1{module_suffixes[0]}"
+    module2_name = f"linear2{module_suffixes[1]}"
+
+    module_pool = nn.ModuleDict(
+        {
+            module1_name: nn.Sequential(nn.Linear(2, 4), nn.ReLU()),
```
henryre: Is it worth it to manually set the weights so we know expected outputs of a forward pass? maybe just to the identity?

test/classification/test_snorkel_classifier.py
bhancock8: Unknown--it's a randomly initialized network.

test/classification/test_snorkel_classifier.py
bhancock8: Hmm, yeah. I'll give it a try.

bhancock8:
Ready for another look @henryre

snorkel/analysis/metrics.py
```diff
@@ -18,7 +18,7 @@ def metric_score(
     preds: np.ndarray,
     probs: np.ndarray,
     metric: str,
-    filter_dict: Optional[Dict[str, List[int]]] = None,
+    filter_dict: Dict[str, List[int]] = {"golds": [0]},
```
henryre: No mutables for defaults plz (https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments)

Note to self: this isn't caught by flake8, but we can with an add-on https://github.com/pycqa/flake8-bugbear

PR #1259, opened 2019-07-10
[EASY] Remove GPU code from cover

henryre:
Should get back to compliance


henryre:
@bhancock8 I tried both, this was the only one that worked 🤷‍♀

PR #1262, opened 2019-07-11
Add docstrings to data and drop verbose helper

bhancock8:
- Add docstrings to the classes in data.py
- Drop the split_data() helper. It's nice, but its functionality is _mostly_ redundant with scikit-learn's train_test_split() and it's a lot of lines of code for something that's not core to the library's value.

**Test plan:**
- pydocstyle passes
- unit tests pass

### DIFF
--- a/snorkel/classification/data.py
+++ b/snorkel/classification/data.py
@@ -27,17 +27,24 @@ Batch = Tuple[XDict, YDict]
 
 
 class DictDataset(Dataset):
-    """An advanced dataset class to handle input data with multipled fields and output
-    data with multiple label sets
+    """A dataset where both the data fields and labels are stored in as dictionaries.
 
-    :param name: the name of the dataset
-    :type name: str
-    :param X_dict: the feature dict where key is the feature name and value is the
-    feature
-    :type X_dict: dict
-    :param Y_dict: the label dict where key is the label name and value is
-    the label
-    :type Y_dict: dict
+    Parameters
+    ----------
+    name
+        The name of the dataset (e.g., this will be used to report metrics on a
+        per-dataset basis)
+    split
+        The name of the split that the data in this object represents
+    X_dict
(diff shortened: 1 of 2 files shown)


snorkel/classification/data.py
```diff
@@ -27,17 +12,24 @@
...
+    name
+        The name of the dataset (e.g., this will be used to report metrics on a
+        per-dataset basis)
+    split
+        The name of the split that the data in this object represents
+    X_dict
+        A map from field name to values (e.g., {"tokens": ..., "uids": ...})
+    Y_dict
+        A map from task name to its corresponding set of labels
+
```
henryre: Add Attributes (just "see above" is fine)

snorkel/classification/data.py
```diff
@@ -87,18 +90,16 @@ def collate_dicts(batch: List[Batch]) -> Batch:
...
-    :param split: str, optional
-    :param collate_fn: the function that merges a list of samples to form a
-    mini-batch, defaults to collate_dicts
-    :param collate_fn: function, optional
+    """A DataLoader that uses the appropriate collate_fn for a `DictDataset`.
+
+    Parameters
+    ----------
+    dataset
+        A DictDataset to wrap
```
henryre: Avoid having param descriptions that just repeat the type

snorkel/classification/data.py
```diff
@@ -87,18 +90,16 @@ def collate_dicts(batch: List[Batch]) -> Batch:
...
+    collate_fn
+        The collate function to use when combining multiple indexed examples for a
+        single batch. Usually the default collate_dicts() method should be used, but
+        it can be overriden if you want to use different collate logic.
```
henryre: kwargs?

bhancock8:
Re: split_data(): It was mostly a utility for general data prep, not core code. I think it's used in the tutorials, but we can update them.

PR #1263, opened 2019-07-11
Fix docstrings for snorkel/labeling/analysis.py

henryre:


snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
 Matrix = Union[np.ndarray, sparse.csr_matrix]
 
 
-############################################################
-# Label Matrix Diagnostics
-############################################################
 def _covered_data_points(L: Matrix) -> np.ndarray:
-    """Returns an indicator vector where ith element = 1 if x_i is labeled by at
-    least one LF."""
+    """Get indicator vector z where z_i = 1 if x_i is labeled by at least one LF."""
```
ajratner: Where is the full docstring here?  Should have input args, return args, and for this- since very user facing- an example

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
     return np.ravel(np.where(L.sum(axis=1) != 0, 1, 0))
 
 
 def _overlapped_data_points(L: Matrix) -> np.ndarray:
-    """Returns an indicator vector where ith element = 1 if x_i is labeled by
-    more than one LF."""
+    """Get indicator vector z where z_i = 1 if x_i is labeled by more than one LF."""
```
ajratner: Same comment as above

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
     return np.where(np.ravel((L != 0).sum(axis=1)) > 1, 1, 0)
 
 
 def _conflicted_data_points(L: sparse.spmatrix) -> np.ndarray:
-    """Returns an indicator vector where ith element = 1 if x_i is labeled by
-    at least two LFs that give it disagreeing labels."""
+    """Get indicator vector z where z_i = 1 if x_i is labeled differently by two LFs."""
```
ajratner: Same comment as above

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
+
+    Parameters
+    ----------
+    L
+        Matrix where L_{i,j} is the label given by the jth LF to the ith candidate
+
+    Returns
+    -------
+    float
+        Fraction of data points with labels
```
ajratner: I think adding quick examples for all of these good since very user facing- should make sure users get metrics

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
+
+    Parameters
+    ----------
+    L
+        Matrix where L_{i,j} is the label given by the jth LF to the ith candidate
+
+    Returns
+    -------
+    float
+        Fraction of data points with overlapping labels
```
ajratner: Example

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
+
+    Parameters
+    ----------
+    L
+        Matrix where L_{i,j} is the label given by the jth LF to the ith candidate
+
+    Returns
+    -------
+    float
+        Fraction of data points with conflicting labels
```
ajratner: Example

snorkel/labeling/analysis.py
```diff
@@ -11,87 +11,123 @@
...
-            jth LF to the ith candidate
+    Parameters
+    ----------
+    L
+        Matrix where L_{i,j} is the label given by the jth LF to the ith candidate
+
+    Returns
+    -------
+    List[List[int]]
+        Unique output labels for each LF
```
ajratner: Example- again I just worry that these all are not quite intuitive from our arbitrary internal naming alone, and should be crystal clear here

snorkel/labeling/analysis.py
henryre: It's a private method...

snorkel/labeling/analysis.py
henryre: Same response as above.

snorkel/labeling/analysis.py
henryre: Same response as above.

snorkel/labeling/analysis.py
ajratner: Put this in the contributing guidelines

PR #1264, opened 2019-07-11
Add slicing docstrings

vincentschen:
@bhancock8 would be particularly helpful to get a look from you given you wrote the first pass.

**Test Plan**: 
* Adds compliance with `pydocstyle snorkel/slicing`
* Ran `tox` locally; passes all tests.

### DIFF
--- a/snorkel/slicing/modules/slice_combiner.py
+++ b/snorkel/slicing/modules/slice_combiner.py
@@ -15,12 +15,21 @@ class SliceCombinerModule(nn.Module):
 
     Parameters
     ----------
-    slice_ind_key : optional
+    slice_ind_key
         Suffix of operation corresponding to the slice indicator heads
-    slice_pred_key : optional
+    slice_pred_key
         Suffix of operation corresponding to the slice predictor heads
-    slice_pred_feat_key : optional
+    slice_pred_feat_key
         Suffix of operation corresponding to the slice predictor features heads
+
+    Attributes
+    ----------
+    slice_ind_key
+        See above
+    slice_pred_key
+        See above
+    slice_pred_feat_key
+        See above
     """
 
     def __init__(
@@ -41,8 +50,9 @@ class SliceCombinerModule(nn.Module):
         Parameters
         ----------
         outputs
-            A dict of data fields containing indicator, predictor, and predictor
-            transform ops
+            A dict of data fields from slicing task flow containing specific keys
+            from indicator ops, pred ops, and pred transform ops (slice_ind_key,
+            slice_pred_key, slice_pred_feat_key) for each slice
 
         Returns
         -------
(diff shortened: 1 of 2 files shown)


snorkel/slicing/modules/slice_combiner.py
```diff
@@ -6,7 +6,7 @@
 
 
 class SliceCombinerModule(nn.Module):
-    """A module for combining the weighted representations learned by slices"""
+    """A module for combining the weighted representations learned by slices."""
```
henryre: Need to document the init params here!

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -21,6 +21,8 @@ def __init__(
         self.slice_pred_feat_key = slice_pred_feat_key
 
     def forward(self, outputs: Dict[str, torch.Tensor]) -> torch.Tensor:  # type: ignore
+        """Reweights and combines predictor representations."""
```
henryre: Add parameters and returns sections

henryre:
Oh jk, just realized this was a draft

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -6,7 +6,22 @@
...
+    """A module for combining the weighted representations learned by slices.
+
+    Intended for use with task flow including:
+        * Indicator operations
+        * Prediction operations
+        * Prediction transform features
+
+    Parameters
+    ----------
+    slice_ind_key : optional
```
henryre: rm optional

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -6,7 +6,22 @@
...
+        Suffix of operation corresponding to the slice indicator heads
+    slice_pred_key : optional
+        Suffix of operation corresponding to the slice predictor heads
+    slice_pred_feat_key : optional
+        Suffix of operation corresponding to the slice predictor features heads
+    """
```
henryre: Add Attributes. Even just "see above" is fine

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -21,7 +36,22 @@ def __init__(
         self.slice_pred_feat_key = slice_pred_feat_key
 
     def forward(self, outputs: Dict[str, torch.Tensor]) -> torch.Tensor:  # type: ignore
+        """Reweights and combines predictor representations given output dict.
+
+        Parameters
+        ----------
+        outputs
+            A dict of data fields containing indicator, predictor, and predictor
```
henryre: A little vague. Are the keys special?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -67,12 +103,15 @@ def forward(self, outputs: Dict[str, torch.Tensor]) -> torch.Tensor:  # type: ig
...
         )
 
+        # Attention weights used to combine each of the slice_representations
         A = (
+            # Combine the indicator (whether we are in the slice or not) and
+            # predictor (confidence of a learned slice head) as attention weights
             F.softmax(indicator_preds * predictor_preds, dim=1)
-            .unsqueeze(-1)
-            .expand([-1, -1, slice_representations.size(-1)])
+            # Match the dimensions of the slice_representations
```
henryre: Hard to read like this. can we just multiline?

```python
A = F.softmax(indicator_preds * predictor_preds, dim=1)
# Comment
A = A.unsqueeze(-1).expand([-1, -1, slice_representations.size(-1)])
```

snorkel/slicing/modules/slice_combiner.py
vincentschen: how come? there are default parameters?

henryre: anything that's also specified by the type hints (besides parameter name) should go, since Sphinx will parse it for us and we don't want to maintain it in two places. this is also imprecise, since `typing.Optional` means something else.

henryre: guessing it was added by autodocstring, so we should see if we can disable

PR #1270, opened 2019-07-11
Add docstrings to Trainer and Schedulers

bhancock8:
See title. Note that most of the Trainer methods are private methods, often just glorified setters, so they're fine with a one-linear comment about what's happening.

### DIFF
--- a/snorkel/classification/training/trainer.py
+++ b/snorkel/classification/training/trainer.py
@@ -23,10 +23,33 @@ Metrics = Dict[str, float]
 
 
 class Trainer:
-    """A class for multi-task learning.
+    """A class for training a SnorkelClassifier.
 
-    :param config: The learning config
-    :type config: dict
+    Parameters
+    ----------
+    name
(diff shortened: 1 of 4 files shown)


snorkel/classification/training/trainer.py
```diff
@@ -369,13 +398,15 @@ def _logging(
         return metric_dict
 
     def _log_metrics(self, metric_dict: Metrics) -> None:
+        """Write metrics."""
```
henryre: This is less useful than the function name

snorkel/classification/training/trainer.py
```diff
@@ -195,7 +223,7 @@ def _set_log_manager(self) -> None:
         )
 
     def _set_optimizer(self, model: nn.Module) -> None:
-        """Set optimizer for learning process."""
+        """Set the optimizer for updating parameters."""
```
henryre: Less useful than function name

PR #1271, opened 2019-07-11
Pin tool versions

henryre:
These can be updated whenever, but will probably be pretty static. Just don't want any unexpected behavior between e.g. Travis and local

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -53,10 +53,9 @@ add-ignore =
     D105,
     D107,
     D202,
-    D203,
     D204,
-    D213,
     D413,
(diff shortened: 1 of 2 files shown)


setup.cfg
```diff
@@ -53,10 +53,9 @@ add-ignore =
     D105,
     D107,
     D202,
-    D203,
     D204,
-    D213,
     D413,
+    D416,
```
vincentschen: what is this? couldn't find it here: http://www.pydocstyle.org/en/2.1.1/error_codes.html

PR #1272, opened 2019-07-12
Naively fix synthetic docstrings

vincentschen:
Test Plan: bring `pydocstyle snorkel/synthetic` to compliance.

Naive, probably needs another pass to clarify.


vincentschen:
> 🚢

Will do post tests!

PR #1274, opened 2019-07-12
Initial doc build

henryre:
* Patches and initial configuration for ReadTheDocs
* Switches to double backticks in docstrings
* Switches Example format to doctest format

**Known issues**
* Using type hints requires installing requirements on machine that builds the docs, since mocks and type hints don't work together very well (https://github.com/agronholm/sphinx-autodoc-typehints/issues/70). RTD machines have a disk space limit, and we can't get all of our reqs on. So the types get rendered incorrectly in a few places (e.g. https://snorkel.readthedocs.io/en/rtd/source/snorkel.labeling.apply.html#snorkel.labeling.apply.dask.DaskLFApplier). The only long-term fix I know of is to host our own API docs.
* We could automate the Sphinx build (or check correctness) via Travis, but this should change infrequently, so adding to backlog.

**Test plan**
https://snorkel.readthedocs.io/en/rtd/index.html


henryre:
@ajratner @bhancock8 @paroma ready for a look. looks huge, but mostly just ` -> ``

PR #1275, opened 2019-07-12
Slicing refactor + fixes

vincentschen:
- [x] Bugfix: `add_slice_tasks` was missing some of the `slice_ind` or `slice_pred` operations when adding to master_combiner's`task_flow`
- [x] SliceCombiner now uses the max score across pred classes as the "confidence"
- [x] Refactor SlicePred into helper functions 
- [x] Cleanup slice task naming (e.g. `task_slice:pred_pred_head` --> `task_slice:pred_head`)
- [x] Fix deprecated pytorch warnings

**Test Plan**: 
- [x] Add tighter tests for convergence (> 95 F1-score) (downside: takes slightly longer to run tests)
- [x] Add additional unit tests for specific SliceCombiner functionality
    - SliceCombiner test cases for different attention weights
    - Check that `task_flow` and `module_pool` elements are correctly configured

### DIFF
--- a/snorkel/slicing/modules/slice_combiner.py
+++ b/snorkel/slicing/modules/slice_combiner.py
@@ -44,9 +44,7 @@ class SliceCombinerModule(nn.Module):
         self.slice_pred_key = slice_pred_key
         self.slice_pred_feat_key = slice_pred_feat_key
 
-    def forward(
-        self, flow_dict: Dict[str, torch.Tensor]
-    ) -> torch.Tensor:  # type: ignore
+    def forward(self, flow_dict: Dict[str, torch.Tensor]) -> torch.Tensor: # type:ignore
         """Reweights and combines predictor representations given output dict.
 
         Parameters
--- a/snorkel/slicing/modules/slice_combiner.py
+++ b/snorkel/slicing/modules/slice_combiner.py
@@ -23,6 +23,9 @@ class SliceCombinerModule(nn.Module):
         Suffix of operation corresponding to the slice predictor heads
     slice_pred_feat_key
         Suffix of operation corresponding to the slice predictor features heads
+    tau
+        Temperature constant for scaling the weighting between indicator prediction
+        and predictor confidences: SoftMax(indicator_pred * predictor_confidence / tau)
 
     Attributes
     ----------
@@ -39,12 +42,14 @@ class SliceCombinerModule(nn.Module):
         slice_ind_key: str = "_ind_head",
         slice_pred_key: str = "_pred_head",
         slice_pred_feat_key: str = "_pred_transform",
+        tau: int = 0.1,
     ) -> None:
         super().__init__()
 
         self.slice_ind_key = slice_ind_key
         self.slice_pred_key = slice_pred_key
         self.slice_pred_feat_key = slice_pred_feat_key
+        self.tau = tau
 
     def forward(  # type:ignore
         self, flow_dict: Dict[str, List[torch.Tensor]]
@@ -102,7 +107,7 @@ class SliceCombinerModule(nn.Module):
         # Attention weights used to combine each of the slice_representations
         # incorporates the indicator (whether we are in the slice or not) and
         # predictor (confidence of a learned slice head)
-        A = F.softmax(indicator_preds * predictor_confidences, dim=1)
+        A = F.softmax(indicator_preds * predictor_confidences / self.tau, dim=1)
 
         # Match dims [batch_size x num_slices x feat_dim] of slice_representations
         A = A.unsqueeze(-1).expand([-1, -1, slice_representations.size(-1)])
--- a/snorkel/slicing/modules/slice_combiner.py
+++ b/snorkel/slicing/modules/slice_combiner.py
@@ -106,21 +106,20 @@ class SliceCombinerModule(nn.Module):
 
         # Concatenate each predictor feature into [batch_size x 1 x feat_dim] tensor
         slice_representations = torch.cat(
-            [
-                outputs[slice_feat_name][0].unsqueeze(1)
-                for slice_feat_name in slice_feat_names
-            ],
-            dim=1,
+            [outputs[slice_feat_name][0] for slice_feat_name in slice_feat_names], dim=1
         )
+
+        # Compute the "confidence" [bach_size x 1] of the prediction using the max score
+        predictor_confidence = torch.max(predictor_preds, dim=1).values.unsqueeze(-1)
 
         # Attention weights used to combine each of the slice_representations
         # incorporates the indicator (whether we are in the slice or not) and
         # predictor (confidence of a learned slice head)
-        A = F.softmax(indicator_preds * predictor_preds, dim=1)
+        A = torch.sigmoid(indicator_preds * predictor_confidence)
 
-        # Match the dimensions of the slice_representations
-        A = A.unsqueeze(-1).expand([-1, -1, slice_representations.size(-1)])
+        # Expand weights and match dims [bach_size x feat_dim] of slice_representations
+        A = A.expand([-1, slice_representations.size(-1)])
 
-        # Reweight representations by class Sum across all classes
-        reweighted_rep = torch.sum(A * slice_representations, dim=1)
+        # Reweight representations with elementwise product
+        reweighted_rep = A * slice_representations
         return reweighted_rep
--- a/snorkel/slicing/modules/slice_combiner.py
+++ b/snorkel/slicing/modules/slice_combiner.py
@@ -1,4 +1,4 @@
-from typing import Dict
+from typing import Dict, List
 
 import torch
 import torch.nn as nn
@@ -60,60 +60,37 @@ class SliceCombinerModule(nn.Module):
             The reweighted predictor representation
         """
 
-        # Gather names of slice heads (both indicator and predictor heads)
-        # This provides a static ordering by which to index into the 'outputs' dict
-        slice_ind_op_names = sorted(
-            [
-                flow_name
-                for flow_name in outputs.keys()
-                if self.slice_ind_key in flow_name
-            ]
+        # Concatenate indicator head predictions into tensor [batch_size x num_slices]
+        indicator_outputs = self._collect_flow_outputs_by_key(
+            outputs, self.slice_ind_key
         )
-        slice_pred_op_names = sorted(
-            [
-                flow_name
-                for flow_name in outputs.keys()
-                if self.slice_pred_key in flow_name
-            ]
+        indicator_preds = torch.cat(
+            [F.softmax(output)[:, 0].unsqueeze(1) for output in indicator_outputs],
+            dim=-1,
         )
 
-        # Concatenate the predictions from the predictor head/indicator head
-        # into a [batch_size x num_slices] tensor
-        indicator_preds = torch.cat(
-            [
-                F.softmax(outputs[slice_ind_name][0])[:, 0].unsqueeze(1)
-                for slice_ind_name in slice_ind_op_names
-            ],
-            dim=-1,
+        # Concatenate predictor head confidences into tensor [batch_size x num_slices]
+        predictor_outputs = self._collect_flow_outputs_by_key(
+            outputs, self.slice_pred_key
         )
         predictor_confidences = torch.cat(
             [
                 F.softmax(
                     # Compute the "confidence" using the max score across classes
-                    torch.max(outputs[slice_pred_name][0], dim=1).values
+                    torch.max(output, dim=1)[0]
                 ).unsqueeze(1)
-                for slice_pred_name in slice_pred_op_names
+                for output in predictor_outputs
             ],
             dim=-1,
         )
 
-        # Collect names of predictor "features" that will be combined into the final
-        # reweighted representation
-        slice_feat_names = sorted(
-            [
-                flow_name
-                for flow_name in outputs.keys()
-                if self.slice_pred_feat_key in flow_name
-            ]
+        # Concatenate each predictor feature (to be combined into reweighted
+        # representation) into [batch_size x 1 x feat_dim] tensor
+        predictor_feat_outputs = self._collect_flow_outputs_by_key(
+            outputs, self.slice_pred_feat_key
         )
-
-        # Concatenate each predictor feature into [batch_size x 1 x feat_dim] tensor
         slice_representations = torch.cat(
-            [
-                outputs[slice_feat_name][0].unsqueeze(1)
-                for slice_feat_name in slice_feat_names
-            ],
-            dim=1,
+            [output.unsqueeze(1) for output in predictor_feat_outputs], dim=1
         )
 
         # Attention weights used to combine each of the slice_representations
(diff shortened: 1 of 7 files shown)


snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
-    def forward(self, outputs: Dict[str, torch.Tensor]) -> torch.Tensor:  # type: ignore
+    def forward(  # type:ignore
+        self, flow_dict: Dict[str, torch.Tensor]
+    ) -> torch.Tensor:
         """Reweights and combines predictor representations given output dict.
 
         Parameters
         ----------
-        outputs
+        flow_dict
```
henryre: Hm the description is very insider-y. Can we make it easier to parse for a deeper who isn't familiar with slicing?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
-        # Match the dimensions of the slice_representations
+        # Match dims [bach_size x num_slices x feat_dim] of slice_representations
         A = A.unsqueeze(-1).expand([-1, -1, slice_representations.size(-1)])
 
-        # Reweight representations by class Sum across all classes
+        # Reweight representations with weighted sum across slices
         reweighted_rep = torch.sum(A * slice_representations, dim=1)
         return reweighted_rep
+
+    def _collect_flow_outputs_by_key(
```
henryre: Should this be module level? No self attributes accessed

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
+        self, flow_dict: Dict[str, torch.Tensor], key: str
+    ) -> List[torch.Tensor]:
+        """Return flow_dict outputs specified by key, ordered by sorted flow_name."""
+        return [
+            flow_dict[flow_name][0]
+            for flow_name in sorted(flow_dict.keys())
+            if key in flow_name
```
henryre: if we're saying suffix in the docstring, should this be .endswith?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
+        # Reweight representations with weighted sum across slices
         reweighted_rep = torch.sum(A * slice_representations, dim=1)
         return reweighted_rep
+
+    def _collect_flow_outputs_by_key(
+        self, flow_dict: Dict[str, torch.Tensor], key: str
+    ) -> List[torch.Tensor]:
+        """Return flow_dict outputs specified by key, ordered by sorted flow_name."""
+        return [
+            flow_dict[flow_name][0]
```
henryre: Are these actually Tensors? Why are we just grabbing the first dim?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
-            ],
-            dim=1,
+            [output.unsqueeze(1) for output in predictor_feat_outputs], dim=1
         )
 
         # Attention weights used to combine each of the slice_representations
         # incorporates the indicator (whether we are in the slice or not) and
         # predictor (confidence of a learned slice head)
-        A = F.softmax(indicator_preds * predictor_preds, dim=1)
+        A = F.softmax(indicator_preds * predictor_confidences, dim=1)
```
henryre: softmax of an element-wise product of two [0, 1] bounded tensors feels like it should have a temperature

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
 
-        # Match the dimensions of the slice_representations
+        # Match dims [bach_size x num_slices x feat_dim] of slice_representations
```
henryre: nit: typo

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -44,83 +44,79 @@ def __init__(
...
         )
-
-        # Concatenate each predictor feature into [batch_size x 1 x feat_dim] tensor
         slice_representations = torch.cat(
-            [
-                outputs[slice_feat_name][0].unsqueeze(1)
-                for slice_feat_name in slice_feat_names
-            ],
-            dim=1,
+            [output.unsqueeze(1) for output in predictor_feat_outputs], dim=1
```
henryre: torch.stack should work for this

snorkel/slicing/utils.py
```diff
@@ -128,7 +130,9 @@ def convert_to_slice_tasks(base_task: Task, slice_names: List[str]) -> List[Task
         ind_head_op = Operation(
             module_name=ind_head_module_name, inputs=head_module_op.inputs
         )
-        ind_task_flow = body_flow + [ind_head_op]
+        ind_task_ops = [ind_head_op]
+        ind_task_flow = body_flow + ind_task_ops
+        slice_task_ops.extend(ind_task_ops)
```
henryre: nit: could just extend() then append()

snorkel/slicing/utils.py
```diff
@@ -190,26 +195,23 @@ def convert_to_slice_tasks(base_task: Task, slice_names: List[str]) -> List[Task
...
 
     master_task = Task(
         name=master_task_name,
         module_pool=master_module_pool,
         task_flow=master_task_flow,
         scorer=base_task.scorer,
     )
-    tasks.append(master_task)
-
+    tasks = slice_tasks + [master_task]
```
henryre: nit: just return here

test/slicing/test_convergence.py
```diff
@@ -87,27 +87,27 @@ def test_convergence(self):
         model = SnorkelClassifier(tasks=tasks)
 
         # Train
-        trainer = Trainer(lr=0.01, n_epochs=30, progress_bar=False)
+        trainer = Trainer(lr=0.001, n_epochs=50, progress_bar=False)
```
henryre: Hm I don't love fiddling with hyperparams in tests. Makes it feel brittle. Is this due to a change in this diff or just globally better settings?

snorkel/slicing/modules/slice_combiner.py
vincentschen: alternative is as a more general helper function in `snorkel/classifier/utils`?

snorkel/slicing/modules/slice_combiner.py
vincentschen: This is a convention of the `flow_dict` (cc @bhancock8) —each contains a mapping form a `flow_name` to a _list_ of potential inputs (i.e. you might have tokens, masks, segments as in BERT).

In most cases, there's a single element, which is why we can index  with `[0]`

henryre: Ok then the type hint is incorrect right?

snorkel/slicing/modules/slice_combiner.py
vincentschen: does [pytorch](https://pytorch.org/docs/stable/nn.html?highlight=softmax#torch.nn.functional.softmax) include this as a parameter? 

also, do you have some intuition about what $\tau$ to choose?

test/slicing/test_convergence.py
vincentschen: They seemed a bit globally better (they are consistent across the 2 integration tests now). Earlier I tried a higher `lr` to reach convergence more quickly for shortened tests, but I'm prioritizing tighter performance bounds (>95F1) for now—we can shorten the tests down the line if necessary.

snorkel/slicing/modules/slice_combiner.py
vincentschen: `flow_dict[flow_name][0]` is a `torch.Tensor` and we return as a list comprehension `[flow_dict[flow_name][0] for ...]`, so type is `List[torch.Tensor]`

henryre: But `flow_dict: Dict[str, torch.Tensor]` should be `flow_dict: Dict[str, List[torch.Tensor]]`?

snorkel/slicing/modules/slice_combiner.py
henryre: No you'd just add it as an init param to this module then do 

```python
F.softmax(indicator_preds * predictor_confidences / self.tau, dim=1)
```

and I have no idea! how peaky do we want this thing?

snorkel/slicing/modules/slice_combiner.py
vincentschen: Yes you're right. :)

vincentschen: Yes you're right. :)

snorkel/slicing/utils.py
vincentschen: sorry I'm not following—extend which?

snorkel/slicing/modules/slice_combiner.py
vincentschen: done^

snorkel/slicing/modules/slice_combiner.py
vincentschen: defaulting to 0.1 for now to make it peakier by default... can surface this as a hyperparameter in applications / tutorials

snorkel/slicing/utils.py
henryre:
```python
slice_task_ops.extend(body_flow)
slice_task_ops.append(ind_head_op)
```

a bit clearer to me

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -34,93 +39,79 @@ class SliceCombinerModule(nn.Module):
 
     def __init__(
         self,
-        slice_ind_key: str = "_ind",
-        slice_pred_key: str = "_pred",
+        slice_ind_key: str = "_ind_head",
+        slice_pred_key: str = "_pred_head",
         slice_pred_feat_key: str = "_pred_transform",
+        tau: float = 0.1,
```
henryre: Are all the reported results with tau = 1.0? might want to default to that

test/slicing/test_slice_combiner.py
```diff
@@ -11,23 +12,191 @@ class SliceCombinerTest(unittest.TestCase):
...
             ],
             "task_slice:base_pred_head": [
+                torch.FloatTensor(batch_size, num_classes).uniform_(0, 1)
+            ],
+        }
+        combiner_module = SliceCombinerModule()
+        combined_rep = combiner_module(outputs)
+        self.assertEqual(tuple(combined_rep.shape), (batch_size, h_dim))
+
+        # edge case (all ones)
```
henryre: Great!!

test/slicing/test_slice_combiner.py
```diff
@@ -11,23 +12,191 @@ class SliceCombinerTest(unittest.TestCase):
...
+    def test_many_slices(self):
+        """Test combiner on 1000 synthetic generated slices."""
+
+        batch_size = 4
+        h_dim = 20
+        num_classes = 2
+
+        outputs = {}
+        for i in range(1000):
+            if i % 2 == 0:
```
henryre: sooooo what's going on here? half super confident 4 predictions, half super confident 2 predictions? maybe add a note. and can we get away with 100?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -21,6 +23,9 @@ class SliceCombinerModule(nn.Module):
         Suffix of operation corresponding to the slice predictor heads
     slice_pred_feat_key
         Suffix of operation corresponding to the slice predictor features heads
+    tau
```
henryre: maybe just `temperature`? no strong opinion

test/slicing/test_slice_combiner.py
```diff
@@ -11,23 +12,191 @@ class SliceCombinerTest(unittest.TestCase):
...
+            "task_slice:a_pred_head": [torch.zeros(batch_size, num_classes)],
+            "task_slice:b_ind_head": [torch.ones(batch_size, 2) * -10.0],
+            "task_slice:b_pred_transform": [torch.ones(batch_size, h_dim) * 2],
+            "task_slice:b_pred_head": [torch.zeros(batch_size, num_classes)],
+        }
+        combiner_module = SliceCombinerModule()
+        combined_rep = combiner_module(outputs)
+        self.assertTrue(torch.allclose(combined_rep, torch.ones(batch_size, h_dim) * 3))
+
+    def test_average_reweighting_by_pred_confidence(self):
```
henryre: Add a test for temperature? Maybe just super high so you can take the max

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -34,93 +39,79 @@ class SliceCombinerModule(nn.Module):
...
+        """Reweight and combine predictor representations given output dict.
 
         Parameters
         ----------
-        outputs
-            A dict of data fields from slicing task flow containing specific keys
-            from indicator ops, pred ops, and pred transform ops (slice_ind_key,
-            slice_pred_key, slice_pred_feat_key) for each slice
+        flow_dict
+            A dict of data fields cached operation outputs from indicator head,
```
bhancock8: typo: "data fields cached operation outputs"

snorkel/slicing/modules/slice_combiner.py
bhancock8: Perhaps an Example would be the best way to do it?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -34,93 +39,79 @@ class SliceCombinerModule(nn.Module):
...
+        predictor_feat_outputs = collect_flow_outputs_by_suffix(
+            flow_dict, self.slice_pred_feat_key
         )
+        slice_representations = torch.stack(predictor_feat_outputs, dim=1)
 
         # Attention weights used to combine each of the slice_representations
         # incorporates the indicator (whether we are in the slice or not) and
         # predictor (confidence of a learned slice head)
-        A = F.softmax(indicator_preds * predictor_preds, dim=1)
+        A = F.softmax(indicator_preds * predictor_confidences / self.tau, dim=1)
```
bhancock8: Is there a more meaningful name we can give A? How about attention_weights?

snorkel/slicing/utils.py
vincentschen: Can't do that because we need `task_flow` to define the `Task`.
I cleaned up the comments a bit to segment the steps a bit more: 
```
        # Define operations for task head
        pred_transform_op = Operation(
            module_name=pred_transform_module_name, inputs=head_module_op.inputs
        )
        pred_head_op = Operation(
            module_name=pred_head_module_name, inputs=[(pred_transform_op.name, 0)]
        )
        pred_task_ops = [pred_transform_op, pred_head_op]
        slice_task_ops.extend(pred_task_ops)

        # Create task flow
        pred_task_flow = body_flow + pred_task_ops

        # Create pred task
```

test/slicing/test_slice_combiner.py
vincentschen: yes half confident in each way; more to test whether the model can handle the scale

vincentschen: 100 slices will work; it isn't much more costly to initialize more

vincentschen:
@henryre ready for another look! 
* Main diff is the new test for temperature: https://github.com/HazyResearch/snorkel/pull/1275/files#diff-148196efe3fdc85436429da44bf5ec95R208

PR #1276, opened 2019-07-12
Change config style from dicts to namedtuples

bhancock8:
Update how configs are specified:
- Default values are specified in `NamedTuple`s
- `merge_config()` is a much simpler version of `recursive_merge_dict()` that converts nested dicts of settings into nested Configs (see example below).
- Mypy has one issue with using the NamedTuple configs: it complains about empty initializations (i.e., using all the defaults, as in `TrainerConfig()`), so we `type: ignore` those lines.

`Classifer`
- [x] Divide snorkel_config into separate `Config` objects
- [x] Write `merge_config()` for nested `Config`s
- [x] Update calls (attributes instead of keys)
- [x] Update types
- [x] Update tests
- [x] Delete snorkel_config.py
- [x] Write Config docstrings

Example
-------
```
config_updates = {
    "n_epochs": 5,
    "optimizer_config": {
        "lr": 0.001,
    }
}
trainer_config = merge_config(TrainerConfig(), config_updates)
```

This diff is large enough; LabelModel will be updated in a subsequent PR.

**Test plan**
Tests are updated to the new syntax and continue to pass post-refactor.
New test added for merge_config()


bhancock8:
@henryre and @vincentschen are probably the two best to view this.

henryre:
Offline: replace `merge_config` with something like `Config._from_dict(...)`

snorkel/classification/training/loggers/checkpointer.py
```diff
@@ -2,54 +2,82 @@
...
+    checkpoint_metric
+        The metric to checkpoint on, of the form "task/dataset/split/metric:mode" where
+        mode is "min" or "max".
+    checkpoint_task_metrics
+        Additional metrics to save best models for. Note that the best model according
+        to `checkpoint_metric` will be the one that is loaded after training and used
+        for early stopping.
+    checkpoint_runway
+        No checkpointing will occur for the first this many checkpoint_units
+    checkpoint_clear
```
vincentschen: could be misleading naming; maybe `checkpoint_clear_except_best`? or something less verbose...

snorkel/classification/training/loggers/log_writer.py
```diff
@@ -71,7 +85,7 @@ def write_config(
         config_filename
             Name of file in logging directory to write to
         """
-        self.write_json(config, config_filename)
+        self.write_json(config._asdict(), config_filename)
```
vincentschen: does `asdict()` still count as a private function?

snorkel/classification/training/trainer.py
```diff
@@ -1,27 +1,171 @@
...
+    eps: float = 1e-8
+
+
+class OptimizerConfig(Config):
+    """Settings common to all optimizers."""
+
+    lr: float = 0.001
+    l2: float = 0.0
+    grad_clip: float = 1.0
+    sgd_config: SGDOptimizerConfig = SGDOptimizerConfig()  # type:ignore
```
vincentschen: this feels a little confusing, given we will only ever have one of these optimizers. 
Can we subclass?

henryre: or a Union

snorkel/classification/training/loggers/log_writer.py
henryre: Nope! It's documented: https://docs.python.org/3/library/collections.html#collections.somenamedtuple._asdict

Underscore is to prevent collision with a field name

snorkel/classification/snorkel_classifier.py
```diff
@@ -21,15 +21,31 @@
...
+    ----------
+    device
+        The device (GPU) to move the model to (-1 is CPU), but device will also be
+        moved to CPU if no GPU device is available
+    dataparallel
+        Whether or not to use PyTorch DataParallel wrappers to automatically utilize
+        multiple GPUs if available
+    """
+
+    device: int = 0
```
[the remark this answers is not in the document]
bhancock8: I think if GPUs are available, the vast majority of the time, people want to use the GPU. And if it's not available, it defaults to the CPU anyway, so nothing will break. I think leaving it at GPU 0 falls under the category of "good defaults"

snorkel/classification/training/loggers/checkpointer.py
bhancock8: I think the usage of checkpoint_clear is fuzzy at best. It only matters in cases where you want to checkpoint on multiple metrics but you don't intend to ever use the checkpoints saved on the other metrics (if you just care about tracking their progress, you can do that separately from the checkpointer). Added a note to the tracker to remove this option, but saving it for a separate PR, since it's unrelated to the config.

snorkel/classification/training/trainer.py
bhancock8: I agree it'd be nice for both the LRScheduler and the Optimizer to not appear to support more than "child" class's parameters, but I think there are issues with both approaches suggested here. When instantiating the AdamOptimizer, for example, the algorithm needs to know what kwargs that optimizer accepts, and unioning the fields would lose that information. And if we subclass OptimizerConfig for the three children, the Trainer still needs to contain all three options so that there's a path to update their values when the Trainer is initialized. I think what we should actually do is just collect "optimizer_kwargs" and "lrscheduler_kwargs" and then pass those to whatever optimizer/scheduler they indicated. Again though, that's orthogonal to this PR of changing the typing of the configs, so I'm going to add that to the tracker for a subsequent update.

bhancock8:
Turns out, "As of version 0.4, this code [the dataclasses backport to 3.6] no longer works with Python 3.7" (https://github.com/ericvsmith/dataclasses)...so no dataclasses after all if we want to support both 3.6 and 3.7 with the same code. So I stuck with typing.NamedTuple for the Configs. I added tracker tasks for some of the design updates we agree we want to make to the Checkpointer/Optimizer/LRScheduler configs that became more apparent once we broke them out and added docstrings, but proposing we merge this in its current stable state to unblock.

snorkel/classification/training/trainer.py
vincentschen: sg, let's track in a separate PR!

snorkel/classification/utils.py
```diff
@@ -73,6 +75,39 @@ def pad_batch(
...
+    }
+    trainer_config = merge_config(TrainerConfig(), config_updates)
+    ```
+    """
+    for key, value in config_updates.items():
+        if isinstance(value, dict):
+            config_updates[key] = merge_config(getattr(config, key), value)
+    return config._replace(**config_updates)
+
+
```
henryre: WAIT CAN'T WE DELETE `RMD`??

bhancock8: It's still used by LabelModel. After LabelModel gets updated, then off with it's head!

PR #1283, opened 2019-07-16
Remove ArrayLike

henryre:
Get rid of `ArrayLike` and `arraylike_to_numpy`, cleaning up internal types significantly. Lots of little changes necessary. If we find specific places where a specific converter helper function would be useful we can add in later.

**Test plan**
* Added unit test for `to_int_label_array`
* Patched and reran existing unit tests

### DIFF
--- a/snorkel/analysis/utils.py
+++ b/snorkel/analysis/utils.py
@@ -46,12 +46,10 @@ def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
     return np.eye(num_classes)[preds.squeeze() - 1]
 
 
-def to_flattened_int_array(
-    X: np.ndarray, flatten: bool = True, cast_to_int: bool = True
-) -> np.ndarray:
-    """Convert an array to a flattened vector of ints.
+def to_int_label_array(X: np.ndarray, flatten: bool = True) -> np.ndarray:
+    """Convert an array to a (possibly flattened) array of ints.
 
-    Flatten [n, 1] arrays to [n] and cast all values to ints.
+    Cast all values to ints and possibly flatten [n, 1] arrays to [n].
     This method is typically used to sanitize labels before use with analysis tools or
     metrics that expect 1D arrays as inputs.
 
@@ -61,8 +59,6 @@ def to_flattened_int_array(
         An array to possibly flatten and possibly cast to int
     flatten
         If True, flatten array into a 1D array
-    cast_to_int
-        If True, cast all values to ints
 
     Returns
     -------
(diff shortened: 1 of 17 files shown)


test/labeling/model/test_baseline.py
```diff
@@ -11,7 +12,7 @@
 
 class BaselineModelTest(unittest.TestCase):
     def test_random_vote(self):
-        L = np.array([[1, 2, 1], [0, 4, 3], [3, 0, 0], [1, 2, 2]])
+        L = csr_matrix(np.array([[1, 2, 1], [0, 4, 3], [3, 0, 0], [1, 2, 2]]))
```
brahmaneya: Instead of having np.array as an intermediate step, why not construct the csr_matrix directly from the list of lists? (same comment for other two occurrences in this file).

snorkel/analysis/utils.py
```diff
@@ -49,71 +46,53 @@ def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
...
-    else:
-        msg = f"Input of type {orig_type} could not be converted to an np.ndarray"
-        raise ValueError(msg)
-
     # Correct shape
     if flatten:
-        if (array_like.ndim > 1) and (1 in array_like.shape):
-            array_like = array_like.flatten()
-        if array_like.ndim != 1:
+        if (X.ndim > 1) and (1 in X.shape):
```
brahmaneya: use X.squeeze() instead of this if block.

Also, I think the 'flatten' argument should be renamed. Flatten generally refers to making the shape 1D indpendent of inital shape, while here you are also checking that there is only one non-1 dimension.

snorkel/analysis/utils.py
```diff
@@ -49,71 +46,47 @@ def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
...
+    if np.any(np.not_equal(np.mod(X, 1), 0)):
+        raise ValueError("Input contains at least one non-integer value.")
+    X = X.astype(np.dtype(int))
 
     # Correct shape
     if flatten:
-        if (array_like.ndim > 1) and (1 in array_like.shape):
-            array_like = array_like.flatten()
-        if array_like.ndim != 1:
+        if (X.ndim > 1) and (1 in X.shape):
```
brahmaneya: use X.squeeze() instead of this if block.

Also, I think the 'flatten' argument should be renamed. Flatten generally refers to making the shape 1D indpendent of inital shape, while here you are also checking that there is only one non-1 dimension.

test/labeling/model/test_baseline.py
brahmaneya: Instead of having np.array as an intermediate step, why not construct the csr_matrix directly from the list of lists? (same comment for other two occurrences in this file).

snorkel/analysis/utils.py
henryre: Hm, how about `flatten_vector`? To imply that if we flatten, it needs to already be vector-like?

snorkel/analysis/utils.py
```diff
@@ -49,71 +46,44 @@ def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
...
-        array_like = array_like.toarray()
-    elif isinstance(array_like, torch.Tensor):
-        array_like = array_like.numpy()
-    elif not isinstance(array_like, np.ndarray):
-        array_like = np.array(array_like)
-    else:
-        msg = f"Input of type {orig_type} could not be converted to an np.ndarray"
-        raise ValueError(msg)
-
+    if np.any(np.not_equal(np.mod(X, 1), 0)):
```
bhancock8: Can we just check the type of the numpy array instead?

henryre: isn't the point that we cast to int if it's a float array but they're all actually integers?

PR #1284, opened 2019-07-16
Labeling test full

brahmaneya:
Added integration test for labeling. It runs the following
1) Preprocesser
2) Labeling functions
3) LabelModel training
4) LabelModel use (predict_proba)

**Test Plan**
No changes to non-testing code. Added new test which is passing.


brahmaneya:
Closing because I accidentally targeted master instead of redux.

PR #1286, opened 2019-07-17
Add `predict(), score()` to LabelModel, add new tie breaking option

paroma:
* Add `predict()` to LabelModel
* Add `score()` function to LabelModel

* Add `break_ties()` to `snorkel/analysis/utils.py` to help convert probabilistic labels to integer labels
     * Add `random` and `true-random` options for tie breaking: first is truly random and second is deterministically random
     * use `prob_to_pred` in LabelModel `predict()` function

__Test Plan__
* Add tests for `predict()` and `score()` in LabelModel tests
* Add tests for `tie_break()` in utils tests
* Add tests for both random options
* Fix `snorkel/classifier` tests to match deterministic random tie breaking results

### DIFF
--- a/snorkel/analysis/utils.py
+++ b/snorkel/analysis/utils.py
@@ -12,7 +12,57 @@ def set_seed(seed: int) -> None:
     torch.manual_seed(seed)
 
 
-def probs_to_preds(probs: np.ndarray) -> np.ndarray:
+def break_ties(
+    Y_prob: np.ndarray, tie_break_policy: Optional[str] = "random"
+) -> np.ndarray:
+    """Break ties among probabilistic labels according to given policy.
+
+    Policies to break ties include:
+    "abstain": return an abstain vote (0)
+    "true-random": randomly choose among the tied options
+    "random": randomly choose among tied option using deterministic hash
+
+    NOTE: if tie_break_policy="true-random", repeated runs may have slightly different results due to difference in broken ties
+
+    Parameters
+    ----------
+    Y_prob
+        An [n,k] array of probabilistic labels
+    tie_break_policy
+        Policy to break ties, by default 'random'
+
+    Returns
+    -------
+    np.ndarray
+        An [n,1] array of integer labels
+    """
+
+    n, k = Y_prob.shape
+    Y_pred = np.zeros(n)
+    diffs = np.abs(Y_prob - Y_prob.max(axis=1).reshape(-1, 1))
+
+    TOL = 1e-5
+    for i in range(n):
+        max_idxs = np.where(diffs[i, :] < TOL)[0]
+        if len(max_idxs) == 1:
+            Y_pred[i] = max_idxs[0] + 1
+        # Deal with "tie votes" according to the specified policy
+        elif tie_break_policy == "random":
+            Y_pred[i] = max_idxs[i % len(max_idxs)] + 1
+        elif tie_break_policy == "true-random":
+            Y_pred[i] = np.random.choice(max_idxs) + 1
+        elif tie_break_policy == "abstain":
+            Y_pred[i] = 0
+        else:
+            raise ValueError(
+                f"tie_break_policy={tie_break_policy} policy not recognized."
+            )
+    return Y_pred
+
+
+def probs_to_preds(
+    probs: np.ndarray, tie_break_policy: Optional[str] = "random"
+) -> np.ndarray:
     """Convert an array of probabilistic labels into an array of predictions.
 
     Parameters
@@ -20,13 +70,15 @@ def probs_to_preds(probs: np.ndarray) -> np.ndarray:
     prob
         A [num_datapoints, num_classes] array of probabilistic labels such that each
         row sums to 1.
+    tie_break_policy
+            Policy to break ties when converting probabilistic labels to predictions, by default 'abstain'
 
     Returns
     -------
     np.ndarray
         A [num_datapoints, 1] array of predictions (integers in [1, ..., num_classes])
     """
-    return np.argmax(probs, axis=1) + 1
+    return break_ties(probs, tie_break_policy)
 
 
 def preds_to_probs(preds: np.ndarray, num_classes: int) -> np.ndarray:
--- a/snorkel/analysis/utils.py
+++ b/snorkel/analysis/utils.py
@@ -13,7 +13,7 @@ def set_seed(seed: int) -> None:
     torch.manual_seed(seed)
 
 
-def _hash(i):
+def _hash(i: int) -> int:
     """Deterministic hash function."""
     byte_string = str(i).encode("utf-8")
     return int(hashlib.sha1(byte_string).hexdigest(), 16)
--- a/snorkel/analysis/utils.py
+++ b/snorkel/analysis/utils.py
@@ -19,10 +19,10 @@ def _hash(i: int) -> int:
     return int(hashlib.sha1(byte_string).hexdigest(), 16)
 
 
-def break_ties(
-    Y_prob: np.ndarray, tie_break_policy: Optional[str] = "random"
+def probs_to_preds(
+    probs: np.ndarray, tie_break_policy: Optional[str] = "random", tol: float = 1e-5
 ) -> np.ndarray:
-    """Break ties among probabilistic labels according to given policy.
+    """Convert an array of probabilistic labels into an array of predictions.
 
     Policies to break ties include:
     "abstain": return an abstain vote (0)
@@ -33,24 +33,33 @@ def break_ties(
 
     Parameters
     ----------
-    Y_prob
-        An [n,k] array of probabilistic labels
+    prob
+        A [num_datapoints, num_classes] array of probabilistic labels such that each
+        row sums to 1.
     tie_break_policy
-        Policy to break ties, by default 'random'
+        Policy to break ties when converting probabilistic labels to predictions, by default 'abstain'
+    tol
+        The minimum difference among probabilities to be considered a tie, by deafult 1e-5
+
 
     Returns
     -------
     np.ndarray
-        An [n] array of integer labels
+        A [n] array of predictions (integers in [1, ..., num_classes])
+
+    Examples
+    --------
+    >>> probs_to_preds(np.array([[0.5, 0.5, 0.5]]), tie_break_policy="abstain")
+    array([0])
+    >>> probs_to_preds(np.array([[0.8, 0.1, 0.1]]))
+    array([1])
     """
-
-    n, k = Y_prob.shape
+    n, k = probs.shape
     Y_pred = np.zeros(n)
-    diffs = np.abs(Y_prob - Y_prob.max(axis=1).reshape(-1, 1))
+    diffs = np.abs(probs - probs.max(axis=1).reshape(-1, 1))
 
-    TOL = 1e-5
     for i in range(n):
-        max_idxs = np.where(diffs[i, :] < TOL)[0]
+        max_idxs = np.where(diffs[i, :] < tol)[0]
         if len(max_idxs) == 1:
             Y_pred[i] = max_idxs[0] + 1
         # Deal with "tie votes" according to the specified policy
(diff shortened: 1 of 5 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -368,10 +369,124 @@ def predict_proba(self, L: sparse.spmatrix) -> np.ndarray:
...
+            elif break_ties == "random":
+                Y_pred[i] = np.random.choice(max_idxs) + 1
+            elif break_ties == "abstain":
+                Y_pred[i] = 0
+            else:
+                raise ValueError(f"break_ties={break_ties} policy not recognized.")
+        return Y_pred
+
+    def predict(
+        self, L: sparse.spmatrix, break_ties: Optional[str] = "abstain"
```
bhancock8: I'm torn here. If the default is abstain, then there's a potential sneaky bug where a person gets way more wrong than they should (since the gold will almost never be abstain). But if the default is random, then a person not knowing that we have random tie-breaking could see that their score is stochastic and freak out. MajorityLabelVoter() just uses vanilla `argmax`, which will take the first option if there's a tie. Could we do the same here and make that the default? If multiple classes are tied, it returns the first class in the tie?

paroma: We could pass in an option for a label to default to in case of a tie, say always the negative class for example. Returning the first class probably makes sense to match MV, but it feels like introducing weird biases?

snorkel/labeling/model/label_model.py
```diff
@@ -354,7 +355,7 @@ def predict_proba(self, L: sparse.spmatrix) -> np.ndarray:
         >>> label_model = LabelModel()
         >>> label_model.train_model(L)
         >>> label_model.predict_proba(L)
-        np.array([1.0, 0.0], [0.0, 1.0], [1.0, 0.0])
+        np.array([[1.0, 0.0], [0.0, 1.0], [1.0, 0.0]])
```
henryre: just a note: you can test these with doctest (https://docs.python.org/3/library/doctest.html). not set up to run automatically with pytest yet, but it's on the list.

snorkel/labeling/model/label_model.py
```diff
@@ -368,10 +369,124 @@ def predict_proba(self, L: sparse.spmatrix) -> np.ndarray:
...
+            A dictionary mapping metric names to metric scores
+
+        Example
+        -------
+        >>> L = sparse.csr_matrix([[1, 1, 0], [2, 2, 0], [1, 1, 0]])
+        >>> label_model = LabelModel()
+        >>> label_model.train_model(L)
+        >>> label_model.score(L, Y=np.array([1, 1, 1]))
+        {'accuracy': 0.66667}
+        >>> label_model.score(L, Y=np.array([1, 1, 1], metrics=["f1"]))
```
henryre: Can we show one where f1 != accuracy?

test/labeling/model/test_label_model.py
```diff
@@ -218,6 +218,59 @@ def test_predict_proba(self):
...
+        label_model.mu = nn.Parameter(label_model.mu_init.clone())
+
+        results = label_model.score(csr_matrix(L), Y=np.array([1, 2]))
+        results_expected = dict(accuracy=0.5)
+        self.assertEqual(results, results_expected)
+
+        results = label_model.score(
+            L=csr_matrix(L), Y=np.array([2, 1]), metrics=["accuracy", "f1"]
+        )
+        results_expected = dict(accuracy=0.5, f1=2 / 3.0)
```
henryre: nit: `2 / 3` works

snorkel/labeling/model/label_model.py
```diff
@@ -368,10 +369,124 @@ def predict_proba(self, L: sparse.spmatrix) -> np.ndarray:
...
+        {'accuracy': 0.66667}
+        """
+        Y_prob = self.predict_proba(L)
+        Y_pred = self.predict(L)
```
henryre: This will compute predict_prob twice. Can we add a `return_probs` flag to `predict()`?

test/labeling/model/test_label_model.py
```diff
@@ -218,6 +218,59 @@ def test_predict_proba(self):
...
+        label_model = LabelModel(k=3)
+        probs = np.array([[0.33, 0.33, 0.33]])
+        preds = label_model._break_ties(probs, break_ties="abstain")
+        true_preds = np.array([0.0])
+        np.testing.assert_array_equal(preds, true_preds)
+
+        # random with ties
+        probs = np.array([[0.33, 0.33, 0.33]])
+        random_preds = []
+        for seed in range(10):
```
henryre: What invariant of LabelModel are we trying to test here?

paroma: whether the random tie breaking is truly random

snorkel/labeling/model/label_model.py
paroma: Created `random` and `true-random` options for tie breaking policy. true-random is non-deterministic and `random` is deterministic random

snorkel/analysis/utils.py
```diff
@@ -12,21 +12,73 @@ def set_seed(seed: int) -> None:
...
+    diffs = np.abs(Y_prob - Y_prob.max(axis=1).reshape(-1, 1))
+
+    TOL = 1e-5
+    for i in range(n):
+        max_idxs = np.where(diffs[i, :] < TOL)[0]
+        if len(max_idxs) == 1:
+            Y_pred[i] = max_idxs[0] + 1
+        # Deal with "tie votes" according to the specified policy
+        elif tie_break_policy == "random":
+            Y_pred[i] = max_idxs[i % len(max_idxs)] + 1
```
bhancock8: Clever! This is much lighter-weight than hashing the whole label array.

snorkel/analysis/utils.py
```diff
@@ -12,21 +12,73 @@ def set_seed(seed: int) -> None:
...
+    Y_prob: np.ndarray, tie_break_policy: Optional[str] = "random"
+) -> np.ndarray:
+    """Break ties among probabilistic labels according to given policy.
+
+    Policies to break ties include:
+    "abstain": return an abstain vote (0)
+    "true-random": randomly choose among the tied options
+    "random": randomly choose among tied option using deterministic hash
+
+    NOTE: if tie_break_policy="true-random", repeated runs may have slightly different results due to difference in broken ties
```
henryre: Great!

snorkel/analysis/utils.py
henryre: Hmmmm is there any way to check if the original label matrix row here was all zeros? That's probably the only time we want to hash the index instead of the row of labels. For example, if two data points both had label outputs [0, 1, 2] and lf_2_acc == lf_3_acc, I'd want the predictions for them to be the same. If they're all zeros, then sure, hash the ID so all of the uncovered points don't have the same class.

snorkel/analysis/utils.py
```diff
@@ -12,21 +12,73 @@ def set_seed(seed: int) -> None:
...
+    ----------
+    Y_prob
+        An [n,k] array of probabilistic labels
+    tie_break_policy
+        Policy to break ties, by default 'random'
+
+    Returns
+    -------
+    np.ndarray
+        An [n,1] array of integer labels
```
henryre: Nit: it's [n]

snorkel/analysis/utils.py
```diff
@@ -12,21 +12,73 @@ def set_seed(seed: int) -> None:
...
+        elif tie_break_policy == "abstain":
+            Y_pred[i] = 0
+        else:
+            raise ValueError(
+                f"tie_break_policy={tie_break_policy} policy not recognized."
+            )
+    return Y_pred
+
+
+def probs_to_preds(
```
henryre: Do probs_to_preds need to be public functions?

henryre: As in, do we want them to show up in the API docs?

snorkel/analysis/utils.py
paroma: is it weird to always prefer a class over the other in this case? say in the binary case, we have `probs=[0.5, 0.5]` and it's due to the LF accuracies that voted on this datapoint being the same. Then we'd end up with all datapoint having class=1 (or class=2), which doesn't necessarily reflect what "random" would be

snorkel/analysis/utils.py
bhancock8: I think so. It's a frequent enough operation, especially since a lot of other models are likely to output probabilities in certain circumstances, and it provides an easy way for people to use the same tie-breaking policies as us so that it's apples to apples. (Also, with our current 1-indexed convention, this method is what performs the +1 offset that makes an [n, 2] matrix of probs become 1s and 2s instead of 0s and 1s).

snorkel/analysis/utils.py
bhancock8: Yeah, I think I like the pseudo-random distribution. The intuitive check I'm using is "if I have 100 times, they should be split approximately 50/50 by class if I'm doing random tie-breaking", just in a way that's deterministic (which this achieves).

bhancock8: i.e., I like the way it currently is in this PR.

henryre: ok then to at least get rid of the cyclical artifacts, let's do `hash(str(i)) % n_ties`

snorkel/labeling/model/label_model.py
```diff
@@ -637,7 +731,7 @@ def save(self, destination: str, **kwargs: Any) -> None:
 
         Example
         -------
-        >>> label_model.save('./saved_label_model')
+        >>> label_model.save('./saved_label_model')  # doctest: +SKIP
```
henryre: Looks like the base is incorrect. A lot of these doctest fixes are already in redux, e.g. https://github.com/HazyResearch/snorkel/blob/redux/snorkel/labeling/model/label_model.py#L643

test/analysis/test_utils.py
```diff
@@ -47,6 +49,50 @@ def test_to_int_label_array(self):
     def test_pred_to_prob(self):
         np.testing.assert_array_equal(preds_to_probs(PREDS, 2), PREDS_ROUND)
 
+    def test_hash(self):
```
henryre: Don't think we need this...not testing an actual property of the implementation. Deterministic-ness is tested in test_break_ties.

snorkel/analysis/utils.py
```diff
@@ -12,21 +13,79 @@ def set_seed(seed: int) -> None:
     torch.manual_seed(seed)
 
 
-def probs_to_preds(probs: np.ndarray) -> np.ndarray:
+def _hash(i: int) -> int:
```
henryre: Great!!

snorkel/labeling/model/label_model.py
paroma: This was the manual fix to make sure the branch doesn’t break. Rebase says it’s updated, I could git pull but that’ll push all the other commits in her too

henryre: can you rebase or cherry-pick instead of merge? you can also set the pull default to rebase https://coderwall.com/p/tnoiug/rebase-by-default-when-doing-git-pull

snorkel/analysis/utils.py
```diff
@@ -12,21 +13,79 @@ def set_seed(seed: int) -> None:
...
+    -------
+    np.ndarray
+        An [n] array of integer labels
+    """
+
+    n, k = Y_prob.shape
+    Y_pred = np.zeros(n)
+    diffs = np.abs(Y_prob - Y_prob.max(axis=1).reshape(-1, 1))
+
+    TOL = 1e-5
```
henryre: nit: all-caps for scoped variable is odd, maybe make a function parameter with this as a default?

snorkel/analysis/utils.py
```diff
@@ -12,21 +13,79 @@ def set_seed(seed: int) -> None:
...
+    tie_break_policy
+            Policy to break ties when converting probabilistic labels to predictions, by default 'abstain'
 
     Returns
     -------
     np.ndarray
         A [num_datapoints, 1] array of predictions (integers in [1, ..., num_classes])
     """
-    return np.argmax(probs, axis=1) + 1
+    return break_ties(probs, tie_break_policy)
```
henryre: Why do we need these two functions if one just calls the other?

snorkel/labeling/model/label_model.py
```diff
@@ -370,10 +371,101 @@ def predict_proba(self, L: sparse.spmatrix) -> np.ndarray:
...
+        Example
+        -------
+        >>> L = sparse.csr_matrix([[1, 1, 0], [2, 2, 0], [1, 1, 0]])
+        >>> label_model = LabelModel(verbose=False)
+        >>> label_model.train_model(L)
+        >>> label_model.predict(L)
+        array([1, 2, 1])
+        """
+        Y_probs = self.predict_proba(L)
+        Y_p = probs_to_preds(Y_probs, tie_break_policy).astype(np.int)
```
henryre: Should the int cast be part of probs_to_preds? Is a float array a valid output?

snorkel/analysis/utils.py
paroma: removing break_ties

PR #1292, opened 2019-07-18
Fix SlicingConvergenceTest

bhancock8:
- Update setUpClass to be recognized by unittest
- Update Trainer intialization to get nested dict (`recursive_merge_dict` is no longer available to find the appropriate depth to update `lr`)

**Test plan:**
Complex tests pass again (were broken).

### DIFF
--- a/test/slicing/test_convergence.py
+++ b/test/slicing/test_convergence.py
@@ -45,7 +45,7 @@ def h(x: DataPoint) -> int:
 
 class SlicingConvergenceTest(unittest.TestCase):
     @classmethod
-    def setup_class(cls):
+    def setUpClass(cls):
         # Ensure deterministic runs
         set_seed(123)
 
@@ -87,7 +87,9 @@ class SlicingConvergenceTest(unittest.TestCase):
         model = SnorkelClassifier(tasks=tasks)
 
         # Train
-        trainer = Trainer(lr=0.001, n_epochs=50, progress_bar=False)
+        trainer = Trainer(
+            **{"optimizer_config": {"lr": 0.001}}, n_epochs=50, progress_bar=False
+        )
         trainer.train_model(model, dataloaders)
         scores = model.score(dataloaders)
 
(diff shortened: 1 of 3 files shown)


test/slicing/test_convergence.py
```diff
@@ -45,7 +45,7 @@ def h(x: DataPoint) -> int:
 
 class SlicingConvergenceTest(unittest.TestCase):
     @classmethod
-    def setup_class(cls):
+    def setUpClass(cls):
```
henryre: Did you grep for any other instances?

test/slicing/test_convergence.py
```diff
@@ -87,7 +87,9 @@ def test_convergence(self):
         model = SnorkelClassifier(tasks=tasks)
 
         # Train
-        trainer = Trainer(lr=0.001, n_epochs=50, progress_bar=False)
+        trainer = Trainer(
+            **{"optimizer_config": {"lr": 0.001}}, n_epochs=50, progress_bar=False
```
henryre: Can't we just do `optimizer_config={"lr": 0.001}`?

test/slicing/test_convergence.py
```diff
@@ -138,7 +140,9 @@ def test_performance(self):
 
         # Train
         # NOTE: Needs more epochs to convergence with more heads
-        trainer = Trainer(lr=0.001, n_epochs=110, progress_bar=False)
+        trainer = Trainer(
+            **{"optimizer_config": {"lr": 0.001}}, n_epochs=110, progress_bar=False
```
henryre: Same

test/slicing/test_convergence.py
vincentschen: it seemed to work running `pytest test/slicing`— is convention then to run `python test/slicing/test_convergence.py`?

bhancock8: Did now. :) Sounds like `pytest test/.../mytest.py` works with `setup_class` or `setUpClass` but when travis runs it only `setUpClass` works. @brahmaneya @vincentschen

henryre: travis just runs pytest so not sure why that would happen. might be due to the coverage driver? but doubt that's the reason. you can try `tox -e coverage` to see if that replicates.

PR #1295, opened 2019-07-19
[EASY] Fix encrypted keys to use --org encryption

vincentschen:
Turns out, [snorkel lives on travis.org](https://travis-ci.org/search/snorkel)...
Another shot at a fix, by encrypting keys using `--org` flag 

> Note that if you’re still using travis-ci.org you need to use --org instead of --com in all of the commands shown on this page.

— https://docs.travis-ci.com/user/notifications/#Slack-notifications


bhancock8:
Nice. I'm not sure that we actually `always` want an email for every success or failure, but I assume this is just so we can test more quickly?

vincentschen:
> Nice. I'm not sure that we actually `always` want an email for every success or failure, but I assume this is just so we can test more quickly?

Right, that was the intention—but let's change it so that the Slack/email notifications should match 1:1.

PR #1294, opened 2019-07-19
Add doctest to check docstring examples

henryre:
Use `pytest-doctestplus` to check our docstring examples. Fixes a few docstring examples to comply.

Also cleans up logger which had two identical functions.

**Test plan**
Manually check Travis execution

PR #1296, opened 2019-07-19
Add doc build

henryre:
Lots of autogen, but only real changes are in `tox.ini`. Adds an environment so you can update the API docs with `tox -e doc`.


paroma:
Is there a way to turn it off for certain docs? Or add exceptions for print statements and rounding for numerical outputs? Otherwise it may be too picky

henryre:
@paroma  there are a couple examples of this in label_model.py

henryre:
@paroma also did you mean to post that on #1294?

PR #1299, opened 2019-07-19
Add metrics helpers

bhancock8:
People (ourselves included) are going to want to use the metrics methods outside of the scorer, and currently that requires an argument for golds, preds, _and_ probs, when no metric actually ever requires both preds and probs. These two skinny wrappers let the user be explicit about what they're passing and reduce extra/unnecessary passing of `None`s.

**Test plan:**
New unit tests confirm that they work as expected and raise errors when they should.

### DIFF
--- a/snorkel/analysis/error_analysis.py
+++ b/snorkel/analysis/error_analysis.py
@@ -20,9 +20,9 @@ def error_buckets(
 
     Parameters
     ----------
-    gold
+    golds
         An np.ndarray of gold (int) labels
-    pred
+    preds
         An np.ndarray of (int) predictions
     X
         Optional, a sequence of examples corresponding to golds/preds
@@ -32,8 +32,7 @@ def error_buckets(
     -------
     Dict
         A mapping of each error bucket to its corresponding indices/examples
-        If X is None, return indices
-            instead.
+        If X is None, return indices instead.
     """
     buckets: Mapping[Tuple[int, int], List[Any]] = defaultdict(list)
     golds = to_int_label_array(golds)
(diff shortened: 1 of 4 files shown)


henryre:
How necessary is metric_score now that we've cleaned up the numerical types? What does it do besides a call to `to_int_label_array`?

paroma:
Perhaps in the doctoring we have a note that says which of the default metrics are used with which one

bhancock8:
The most important thing it does is pull out the arguments that each sklearn metric expects (some combination of golds, probs, preds, but never all three). It's what lets you specify a list of metrics to calculate by name (like ["accuracy", "coverage", "roc_auc"]) without having to pass three different functions three different signatures. It also allows you to say "don't penalize abstain votes for LFs", which we usually want to do. I think we can probably drop the int conversions now though, and maybe support Paroma's suggestion as well of calculating preds if probs are given so it stays just one function.

bhancock8:
Made the changes we talked about. Turns out the `to_int_label_array()` method is still rather helpful as a format checker (informing when labels are of the wrong size or type). But I was able to toss the two new methods; now all three label types (golds, preds, probs) are optional but it will check and complain if it doesn't have the label types required by the metric you requested. And if you pass in probs when it needs preds, it can take care of that conversion for you automatically with good 'ole `probs_to_preds`.

bhancock8:
Changed my mind. explicit > implicit, so removing the probs to preds auto conversion, but keeping the flexible of leaving all three types optional and then just complaining if they didn't pass what their metrics needs.

PR #1303, opened 2019-07-22
Fix get_accuracies() function

paroma:
* Remove user option to pass in conditional probabilities (highly unlikely to have prior knowledge of this) for `get_accuracies`
* Adjust accuracy by accounting for coverage
* add `L_train` as attribute for `LabelModel` (for ease of computation)

__Test Plan__
* Add comprehensive LF accuracy test
* Add <100% coverage LFs for accuracy test (preview. version was not testing <100% coverage LF with L matrix input)

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -279,13 +279,11 @@ class LabelModel(nn.Module):
         else:
             return c_probs
 
-    def get_accuracies(self, L: Union[np.ndarray, sparse.spmatrix], probs: Optional[np.ndarray] = None) -> np.ndarray:
+    def get_accuracies(self) -> np.ndarray:
         """Return the vector of LF accuracies.
 
         Parameters
         ----------
-        probs
-            Conditional probabilities, by default None
         L
             An [n,m] matrix with values in {0,1,...,k}
 
@@ -299,21 +297,16 @@ class LabelModel(nn.Module):
         >>> L = np.array([[0, 0, -1], [1, 1, -1], [0, 0, -1]])
         >>> label_model = LabelModel(verbose=False)
         >>> label_model.train_model(L)
-        >>> np.around(label_model.get_accuracies(L), 2)
-        array([0.9 , 0.9 , 0.01])
+        >>> np.around(label_model.get_accuracies(), 2)
+        array([0.9, 0.9])
         """
         accs = np.zeros(self.m)
         for i in range(self.m):
-            if probs is None:
-                cps = self._get_conditional_probs(source=i)[1:, :]
-            else:
-                cps = probs[
-                    i * (self.cardinality + 1) : (i + 1) * (self.cardinality + 1)
-                ][1:, :]
+            cps = self._get_conditional_probs(source=i)[1:, :]
             accs[i] = np.diag(cps @ self.P.numpy()).sum()
 
-        coverage = np.mean(L != 0, axis=0)
-        return accs / coverage
+        coverage = np.mean(np.asarray(self.L_train) != 0, axis=0)
+        return np.clip(accs / coverage, 0.01, 1.0)
 
     def predict_proba(self, L: np.ndarray) -> np.ndarray:
         r"""Return label probabilities P(Y | \lambda).
@@ -632,6 +625,7 @@ class LabelModel(nn.Module):
         self._set_class_balance(class_balance, Y_dev)
         self._set_constants(L_shift)
         self._create_tree()
+        self.L_train = L_train
 
         # Compute O and initialize params
         if self.config["verbose"]:  # pragma: no cover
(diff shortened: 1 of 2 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -304,13 +304,13 @@ def _get_conditional_probs(self, source: Optional[int] = None) -> np.ndarray:
...
 
-    def get_accuracies(self, probs: Optional[np.ndarray] = None) -> np.ndarray:
+    def get_accuracies(self) -> np.ndarray:
         """Return the vector of LF accuracies.
 
         Parameters
         ----------
-        probs
-            Conditional probabilities, by default None
+        L
```
henryre: ?

test/labeling/model/test_label_model.py
```diff
@@ -14,6 +13,27 @@
 
 
 class LabelModelTest(unittest.TestCase):
+    def _create_L(
```
henryre: Why is this a class method? Confusing since it's only the L for one test

test/labeling/model/test_label_model.py
```diff
@@ -14,6 +13,27 @@
...
+        self, accs=[0.95, 0.6, 0.7, 0.55, 0.8], coverage=[1.0, 0.8, 1.0, 1.0, 1.0]
```
henryre: Having these totally random (and mutable!) defaults is odd

test/labeling/model/test_label_model.py
```diff
@@ -14,6 +13,27 @@
...
+
+        L = np.zeros((1000, len(accs)))
+        Y = np.zeros(1000)
+
+        for i in range(1000):
+            Y[i] = 1 if np.random.rand() <= 0.5 else 2
+
+            for j in range(5):
+                if np.random.rand() <= coverage[j]:
+                    L[i, j] = Y[i] if np.random.rand() <= accs[j] else opp(Y[i])
```
henryre: nit: get rid of scoped function

`-1 * Y[i] + 3`

snorkel/labeling/model/label_model.py
```diff
@@ -319,22 +319,19 @@ def get_accuracies(self, probs: Optional[np.ndarray] = None) -> np.ndarray:
...
-                cps = self._get_conditional_probs(source=i)[1:, :]
-            else:
-                cps = probs[
-                    i * (self.cardinality + 1) : (i + 1) * (self.cardinality + 1)
-                ][1:, :]
+            cps = self._get_conditional_probs(source=i)[1:, :]
             accs[i] = np.diag(cps @ self.P.numpy()).sum()
-        return accs
+
+        coverage = np.mean(np.asarray(self.L_train) != 0, axis=0)
```
henryre: self.L_train.toarray()

henryre: Can also get from snorkel.labeling.analysis

snorkel/labeling/model/label_model.py
```diff
@@ -654,6 +651,7 @@ def train_model(
         self._set_class_balance(class_balance, Y_dev)
         self._set_constants(L_train)
         self._create_tree()
+        self.L_train = L_train
```
henryre: Let's compute coverage here via snorkel.labeling.analysis rather than carrying around the matrix

paroma:
The accuracy tests didn't have LFs with <100% accuracy when testing without passing in probs (did include abstains in second test, but we also passed in conditional probs explicitly in that case). Removing probs forces testing using only differences in L structure

snorkel/labeling/model/label_model.py
```diff
@@ -304,37 +305,28 @@ def _get_conditional_probs(self, source: Optional[int] = None) -> np.ndarray:
...
-                cps = self._get_conditional_probs(source=i)[1:, :]
-            else:
-                cps = probs[
-                    i * (self.cardinality + 1) : (i + 1) * (self.cardinality + 1)
-                ][1:, :]
+            cps = self._get_conditional_probs(source=i)[1:, :]
             accs[i] = np.diag(cps @ self.P.numpy()).sum()
-        return accs
+
+        return np.clip(accs / self.coverage, 0.01, 1.0)
```
henryre: Wait what's this for?

henryre: the clip

paroma: Sometimes the accuracies/coverage can become 1.0001 etc - to prevent that. There’s a similar value clip for cond. probs as well

henryre: The lower bound is the one I don't understand. If there's some divide-by-zero concern, let's at least do 1e-6

henryre:
@paroma ok great! so the new `test_get_accuracy` would have caught the bug?

snorkel/labeling/model/label_model.py
```diff
@@ -279,37 +280,28 @@ def _get_conditional_probs(self, source: Optional[int] = None) -> np.ndarray:
...
-
         Returns
         -------
         np.ndarray
             [m,1] vector of LF accuracies
 
         Example
         -------
-        >>> L = np.array([[0, 0, -1], [1, 1, -1], [0, 0, -1]])
+        >>> L = np.array([[0, 0], [1, 1], [0, 0]])
```
henryre: Do we want some abstains here?

PR #1304, opened 2019-07-22
[EASY] Fix LabelModel print statement

bhancock8:
Add parentheses so LabelModel prints more than once.

### DIFF
--- a/snorkel/labeling/model/logger.py
+++ b/snorkel/labeling/model/logger.py
@@ -20,7 +20,7 @@ class Logger:
 
     def __init__(self, log_train_every: int) -> None:
         self.log_train_every = log_train_every
-        self.unit_count = 0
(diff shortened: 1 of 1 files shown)


snorkel/labeling/model/logger.py
```diff
@@ -31,7 +31,7 @@ def check(self) -> bool:
             Whether to log or not based on logging frequency
         """
         self.unit_count += 1
-        return self.unit_count - 1 % self.log_train_every == 0
+        return (self.unit_count - 1) % self.log_train_every == 0
```
paroma: can we do `return (self.unit_count) % self.log_train_every == 0` here and start with `send.unit_count = -1` in the `__init__` function? then it prints at epoch 0, 10, ... instead of 1, 11, ...

PR #1305, opened 2019-07-22
Update Logger.log() doctest

bhancock8:
Update docstring to reflect that `Logger.unit_count == -1` at first now unless we run a `check()` operation first.

**Test plan:**
tox -e doctest passes now

### DIFF
--- a/snorkel/labeling/model/logger.py
+++ b/snorkel/labeling/model/logger.py
@@ -45,13 +45,6 @@ class Logger:
         ------
         Exception
             If metric names formatted incorrectly
-
-        Example
-        -------
-        >>> logger = Logger(log_train_every=5)
-        >>> metrics_dict = {"train/loss": 5.00}
-        >>> logger.log(metrics_dict)
-        [-1 epochs]: TRAIN:[loss=5.000]
         """
         score_strings: DefaultDict[str, List[str]] = defaultdict(list)
         for full_name, value in metrics_dict.items():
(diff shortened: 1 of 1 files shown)


snorkel/labeling/model/logger.py
```diff
@@ -50,7 +50,7 @@ def log(self, metrics_dict: Dict[str, float]) -> None:
         -------
         >>> logger = Logger(log_train_every=5)
         >>> metrics_dict = {"train/loss": 5.00}
-        >>> logger.log(metrics_dict)
+        >>> if logger.check(): logger.log(metrics_dict)
```
henryre: This is extremely bizarre. Not an example usage. Looks like we hacked this in to pass a doctest.

bhancock8: `logger.log()` directly reports `[-1 epochs]:`, which is odd, but it will never be called in the wild without `logger.check()` being called first. We can simplify the doctest case; they'll just see a printout that they shouldn't ever see in real usage.

snorkel/labeling/model/logger.py
```diff
@@ -51,7 +51,7 @@ def log(self, metrics_dict: Dict[str, float]) -> None:
         >>> logger = Logger(log_train_every=5)
         >>> metrics_dict = {"train/loss": 5.00}
         >>> logger.log(metrics_dict)
-        [0 epochs]: TRAIN:[loss=5.000]
+        [-1 epochs]: TRAIN:[loss=5.000]
```
henryre: Also weird...accepting to unblock but let's ask ourselves if this even needs an example in the docstring

henryre:
@bhancock8 see first review...why wasn't this caught by Travis and `tox` locally?

PR #1306, opened 2019-07-22
[EASY] Change `k` to `cardinality`

paroma:
See title

__Test Plan__
Tests pass

### DIFF
--- a/test/labeling/test_convergence.py
+++ b/test/labeling/test_convergence.py
@@ -90,7 +90,7 @@ class LabelingConvergenceTest(unittest.TestCase):
         self.assertEqual(L_train.shape, (self.N_TRAIN, len(labeling_functions)))
 
         # Train LabelModel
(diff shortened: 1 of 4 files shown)


test/labeling/test_convergence.py
```diff
@@ -90,7 +90,7 @@ def test_labeling_convergence(self) -> None:
         self.assertEqual(L_train.shape, (self.N_TRAIN, len(labeling_functions)))
 
         # Train LabelModel
-        label_model = LabelModel(cardinality=self.k, verbose=False)
+        label_model = LabelModel(cardinality=self.cardinality, verbose=False)
```
brahmaneya: There is a cls.k above that you will also have to change

brahmaneya: Actually, it is strange that the test passed despite this.

paroma: it didn't, Travis would have caught it. fixing

PR #1307, opened 2019-07-23
[EASY] Probs to preds

bhancock8:
Catch a simple error mode where users pass a 1D vector into probs_to_preds and get all 1s.

**Test plan:**
Add unit test confirming error is raised.

### DIFF
--- a/snorkel/analysis/utils.py
+++ b/snorkel/analysis/utils.py
@@ -54,14 +54,17 @@ def probs_to_preds(
     >>> probs_to_preds(np.array([[0.8, 0.1, 0.1]]))
     array([1])
     """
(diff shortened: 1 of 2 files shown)


snorkel/analysis/utils.py
```diff
@@ -55,6 +55,9 @@ def probs_to_preds(
     array([1])
     """
     n, k = probs.shape
+    if k <= 1:
+        raise ValueError(f"probs must have at least 2 columns. Instead, got {k}.")
```
paroma: nit: can we change to mirror columns = cardinality here somehow?

bhancock8:
Ready for another look. @paroma @brahmaneya

PR #1308, opened 2019-07-23
Remove synthetic tutorial

henryre:
Will have to go eventually (probably won't be ported), and synthetic generators weren't used elsewhere.

**Test plan**
tox passes locally
No coverage drop (manual since codecov is down)


snorkel/synthetic/synthetic_data.py
```diff
@@ -1,14 +1,8 @@
...
 import numpy as np
-import pandas as pd
 from scipy.sparse import csr_matrix, lil_matrix
 
-from snorkel.augmentation.tf import LambdaTransformationFunction
-from snorkel.labeling.lf import LabelingFunction
-from snorkel.types import DataPoint
-
 
 def generate_simple_label_matrix(
```
vincentschen: where is this used now?

paroma: in label model tests

PR #1309, opened 2019-07-23
Change label conventions

henryre:
LFs return `None` for abstain, abstains represented as `None` in label matrix, label matrix is dense outside of compute modules. This covers a lot of changes. Others:

* Add LFAnalysis with all LF analysis tools as methods. Prevents constant re-cast to sparse. GitHub tracked this change poorly, but it's mostly an indent and adding a bunch of `self.`s
* Internal computation for LF analysis and label model is with shifted sparse matrix: `csr_matrix(L + 1)`
* Remove confusion matrix (redundant with sklearn)
* Remove label conversion utilities
* Remove unused synthetic test utilities

**Test plan**
Locally: `tox -e complex,spark`, `tox`


snorkel/analysis/error_analysis.py
```diff
@@ -1,14 +1,14 @@
 from collections import defaultdict
-from typing import Any, Counter, Iterable, List, Mapping, Optional, Sequence, Tuple
+from typing import DefaultDict, Dict, List, Tuple
 
 import numpy as np
 
 from .utils import to_int_label_array
 
+Coord = Tuple[int, int]
```
bhancock8: nit: Don't see a reason to abbreviate `Coordinate`; we only use it twice.

snorkel/analysis/error_analysis.py
```diff
@@ -1,14 +1,14 @@
...
 
-def error_buckets(
-    golds: np.ndarray, preds: np.ndarray, X: Optional[Sequence[Any]] = None
-) -> Mapping[Tuple[int, int], Any]:
+
+def error_buckets(golds: np.ndarray, preds: np.ndarray) -> Dict[Coord, List[int]]:
     """Return examples (or their indices) bucketed by gold label/pred label combination.
```
bhancock8: Looks like you removed the functionality to return the examples; now it only returns indices. That's fine, and makes for clearer type checking, but let's update the docstring.

snorkel/analysis/utils.py
```diff
@@ -20,12 +20,12 @@ def _hash(i: int) -> int:
 
 
 def probs_to_preds(
-    probs: np.ndarray, tie_break_policy: Optional[str] = "random", tol: float = 1e-5
+    probs: np.ndarray, tie_break_policy: str = "random", tol: float = 1e-5
```
bhancock8: Good catch

snorkel/labeling/analysis.py
```diff
@@ -4,399 +4,349 @@
...
-    >>> label_overlap(L)
-    0.6
-    """
-    return _overlapped_data_points(L).sum() / L.shape[0]
-
-
-def label_conflict(L: sparse.spmatrix) -> float:
-    """Compute the fraction of data points with conflicting (non-abstain) labels.
+class LFAnalysis:
+    """Run analyses on LFs using label matrix.
```
bhancock8: Add Attributes

bhancock8: So some of the methods use L (ABSTAIN = 0) and some use _L_sparse (ABSTAIN = -1). Is that just for convenience, or are there some of these helpers that absolutely require the original L in order to be calculated?

henryre: Some only operate over NumPy arrays

henryre:
@bhancock8 yeah I was torn on how to return abstains. It’s nice syntactic sugar to just drop the returns. But mypy complains (so I added it back to tests) and that’s likely because explicit > implicit. And a Union return type is unnecessary complexity (which actually ends up being difficult to pickle). And lack of clarity in the None to -1 conversion. So I agree, let’s just do -1.

HiromuHota:
If you don't mind, please let us know the reason for changing the label conventions.
I'd like Fonduer to adapt to this change, but the Fonduer team is wondering the reason behind it.
(cc @SenWu and @lukehsiao)

henryre:
Hi @HiromuHota, great question! tl;dr this change reduces user friction significantly since people no longer have to the Snorkel label space to their classifier's label space.

Using 0 for ABSTAIN made it easy for Snorkel to internally leverage sparse matrices. However, it forced users to do a lot of massaging to align the Snorkel class label space (which is 1-indexed) with their classifier label space (which is 0-indexed). We prioritized user-friendliness and removed sparse matrix usage since our models would do an internal cast to dense anyways (and for the vast majority of use cases, this is reasonable memory-wise). We plan to add back internal usage of sparse matrices in a later version.

HiromuHota:
Thank you so much @henryre for your quick reply!
I also think user-friendliness is important for wider adoption of Snorkel.

PR #1310, opened 2019-07-23
Cleanup checkpointer + fix checkpoint_metric silent failure from malformed naming

vincentschen:
1. BugFix: Silent checkpointing failure
https://github.com/HazyResearch/snorkel/blob/redux/snorkel/classification/training/loggers/checkpointer.py#L138-L139 is never triggered for checkpointing.

When the `checkpointer_config` is specified in the form `metric:mode`, checkpointing silently fails:
```
trainer_config = {
    "progress_bar": True,
    "n_epochs": 10,
    "optimizer_config":{"lr": 1e-5},
    "checkpointing": True,
    "checkpointer_config": {
        "checkpoint_dir": "checkpoints",
        "checkpoint_metric": "spam_task/spam_dataset/valid/accuracy:max"
    }
}

trainer = Trainer(**trainer_config)
```

* `checkpoint_task_metrics` is populated with `checkpoint_metric`, which is usually specified as `metric:mode` in the tests
* however, `checkpoint_metric` never passed as `metric:mode` in end to end training, because it always end up being passed into the checkpointer as `task/dataset/split/metric:mode`

so, it silently fails to checkpoint if `checkpoint_metric` is of the form `metric:mode`, instead of `task/dataset/split/metric:mode`

2. Cleanup: Decouple `checkpoint_dir` from `log_writer.log_dir`—they are set completely independently for now.
3. Cleanup: Add `checkpoint_runway` validation (cannot be < 0) 

**Test Plan**
* Added additional test under `test_bad_metric_name`
* Fixed previous tests with correctly formatted metrics
* Add additional unit tests for (2) / (3)
* Pass with `tox -e complex` and full `tox`

### DIFF
--- a/snorkel/classification/training/loggers/checkpointer.py
+++ b/snorkel/classification/training/loggers/checkpointer.py
@@ -62,6 +62,7 @@ class Checkpointer:
         self, counter_unit: str, evaluation_freq: float, **kwargs: Any
     ) -> None:
         self.config = CheckpointerConfig(**kwargs)
+        self._validate_config()
 
         # Pull out checkpoint settings
         self.checkpoint_unit = counter_unit
@@ -70,9 +71,6 @@ class Checkpointer:
         self.checkpoint_runway = self.config.checkpoint_runway
         self.checkpoint_factor = self.config.checkpoint_factor
         self.checkpoint_condition_met = False
-
-        if self.checkpoint_dir is None:
-            raise ValueError("Checkpointing is on but no checkpoint_dir was specified.")
 
         # Collect all metrics to checkpoint
         self.checkpoint_metric = self._make_metric_map([self.config.checkpoint_metric])
@@ -198,6 +196,16 @@ class Checkpointer:
             model.load(best_model_path)
 
         return model
+
+    def _validate_config(self) -> None:
+        if self.config.checkpoint_dir is None:
+            raise ValueError("Checkpointing is on but no checkpoint_dir was specified.")
+
+        split_checkpoint_metric = self.config.checkpoint_metric.split("/")
+        if len(split_checkpoint_metric) != 4:
+            raise ValueError(
+                "checkpoint_metric must be formatted 'task/dataset/split/metric:mode'."
+            )
 
     def _make_metric_map(
         self, metric_mode_iter: Optional[Iterable[str]]
--- a/snorkel/classification/training/loggers/checkpointer.py
+++ b/snorkel/classification/training/loggers/checkpointer.py
@@ -199,9 +199,6 @@ class Checkpointer:
         return model
 
     def _validate_config(self) -> None:
-        if self.config.checkpoint_dir is None:
-            raise ValueError("Checkpointing is on but no checkpoint_dir was specified.")
-
         split_checkpoint_metric = self.config.checkpoint_metric.split("/")
         if len(split_checkpoint_metric) != 4:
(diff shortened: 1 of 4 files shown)


snorkel/classification/training/loggers/checkpointer.py
```diff
@@ -90,18 +88,19 @@ def __init__(
         if self.checkpoint_freq <= 0:
             raise ValueError(
                 f"Invalid checkpoint freq {self.checkpoint_freq}, "
-                f"must be greater 0."
+                f"must be greater than 0."
```
henryre: 🕵️‍♀️

snorkel/classification/training/loggers/checkpointer.py
```diff
@@ -90,18 +88,19 @@ def __init__(
...
-            f"Save checkpoints at {self.checkpoint_dir} every "
+            f"Save checkpoints at '{self.checkpoint_dir}' every "
             f"{self.checkpoint_freq} {self.checkpoint_unit}."
         )
 
-        logging.info(
-            f"No checkpoints will be saved before {self.checkpoint_runway} "
-            f"{self.checkpoint_unit}."
-        )
+        if self.checkpoint_runway > 0:
```
henryre: nice catch

snorkel/classification/training/loggers/checkpointer.py
```diff
@@ -199,6 +198,22 @@ def load_best_model(self, model: SnorkelClassifier) -> SnorkelClassifier:
 
         return model
 
+    def _validate_config(self) -> None:
+        if self.config.checkpoint_dir is None:
```
henryre: `None` isn't a valid input anymore, so probably don't need this. Shouldn't need to throw exceptions for every contract violation.

snorkel/classification/training/loggers/checkpointer.py
```diff
@@ -199,6 +198,22 @@ def load_best_model(self, model: SnorkelClassifier) -> SnorkelClassifier:
...
+        split_checkpoint_metric = self.config.checkpoint_metric.split("/")
+        if len(split_checkpoint_metric) != 4:
+            raise ValueError(
+                "checkpoint_metric must be formatted 'task/dataset/split/metric:mode'."
+            )
+
+        if self.config.checkpoint_runway < 0:
+            raise ValueError(
+                f"Invalid checkpoint_runway {self.config.checkpoint_runway}, "
+                f"must be greater than 0."
```
henryre: geq

test/classification/training/loggers/test_checkpointer.py
```diff
@@ -21,57 +21,65 @@ def test_checkpointer(self) -> None:
             **log_manager_config,
             checkpoint_dir=self.test_dir,
             checkpoint_runway=3,
-            checkpoint_metric="f1:max",
+            checkpoint_metric="task/dataset/valid/f1:max",
```
henryre: Ok so what's going on with these? Not actually a valid metric specification? And with the updated tests, we'd catch this now?

vincentschen: right—`f1:max` is valid if passed to the checkpointer directly.
however, in any real application, `checkpoint_metric` is passed through the `trainer`, which formats the metric name with the applicable task/dataset information. 
that's where the silent bug was happening.

PR #1311, opened 2019-07-23
[EASY] Bound all package versions

henryre:
The pandas 0.25 breakage made it clear that we should bound our entire scientific stack. This folds in https://github.com/HazyResearch/snorkel/pull/1271. Upgrades a few packages, no breaks. Run `tox -r` to recreate your test environment and ensure compliance.

**Test plan**
`tox -r` and run complex tests locally


henryre:
Good point re: deployability. Hm I was going off SpaCy, which does (https://github.com/explosion/spaCy/blob/master/setup.py#L226). We could be more reactive I guess? Or maybe just for install_requires packages? But tools should be bounded, want all evaluations to be consistent.

PR #1314, opened 2019-07-23
Update LabelModel config

bhancock8:
- Replace config dict with Config namedtuples (`LabelModelConfig` and `TrainConfig`)
- Delete recursive_merge_dict 🎉 
- Refactor train_model() -> fit(); log_train_every -> log_freq
- Move optimizers, lr_schedulers, and `merge_config()` to utils/ (since they're used by both Classifier and LabelModel)
- Added docstrings describing each attribute in the configs
- Move lr/l2 directly accessible in `TrainConfig`/`TrainerConfig` instead of nesting in `OptimizerConfig`

**Test plan:**
- Updated tests to use the new signatures
- ✔ `tox`


snorkel/classification/utils.py
```diff
@@ -108,66 +107,6 @@ def merge_config(config: Config, config_updates: Dict[str, Any]) -> Config:
     return config._replace(**config_updates)
 
 
-def recursive_merge_dicts(
```
henryre: 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏 👏

snorkel/utils/optimizers.py
```diff
@@ -0,0 +1,34 @@
...
+    """Settings for Adamax optimizer."""
+
+    betas: Tuple[float, float] = (0.9, 0.999)
+    eps: float = 1e-8
+
+
+class OptimizerConfig(Config):
+    """Settings common to all optimizers."""
+
+    lr: float = 0.001
```
henryre: Hmmmm this is a pretty critical value. Do we believe this is better than 0.01 for the label model? cc @paroma

henryre: Can we surface lr and l2 one level higher? Just since they're particularly important

test/labeling/test_convergence.py
```diff
@@ -82,7 +82,7 @@ def test_labeling_convergence(self) -> None:
 
         # Train LabelModel
         label_model = LabelModel(cardinality=self.cardinality, verbose=False)
-        label_model.train_model(L_train, lr=0.01, l2=0.0, n_epochs=100)
+        label_model.fit(L_train, lr=0.01, l2=0.0, n_epochs=100)
```
henryre: wait why can we do lr at the top level here?

snorkel/utils/optimizers.py
paroma: Default `n_epochs` changed to account for lower LR. 0.01 is better for LabelModel (as in, it works most of the time with the different matrices we've had across tests and tutorials). I don't have a strong opinion, we can explicitly change LR in the label model if needed

snorkel/labeling/model/label_model.py
```diff
@@ -9,17 +9,69 @@
...
+
+    Parameters
+    ----------
+    n_epochs
+        The number of epochs to train (where each epoch is a single optimization step)
+    l2
+        Centered L2 regularization strength (int, float, or np.array)
+    optimizer
+        Which optimizer to use (one of ["sgd", "adam", "adamax"])
+    optimizer_config
```
paroma: is lr in the config for sgd? maybe a pointer for that

snorkel/labeling/model/label_model.py
```diff
@@ -416,7 +460,7 @@ def score(
         -------
         >>> L = np.array([[1, 1, -1], [0, 0, -1], [1, 1, -1]])
         >>> label_model = LabelModel(verbose=False)
-        >>> label_model.train_model(L)
+        >>> label_model.fit(L, seed=123)
```
paroma: is the seed necessary?

snorkel/labeling/model/label_model.py
```diff
@@ -376,7 +420,7 @@ def predict(
         -------
         >>> L = np.array([[0, 0, -1], [1, 1, -1], [0, 0, -1]])
         >>> label_model = LabelModel(verbose=False)
-        >>> label_model.train_model(L)
+        >>> label_model.fit(L, seed=123)
```
paroma: same as below

test/labeling/model/test_label_model.py
```diff
@@ -242,45 +244,97 @@ def test_model_loss(self):
...
 
-        label_model.train_model(L, n_epochs=10, lr=0.01, momentum=0.9)
+        label_model.fit(L, n_epochs=10)
         next_loss = label_model._loss_mu().item()
 
         self.assertLessEqual(next_loss, init_loss)
 
         with self.assertRaisesRegex(Exception, "Loss is NaN."):
-            label_model.train_model(L, n_epochs=10, lr=1e8)
+            label_model.fit(L, n_epochs=10, optimizer_config={"lr": 1e8})
```
paroma: we can initialize LabelModel again here so the lr doesn't have to passed in like this (we don't want to encourage this usage right?)

test/labeling/model/test_label_model.py
```diff
@@ -299,7 +353,7 @@ def test_label_model(self) -> None:
 
         # Train LabelModel
         label_model = LabelModel(cardinality=self.cardinality, verbose=False)
-        label_model.train_model(L, lr=0.01, l2=0.0, n_epochs=100)
+        label_model.fit(L, n_epochs=200, optimizer_config={"lr": 0.01}, seed=123)
```
paroma: same as above

tutorials/workshop/Workshop_3_Generative_Model_Training.ipynb
```diff
@@ -122,12 +122,12 @@
     "from snorkel.analysis.utils import convert_labels\n",
     "\n",
     "label_model = LabelModel(k=2, verbose=True, seed=123)\n",
-    "label_model.train_model(\n",
+    "label_model.fit(\n",
```
paroma: Line above also need `cardinality=2` instead of `k=2`

snorkel/labeling/model/label_model.py
bhancock8: Not anymore. The original lr (0.01) and n_epochs (100) were restored, and we seem to get consistent convergence for these examples.

bhancock8:
@paroma lr is now accessible top-level, which addresses most of your comments. Other suggested changes were made as well.

PR #1320, opened 2019-07-24
Fix doctest after default lr change

paroma:
* Change doctest example to have no all abstain LF
* Change doctest example to pass in LR and avoid randomness in output

__Test Plan__
Existing tests pass


henryre:
Why does this fix the non-determinism?

paroma:
Closed in favor of #1327

PR #1323, opened 2019-07-24
Support eval on slices

vincentschen:
*Changes*:
* Update slicing naming convention from `{task}_slice:{slice_name}` --> `{task}:slice_{slice_name}`
* Add support for `active` label filtering in `SnorkelClassifier` predict/scoring 
* Update slicing utils to better support baseline labelsets

*Test Plan*:
* Add tests for `task:other_label` label convention
* Add test for filtered predictions
* Ran `tox` and `tox -e complex`


snorkel/classification/snorkel_classifier.py
```diff
@@ -328,10 +322,23 @@ def predict(
...
-            prob_batch_dict = self._calculate_probs(X_batch_dict, Y_batch_dict.keys())
-            for task_name, Y in Y_batch_dict.items():
-                prob_dict_list[task_name].extend(prob_batch_dict[task_name])
-                gold_dict_list[task_name].extend(Y.cpu().numpy())
+            labels_to_tasks = self._get_labels_to_tasks(Y_batch_dict.keys())
+            prob_batch_dict = self._calculate_probs(
+                X_batch_dict, labels_to_tasks.values()
+            )
+            for label_name, Y in Y_batch_dict.items():
+                task_name = labels_to_tasks[label_name]
```
henryre: ok so the idea here is to extract the task_name from the new format so we know how which head output to use, but then report the metric using the full metric name (with slice)?

test/classification/test_snorkel_classifier.py
```diff
@@ -83,6 +83,44 @@ def test_predict(self):
...
+            X_dict={"data": X},
+            Y_dict={"task:other_label": Y},
+        )
+        dataloader = DictDataLoader(dataset, batch_size=BATCH_SIZE)
+
+        results = model.predict(dataloader)
+        self.assertEqual(sorted(list(results.keys())), ["golds", "probs"])
+        self.assertIn("task:other_label", results["golds"])
+        self.assertIn("task:other_label", results["probs"])
+        self.assertNotIn("task", results["golds"])
```
henryre: Great!

test/classification/test_snorkel_classifier.py
```diff
@@ -83,6 +83,44 @@ def test_predict(self):
...
+            results["probs"]["task:other_label"], np.ones((NUM_EXAMPLES, 2)) * 0.5
+        )
+
+        # Now, test malformed version
+        Y = torch.ones(NUM_EXAMPLES, 1).long()
+        dataset = DictDataset(
+            name="dataset",
+            split="train",
+            X_dict={"data": X},
+            Y_dict={"task:other_label:bad!": Y},
```
henryre: I'm lazy...do we already have a test for a bad name without `:`?

snorkel/slicing/utils.py
```diff
@@ -16,6 +16,8 @@ def add_slice_labels(
     base_task: Task,
     slice_labels: np.ndarray,
     slice_names: List[str],
+    include_base: bool = True,
```
henryre: Why do we need this?

snorkel/classification/snorkel_classifier.py
```diff
@@ -289,7 +283,7 @@ def _calculate_probs(
 
         Returns
         -------
-        Dict[str, Iterable[torch.Tensor]]
+        Dict
```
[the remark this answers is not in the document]
vincentschen: I thought we decided not to do types in docstrings?

snorkel/classification/snorkel_classifier.py
vincentschen: exactly! to de-couple the notion of label_name and task_name a bit; because we had the assumption that they always mapped 1:1, but now there might be cases where many label_names map to one task_name

snorkel/slicing/utils.py
vincentschen: in our formulation, we treat the original task as a slice (trivial slice: all examples belong to the slice) so that we can combine representations from the original task as well. 

as a result, we need to add the base task's labels here as well

snorkel/classification/snorkel_classifier.py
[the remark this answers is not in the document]
henryre: For return we do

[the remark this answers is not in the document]
henryre: https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_numpy.html

snorkel/classification/snorkel_classifier.py
henryre: Now that I'm thinking about this, could we just supply the metadata to the dataloader separately, then alter the string representation for the printout?

snorkel/classification/snorkel_classifier.py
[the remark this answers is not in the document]
vincentschen: got it! will keep it in

snorkel/classification/snorkel_classifier.py
vincentschen: what kind of metadata? the minimum amount of information we provide is the slice labels (NxK masks for N examples and K slices)

snorkel/slicing/utils.py
henryre: right but it seems like we never have it be false?

snorkel/classification/snorkel_classifier.py
henryre: no i mean we set the label_name as `{base_task}:{metadata}` now to hack in the metadata. can we just pass in `{metadata}` as a separate field to the object (which then constructs the label name to print to screen) and have the label name still be `{base_task}`. that might break something else, idk

snorkel/classification/snorkel_classifier.py
```diff
@@ -328,10 +322,23 @@ def predict(
...
+                X_batch_dict, labels_to_tasks.values()
+            )
+            for label_name, Y in Y_batch_dict.items():
+                task_name = labels_to_tasks[label_name]
+
+                # Filter out inactive probs and golds
+                active = get_active_mask(Y)
+
+                # Gather the output from the appropriate task
+                probs = prob_batch_dict[task_name][active].cpu().numpy()
```
bhancock8: The issue I see here (and I believe the reason why active filtering in this step was removed) is that the user loses track of which examples correspond to which predictions/golds. If I have metadata corresponding to my 50 data points, then I run predict and get back 30 preds/golds because 20 were filtered, how do I map them to the rest of my data? One option is to tell the user to pass through UIDs in their X_dict. But alternatively, could we have `predict()` just return ABSTAIN  labels for those examples and then deal with not including those in our metrics using the filter_dict in metric_score?

bhancock8: Just seems like if I tell a method to make predictions for 50 datapoints, I should get back 50 predictions, even if some of those are "I don't know".

snorkel/classification/snorkel_classifier.py
```diff
@@ -352,10 +359,39 @@ def predict(
...
+
+                labels_to_tasks[label_name] = split_label_name[0]
+        return labels_to_tasks
+
     @torch.no_grad()
     def score(self, dataloaders: List["DictDataLoader"]) -> Dict[str, float]:
         """Calculate scores for the provided DictDataLoaders.
 
+        For label names that do not correspond directly to a task name:
+            If label_name is follows convention: "{base_task}:{other_label}"
```
bhancock8: is follows

snorkel/classification/utils.py
```diff
@@ -72,6 +72,33 @@ def pad_batch(
     return padded_batch, mask_batch
 
 
+def get_active_mask(Y: torch.Tensor) -> torch.Tensor:
+    """Return a mask of the active samples given some label Tensor Y.
```
bhancock8: Let's define 'active' at least once. i.e., it's active if at least one value in its row is not ABSTAIN.

snorkel/classification/snorkel_classifier.py
vincentschen: what setting has this been a problem in?
in general though, happy to make the change to support filtering in the scorer but surface abstain in predictions as well

bhancock8: Jared today was saying that he won't use a system that doesn't have support UIDs, since he's been burned too many times by hidden shuffles that happen and screw up alignment between data and labels. We don't have that here, per se, but it's a similar issue, I think. It would be a confusing error if a person wanted to do something with their predictions and somewhere in the middle of their pipeline their dataset size shrunk.

vincentschen: Okay, yes—it will probably also be easier to implement UIDs (potentially down the line) if this happens in the Scorer.

snorkel/classification/snorkel_classifier.py
vincentschen: at the moment, the abstraction requires that we copy and mask the original base_task labels to create slice labels

when training a slice model, these slice labels are used directly to compute loss, etc:
```
add_slice_labels(
    dl_valid, spam_task, S_valid, slice_names
)
```
when _evaluating_ a non-slice model, these slice_labels can be used to evaluate the primary slice_head:
```
add_slice_labels(
    dl_valid, spam_task, S_valid, slice_names, 
    include_base=False, include_ind=False
)
```

we _could_ store the metadata as a separate argument. in that case, we would not clone the base_task's labels, and we'd need to handle the training labels for the slice model to in a separate case

for speed, I'd suggest that we see if this API pattern makes sense (can share this in the context of tutorial soon)—and iterate on it afterwards

test/classification/test_snorkel_classifier.py
vincentschen: without the `:` is treated as a normal labelset, so it's covered in other test cases

snorkel/classification/snorkel_classifier.py
```diff
@@ -300,7 +294,7 @@ def _calculate_probs(
         outputs = self.forward(X_dict, task_names)
 
         for task_name in task_names:
-            prob_dict[task_name] = self.output_funcs[task_name](outputs).cpu().numpy()
+            prob_dict[task_name] = self.output_funcs[task_name](outputs)
```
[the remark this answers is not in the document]
vincentschen: got it—this was also resulting in a bug: we were constructing the `active` mask as a tensor and attempting to mask the `np` array `probs`, which was silently failing

vincentschen:
Closing for now— addressed in different implementation: https://github.com/HazyResearch/snorkel/pull/1347

PR #1324, opened 2019-07-25
Add f to formatted string. Correct roc_auc formula.

brahmaneya:
Add f to formatted string. Correct roc_auc formula. 

**Test Plan**
Updated test for changed code. tox is passing.


bhancock8:
Assuming the tests pass.

PR #1327, opened 2019-07-25
hot fix non-determinism in tests

paroma:
* Change `predict_proba` doc string to have a an L matrix without all abstain LF
*  Add seed in data generation for `test_accuracies`

__Test Plan__
* `tox` passes multiple times locally

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -365,13 +365,13 @@ class LabelModel(nn.Module):
 
         Example
         -------
-        >>> L = np.array([[0, 0, -1], [1, 1, -1], [0, 0, -1]])
+        >>> L = np.array([[0, 0, 0], [1, 1, 1], [1, -1, 1]])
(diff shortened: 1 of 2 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -365,13 +365,13 @@ def predict_proba(self, L: np.ndarray) -> np.ndarray:
...
         -------
-        >>> L = np.array([[0, 0, -1], [1, 1, -1], [0, 0, -1]])
+        >>> L = np.array([[0, 0, 0], [1, 1, 1], [1, -1, 1]])
         >>> label_model = LabelModel(verbose=False)
         >>> label_model.fit(L)
         >>> np.around(label_model.predict_proba(L), 1)
         array([[1., 0.],
                [0., 1.],
-               [1., 0.]])
+               [0., 1.0]])
```
henryre: it's not 1. ?

PR #1328, opened 2019-07-25
Fix no labels edge case

henryre:
Used to throw an ambiguous ValueError if there are no labels returned by any LFs. This fixes that edge case. Not error-worthy, since it's a valid label matrix (e.g. for a single, very precise LF on a small dev set)

**Test plan**
TDD, added new test case to verify catch

### DIFF
--- a/snorkel/labeling/apply/core.py
+++ b/snorkel/labeling/apply/core.py
@@ -33,9 +33,11 @@ class BaseLFApplier:
         self._lfs = lfs
 
     def _matrix_from_row_data(self, labels: List[RowData]) -> np.ndarray:
-        row, col, data = zip(*chain.from_iterable(labels))
         L = np.zeros((len(labels), len(self._lfs)), dtype=int) - 1
-        L[row, col] = data
+        # NB: this check will short-circuit, so ok for large L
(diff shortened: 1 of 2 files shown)


snorkel/labeling/apply/core.py
```diff
@@ -33,9 +33,11 @@ def __init__(self, lfs: List[LabelingFunction]) -> None:
         self._lfs = lfs
 
     def _matrix_from_row_data(self, labels: List[RowData]) -> np.ndarray:
-        row, col, data = zip(*chain.from_iterable(labels))
         L = np.zeros((len(labels), len(self._lfs)), dtype=int) - 1
-        L[row, col] = data
+        # NB: this check will short-circuit, so ok for large L
+        if any(map(len, labels)) > 0:
```
brahmaneya: any will return True when it finds a non-zero entry, so you don't need the '> 0' at the end.

henryre: hmmmm well that's what was happening anyways because i had a pretty trick bug (which ends up having no impact). i meant to do 

```
any(map(lambda x: len(x) > 0, labels))
```

PR #1329, opened 2019-07-25
Catch no parens on LFs/mappers

henryre:
Common silent failure: no parens on a mapper or LF. This throws a meaningful error if you try to do that. Looked into allowing both parens and non-parens, and it makes the decorator definition unreadable (whereas now it's extremely readable). I also like the parens because they imply to the user that there are more options.

**Test plan**
Added unit tests

### DIFF
--- a/snorkel/map/core.py
+++ b/snorkel/map/core.py
@@ -9,6 +9,8 @@ import pandas as pd
 
 from snorkel.types import DataPoint, FieldMap
 
+MapFunction = Callable[[DataPoint], Optional[DataPoint]]
+
 
 def get_parameters(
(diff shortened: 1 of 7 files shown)


snorkel/map/core.py
```diff
@@ -309,10 +311,14 @@ class lambda_mapper:
     -------
     >>> @lambda_mapper()
     ... def concatenate_text(x):
-    ...     x.article = f"{title} {body}"
+    ...     x.article = f"{x.title} {x.body}"
```
bhancock8: How was doctest passing before this?

henryre: it didn't actually execute the mapper. it does now!

PR #1330, opened 2019-07-25
Add LF names to analysis init

henryre:
Avoids repetitive boilerplate of extracting LF names for analysis.

A couple decision points:
* Didn't add Y to init since you might want to try against a couple label sets (?)
* Did LFs rather than LF names to 1. avoid name extraction boilerplate and 2. play defense if we want to extract other properties.

**Test plan**
Updated LF name unit test

### DIFF
--- a/snorkel/labeling/analysis.py
+++ b/snorkel/labeling/analysis.py
@@ -1,3 +1,4 @@
+from collections import OrderedDict
 from itertools import product
 from typing import List, Optional, Union
 
@@ -19,6 +20,13 @@ class LFAnalysis:
     L
         Label matrix where L_{i,j} is the label given by the jth LF to the ith
         candidate (using -1 for abstain)
+    lfs
+        Labeling functions used to generate ``L``
+
+    Raises
+    ------
+    ValueError
+        If number of LFs and number of LF matrix columns differ
 
     Attributes
     ----------
@@ -31,7 +39,14 @@ class LFAnalysis:
     ) -> None:
         self.L = L
         self._L_sparse = sparse.csr_matrix(L + 1)
-        self._lf_names = None if lfs is None else [lf.name for lf in lfs]
+        self._lf_names = None
+        if lfs is not None:
+            if len(lfs) != self._L_sparse.shape[1]:
+                raise ValueError(
+                    f"Number of LFs ({len(lfs)}) and number of "
+                    f"LF matrix columns ({self._L_sparse.shape[1]}) are different"
+                )
+            self._lf_names = [lf.name for lf in lfs]
 
     def _covered_data_points(self) -> np.ndarray:
         """Get indicator vector z where z_i = 1 if x_i is labeled by at least one LF."""
@@ -314,8 +329,6 @@ class LFAnalysis:
         Y
             [n] or [n, 1] np.ndarray of gold labels. If provided, the empirical accuracy
             for each LF will be calculated.
-        lf_names
-            Name of each LF. If None, indices are used.
         est_accs
             Learned accuracies for each LF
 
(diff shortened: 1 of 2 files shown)


snorkel/labeling/analysis.py
```diff
@@ -24,9 +26,12 @@ class LFAnalysis:
         See above.
     """
 
-    def __init__(self, L: np.ndarray) -> None:
+    def __init__(
+        self, L: np.ndarray, lfs: Optional[List[LabelingFunction]] = None
+    ) -> None:
         self.L = L
         self._L_sparse = sparse.csr_matrix(L + 1)
+        self._lf_names = None if lfs is None else [lf.name for lf in lfs]
```
brahmaneya: Check for L.shape[1] == len(lfs)?

snorkel/labeling/analysis.py
```diff
@@ -323,9 +325,11 @@ def lf_summary(
             Summary statistics for each LF
         """
         n, m = self.L.shape
-        if lf_names is not None:
+        lf_names: Union[List[str], List[int]]
+        if self._lf_names is not None:
             col_names = ["j"]
```
brahmaneya: col_names seems to just be the keys of d. So we can have d be a collection.OrderedDict and just d.keys() in place of col_names.

snorkel/labeling/analysis.py
bhancock8: +1

snorkel/labeling/analysis.py
```diff
@@ -300,10 +305,7 @@ def lf_empirical_probs(self, Y: np.ndarray, k: int) -> np.ndarray:
         return P
 
     def lf_summary(
-        self,
-        Y: Optional[np.ndarray] = None,
-        lf_names: Optional[Union[List[str], List[int]]] = None,
-        est_accs: Optional[np.ndarray] = None,
+        self, Y: Optional[np.ndarray] = None, est_accs: Optional[np.ndarray] = None
```
bhancock8: Remove lf_names from Parameters

snorkel/labeling/analysis.py
paroma: add check for matching LF names? in case people copy pasted the same LF twice/named two LFs the same thing

test/labeling/test_analysis.py
```diff
@@ -105,7 +106,13 @@ def test_lf_summary(self) -> None:
...
+
+        def f(x):
+            return -1
+
+        lfs = [LabelingFunction(s, f) for s in names]
+        lfa = LFAnalysis(np.array(L), lfs)
+        df = lfa.lf_summary(self.Y, est_accs=est_accs)
         df_expected = pd.DataFrame(
             {
                 "j": [0, 1, 2, 3, 4, 5],
```
paroma: Should this be the names of the LFs (abcdef)? If not, is there a test to check that the names in the DataFrame are correct?

snorkel/labeling/analysis.py
henryre: would that be invalid?

test/labeling/test_analysis.py
henryre: yes this tests that the names are correct (see the index set below)

bhancock8:
Ooh, +1 to @paroma's comment about checking for duplicate LF names. I don't see any good reason for LFs to have duplicate names, and this could easily catch the common mistake (that I've made many times) of not updating an LF's name when copying and pasting so I end up adding the same LF twice.

paroma:
Duplicates should be checked somewhere -  in the applier probably. And it’s not technically invalid but a warning message that two repeated names exist would help I think

henryre:
Agree a check somewhere would be good, but let's be more intentional about where that happens. The LF class itself could act as a registry, etc. Lots of better options than here. Leaving for now, will make a note to add.

PR #1337, opened 2019-07-26
Skip Flakey Tests

paroma:
* Set seed for LabelModel docs
* Increase delta for checking equality to account for rounding issues
* Skip flakey tests

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -340,7 +340,7 @@ class LabelModel(nn.Module):
         -------
         >>> L = np.array([[1, 1, 1], [1, 1, -1], [-1, 0, 0], [0, 0, 0]])
         >>> label_model = LabelModel(verbose=False)
-        >>> label_model.fit(L)
+        >>> label_model.fit(L, n_epochs=1000, seed=123)
         >>> np.around(label_model.get_accuracies(), 2)
         array([0.99, 0.99, 0.99])
         """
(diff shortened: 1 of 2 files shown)


henryre:
@paroma what's the status here? we at least need to skip the non-deterministic ones to unblock people (got a tox failure for the [0.99, 0.99, 0.99] doctest or whatever)

snorkel/labeling/model/label_model.py
```diff
@@ -366,9 +366,9 @@ def predict_proba(self, L: np.ndarray) -> np.ndarray:
 
         Example
         -------
-        >>> L = np.array([[0, 0, 0], [1, 1, 1], [1, -1, 1]])
+        >>> L = np.array([[0, 0, 0], [1, 1, 1], [1, 1, 1]])
```
henryre: Probably a worse example without an abstain. In general, this L is just not super realistic.

snorkel/labeling/model/label_model.py
```diff
@@ -340,7 +340,7 @@ def get_accuracies(self) -> np.ndarray:
         -------
         >>> L = np.array([[1, 1, 1], [1, 1, -1], [-1, 0, 0], [0, 0, 0]])
         >>> label_model = LabelModel(verbose=False)
-        >>> label_model.fit(L)
+        >>> label_model.fit(L, n_epochs=1000, seed=123)
```
henryre: Is there any other source of randomness when we set this seed? If not, we can drop the number of epochs and make the Ls more realistic.

paroma:
I commented out the doctests for now, revisiting post-release

PR #1342, opened 2019-07-28
Restore dtypes after pandas data augmentation

brahmaneya:
Restore dtypes after pandas data augmentation.

Test Plan
Added lines to existing test to check dtypes after PandasTFApplier apply(), and apply_generator().

### DIFF
--- a/test/augmentation/apply/test_tf_applier.py
+++ b/test/augmentation/apply/test_tf_applier.py
@@ -31,6 +31,7 @@ def modify_in_place(x: DataPoint) -> DataPoint:
 
 
 DATA = [1, 2, 3]
+STR_DATA = ["x", "y", "z"]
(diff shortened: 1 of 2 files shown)


test/augmentation/apply/test_tf_applier.py
```diff
@@ -158,6 +158,7 @@ def test_tf_applier_pandas(self):
         applier = PandasTFApplier([square], policy)
         df_augmented = applier.apply(df, progress_bar=False)
         df_expected = pd.DataFrame(dict(num=[1, 16, 81]), index=[0, 1, 2])
+        self.assertEqual(df_augmented.num.dtype, "int64")
```
henryre: Did we check that these fail before updating PandasTFApplier?

brahmaneya: Just checked, I needed to add some strings to the df to make it fail. If it it all ints, then the dtype stays as int64.

PR #1344, opened 2019-07-29
Fix is_package indentation bug

AndyLc:
Encountered this bug while playing around with snorkel.

The bug was that we never go through a single iteration of the loop because either True is returned or False is immediately returned. Instead we should be looping through all packages, and if True is never returned, False should be returned.


henryre:
Hi @AndyLc, we've moved the information extraction version of Snorkel (v0.7) to https://github.com/snorkel-team/snorkel-extraction. Feel free to reopen this PR there! Just ping me and I'll accept right away.

PR #1346, opened 2019-07-31
Genericize error_buckets

henryre:
Based on session feedback. 1. remove gold/pred hardcoding 2. make the name more generic (not all errors) and 3. allow n-ary

**Test plan**
Updated unit tests

### DIFF
--- a/snorkel/analysis/error_analysis.py
+++ b/snorkel/analysis/error_analysis.py
@@ -6,33 +6,32 @@ import numpy as np
 from .utils import to_int_label_array
 
 
-def error_buckets(
-    golds: np.ndarray, preds: np.ndarray
-) -> Dict[Tuple[int, int], List[int]]:
-    """Return data point indices bucketed by gold label/pred label combination.
+def label_buckets(*y: np.ndarray) -> Dict[Tuple[int, int], np.ndarray]:
+    """Return data point indices bucketed by label combinations.
 
-    Returned buckets[i,j] is a list of items with predicted label i and true label j.
-    For a binary problem with (1=positive, 0=negative):
-        buckets[1,1] = true positives
-        buckets[1,0] = false positives
-        buckets[0,1] = false negatives
-        buckets[0,0] = true negatives
+    A common use case is calling ``label_buckets(Y_gold, Y_pred)`` where
+    ``Y_gold`` is a set of gold (i.e. ground truth) labels and ``Y_pred`` is a
+    corresponding set of predicted labels.
+    The returned ``buckets[(i, j)]`` is a NumPy array of data point indices with
+    predicted label i and true label j.
+    For a binary problem with (1 = positive, 0 = negative):
+        ``buckets[(1, 1)]`` = true positives
+        ``buckets[(1, 0)]`` = false negatives
+        ``buckets[(0, 1)]`` = false positives
+        ``buckets[(0, 0)]`` = true negatives
 
     Parameters
     ----------
-    golds
-        An np.ndarray of gold (int) labels
-    preds
-        An np.ndarray of (int) predictions
+    *y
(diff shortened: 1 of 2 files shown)


snorkel/analysis/error_analysis.py
```diff
@@ -6,33 +6,32 @@
 from .utils import to_int_label_array
 
 
-def error_buckets(
-    golds: np.ndarray, preds: np.ndarray
-) -> Dict[Tuple[int, int], List[int]]:
-    """Return data point indices bucketed by gold label/pred label combination.
+def label_buckets(*y: np.ndarray) -> Dict[Tuple[int, int], np.ndarray]:
```
brahmaneya: Why remove the return type?

snorkel/analysis/error_analysis.py
```diff
@@ -6,33 +6,32 @@
...
+    """Return data point indices bucketed by label combinations.
 
-    Returned buckets[i,j] is a list of items with predicted label i and true label j.
-    For a binary problem with (1=positive, 0=negative):
-        buckets[1,1] = true positives
-        buckets[1,0] = false positives
-        buckets[0,1] = false negatives
-        buckets[0,0] = true negatives
+    A common use case is calling ``label_buckets(Y_gold, Y_pred)`` where
```
bhancock8: `buckets = label_buckets...` so that when we refer to buckets[(i,j)] it's clear what that is.

snorkel/analysis/error_analysis.py
```diff
@@ -6,33 +6,32 @@
...
+    *y
+        A list of np.ndarray of (int) labels
 
     Returns
     -------
     Dict
-        A mapping of each error bucket to its corresponding indices
+        A mapping of each label bucket to a NumPy array of its corresponding indices
     """
     buckets: DefaultDict[Tuple[int, int], List[int]] = defaultdict(list)
```
brahmaneya: Tuple[int, int] assumes you have exactly two input arguments.

bhancock8: Add quick error check that all arrays are of equal length

snorkel/analysis/error_analysis.py
bhancock8: `label buckets` can be read as `label the buckets`, which we want to avoid, especially with a lot of classes in this library that do actually label stuff and have label in their name. How about `get_label_buckets` or `get_label_groups`?

test/analysis/test_error_analysis.py
```diff
@@ -2,20 +2,33 @@
...
-from snorkel.analysis.error_analysis import error_buckets
+from snorkel.analysis.error_analysis import label_buckets
 
 
 class ErrorAnalysisTest(unittest.TestCase):
-    def test_error_buckets(self) -> None:
-        golds = np.array([1, 2, 3, 1, 2, 3])
-        preds = np.array([[2], [1], [3], [1], [1], [3]])
-        buckets = error_buckets(golds, preds)
+    def test_label_buckets(self) -> None:
```
brahmaneya: Since we now support more than two arguments, there should be at least one test case for that.

brahmaneya:
Meta-comment: This is going to break the tutorials. If we have other similar API changes incoming, perhaps we should batch them (at least submit them all on the same day), so we can fix the tutorials in one go?

snorkel/analysis/error_analysis.py
henryre: it's still there...

test/analysis/test_error_analysis.py
henryre: the third one does that

henryre:
@brahmaneya that sounds good (for API-breaking changes) since people are using the tutorials now

PR #1351, opened 2019-08-05
Move snorkel.labeling.preprocessor to snorkel.preprocessor

vincentschen:
**Key Changes**
* Move `snorkel.labeling.preprocess` -> `snorkel.preprocess`
* Move `LabelingFunction`'s `preprocessor` arg -> `pre`

NOTE: needs associated changes in `snorkel-tutorials` (WIP)

**Test plan**
* Ran tox


snorkel/preprocess/core.py
```diff
@@ -6,7 +6,7 @@
 pipeline. This allows LFs to share common preprocessing steps that
 make it easier to express labeling logic. A simple example for text
 processing is concatenating the title and body of an article. For a
-more complex example, see ``snorkel.labeling.preprocess.nlp.SpacyPreprocessor``.
+more complex example, see ``snorkel.preprocess.nlp.SpacyPreprocessor``.
```
henryre: this docstring contains specific text about LFs. change to "Snorkel operators"

henryre: and do a general pass for this

snorkel/labeling/lf/core.py
```diff
@@ -1,6 +1,6 @@
 from typing import Any, Callable, List, Mapping, Optional
 
-from snorkel.labeling.preprocess import BasePreprocessor
+from snorkel.preprocess import BasePreprocessor
```
henryre: should we just change move all of this to `snorkel.map` instead of `snorkel.preprocess` and remove the indirection?

brahmaneya:
While we are doing this, should we also rename pre to preprocessor here? https://github.com/HazyResearch/snorkel/blob/redux/snorkel/map/core.py#L93

@henryre @bhancock8 @vincentschen

snorkel/labeling/lf/core.py
vincentschen: `map` is pretty vague—we could merge `map` into `snorkel.preprocess`

henryre:
@brahmaneya I'd vote to move all of this to `snorkel.map` then change the `preprocessor` arg of LF to pre so there's less indirection. there's one subtle difference which is that general mappers can return `None` (e.g. TFs) but preprocessors cannot. any opinions on how to handle that? cc @vincentschen

brahmaneya:
> @brahmaneya I'd vote to move all of this to `snorkel.map` then change the `preprocessor` arg of LF to pre so there's less indirection. there's one subtle difference which is that general mappers can return `None` (e.g. TFs) but preprocessors cannot. any opinions on how to handle that? cc @vincentschen

+1 to moving to map and renaming to pre. W.r.t. returning None, currently the code checks for None and throws ValueError for None outputs (https://github.com/HazyResearch/snorkel/blob/redux/snorkel/labeling/lf/core.py#L64) . We could keep it that way, or just re-use the original x when pre(x) is None.

vincentschen:
> > @brahmaneya I'd vote to move all of this to `snorkel.map` then change the `preprocessor` arg of LF to pre so there's less indirection. there's one subtle difference which is that general mappers can return `None` (e.g. TFs) but preprocessors cannot. any opinions on how to handle that? cc @vincentschen
> 
> +1 to moving to map and renaming to pre. W.r.t. returning None, currently the code checks for None and throws ValueError for None outputs (https://github.com/HazyResearch/snorkel/blob/redux/snorkel/labeling/lf/core.py#L64) . We could keep it that way, or just re-use the original x when pre(x) is None.

I personally like the naming of `preprocess`.. very obvious to folks what utilities we provide help with. Map seems much vaguer...

vincentschen:
@brahmaneya I'm also okay with calling it `pre` in the Mapper, because the notion of a preprocessor is technically distinct. It makes sense to me that `Mapper.pre` refers to anything that happens _before_ the mapper itself, whereas a `Preprocessor` inherits from `Mapepr`

snorkel/preprocess/core.py
vincentschen: thanks for the catch!

PR #1353, opened 2019-08-06
Fix doc issues

henryre:
Fixes several documentation issues:

* No longer required to run sphinx-apidoc (or tox -e doc) before committing, automatically run in RTD now
* Add source links
* Remove members of nn.Module from child classes
* Fix some issues with formatting ("See ..." and "backtickbacktick s")

**Test plan**
* Local tox -e doc checks
* Will verify RTD build post-hoc


henryre:
* it's just escaping "s" since "backtickbacktick s" is some special thing that renders incorrectly. 
* they were blank pages/useless

docs/conf.py
```diff
@@ -13,6 +13,8 @@
 import os
 import sys
 
+import torch
```
vincentschen: [Q] why are we importing this?

henryre: see below, we're reading the members of nn.Module to exclude them from child classes

PR #1358, opened 2019-08-07
Refactor import paths

henryre:
Our import paths are too long. All user-facing modules/functions are now accessible via `from snorkel.XXX import MyThing`. The exception is when non-essential packages (PySpark, SpaCy, etc.) are required. Most intermediate __init__.py imports are removed (e.g. `snorkel/labeling/apply/__init__.py` imports are removed in favor of snorkel/labeling/__init__.py` imports) for deduplication and consistent API access.

**Test plan**
* Updated unit tests
* Ran complex and spark tests


snorkel/classification/training/trainer.py
```diff
@@ -12,7 +12,12 @@
     ClassifierConfig,
     SnorkelClassifier,
 )
-from snorkel.classification.training import (
+from snorkel.types import Config
+from snorkel.utils.config_utils import merge_config
+from snorkel.utils.lr_schedulers import LRSchedulerConfig
+from snorkel.utils.optimizers import OptimizerConfig
```
brahmaneya: Are these no user facing? (since they are three levels deep).

henryre: correct, they are not

test/labeling/apply/test_lf_applier.py
```diff
@@ -7,9 +7,8 @@
 import pytest
 from dask import dataframe as dd
 
-from snorkel.labeling.apply import LFApplier, PandasLFApplier
+from snorkel.labeling import LFApplier, PandasLFApplier, labeling_function
 from snorkel.labeling.apply.dask import DaskLFApplier, PandasParallelLFApplier
```
brahmaneya: Dask stuff should be two levels deep right?

henryre: no, it's not an essential library

henryre: https://github.com/HazyResearch/snorkel/blob/redux/requirements.txt#L9

PR #1359, opened 2019-08-07
RTD improvements

henryre:
* Switch to autosummary from apidoc
  * Makes documentation much more legible
  * New page for every member, so more indexable
  * Requires creation of package page source files, handled by sync_api_docs.py
  * This has to be run manually and committed, however it is done by `tox -e fix` and checked by `tox -e check`
  * Currently documents everything importable via `from snorkel.XXX import YYY`, plus additional ones specified in packages.json (see #1358 for reasoning)
* Update Sphinx packages, simplify mocking
* Add logo to RTD, fix some paths, dynamic copyright
* Moves tbX as essential package, removes torchtext (unused)

Here's an initial build, ~~but there are issues with mocked packages~~: https://snorkel.readthedocs.io/en/rtd-nice/. compare to https://snorkel.readthedocs.io/en/redux/

~~Contacting RTD folks now to see if we can get them installed correctly.~~

**Test plan**
* Verify check functionality of sync_api_docs manually
* Verify tox commands manually
* Testing RTD compat as we go


henryre:
@bhancock8 tbX move and classification errors are the same. tbx is on the import path for Trainer, so it's an essential package (core functionality fails if it isn't installed). Is there any way we can change it so that it isn't always imported with trainer if a different logger is used?

henryre:
Fixed with a Sphinx update!

henryre:
https://snorkel.readthedocs.io/en/rtd-nice/index.html

henryre:
@bhancock8 it’s

docs/index.rst
```diff
@@ -10,21 +10,18 @@ If you're looking for technical details on Snorkel's API,
 you're in the right place.
 
 For more narrative walkthroughs of Snorkel fundamentals or
-example use cases, check out our `homepage <http://snorkel.stanford.edu/>`_
+example use cases, check out our `homepage <http://snorkel.org/>`_
```
vincentschen: 🎉

scripts/sync_api_docs.py
```diff
@@ -0,0 +1,91 @@
...
+    for name in dir(package):
+        if name.startswith("_"):
+            continue
+        obj = getattr(package, name)
+        if isinstance(obj, type) or callable(obj):
+            members.append(name)
+    return members
+
+
+def main(check: bool) -> None:
```
vincentschen: some quick high-level comments about what this file is executing?

vincentschen: code is mostly self-explaining; this is just for folks who want to quickly get a sense

henryre:
@vincentschen want to land to get everyone synced, but will address all of these in a follow-up

bhancock8:
@henry the better option for TensorboardX is to just use PyTorch's native tensorboard support (as of 1.1, I think?). Also on the Asana tracker for v0.10 atm.

PR #1363, opened 2019-08-07
Address doc comments

henryre:
As requested in #1359


docs/packages/utils.rst
```diff
@@ -1,7 +1,7 @@
 Snorkel Utils Package
 ---------------------
 
-None
+General modeling utilities shared across Snorkel.
```
vincentschen: I would be hesitant to bucket them purely under modeling— even broader `ML utilities` could be safer

PR #1366, opened 2019-08-08
Add soft cross entropy loss

bhancock8:
1. Add cross_entropy_soft() to mirror F.cross_entropy() (which only supports integer labels)
2. Rename the current helpers ce_loss() and softmax() to cross_entropy_from_outputs() and softmax_from_outputs() to emphasize that these methods are wrappers that pull one tensor from the outputs dict and operate over that
3. Add tests (adapted from MeTaL) confirming that soft cross entropy loss behaves as expected.

**Test plan:**
New tests for soft cross entropy in `test_loss.py`
Corresponding PR in `snorkel-tutorials` updates class names and imports there


snorkel/classification/loss.py
```diff
@@ -0,0 +1,122 @@
...
+
+    Raises
+    ------
+    ValueError
+        If an invalid reduction keyword is submitted
+    """
+    num_points, num_classes = input.shape
+    # Note that t.new_zeros, t.new_full put tensor on same device as t
+    cum_losses = input.new_zeros(num_points)
+    for y in range(num_classes):
```
henryre: Why not just vectorize? a la https://github.com/tensorflow/tensorflow/blob/r1.14/tensorflow/python/keras/backend.py#L4149

henryre: Is F.cross_ent not able to operate over axes?

test/classification/test_loss.py
```diff
@@ -0,0 +1,94 @@
...
+        Y_golds_probs = torch.Tensor(preds_to_probs(Y_golds.numpy(), num_classes=3))
+        # Predict [1, 1, 1]
+        Y_probs = torch.tensor(
+            [[-100.0, 100.0, -100.0], [-100.0, 100.0, -100.0], [-100.0, 100.0, -100.0]]
+        )
+        weight1 = torch.FloatTensor([1, 1, 1])
+        weight2 = torch.FloatTensor([1, 2, 1])
+        weight3 = weight2 * FACTOR
+
+        ces_loss0 = cross_entropy_soft(Y_probs, Y_golds_probs).numpy()
```
henryre: These are difficult to read bleow

test/classification/test_loss.py
```diff
@@ -0,0 +1,94 @@
...
+        ces_loss1 = cross_entropy_soft(Y_probs, Y_golds_probs, weight=weight1).numpy()
+        ces_loss2 = cross_entropy_soft(Y_probs, Y_golds_probs, weight=weight2).numpy()
+        ces_loss3 = cross_entropy_soft(Y_probs, Y_golds_probs, weight=weight3).numpy()
+        # Note that PyTorch's cross-entropy loss has the unusual behavior weights
```
henryre: nit: wording mixup

snorkel/classification/__init__.py
```diff
@@ -1,9 +1,18 @@
 """PyTorch-based multi-task learning framework for discriminative modeling."""
 
 from .data import DictDataLoader, DictDataset  # noqa: F401
+from .loss import (  # noqa: F401
+    cross_entropy_from_outputs,
```
henryre: These are all user-facing?

snorkel/classification/loss.py
```diff
@@ -0,0 +1,122 @@
...
+    "categorical_crossentropy" accept float labels natively.
+
+    Note that the method signature is intentionally very similar to F.cross_entropy()
+    so that it can be used as a drop-in replacement when target labels are changed from
+    from a 1D tensor of ints to a 2D tensor of probabilities.
+
+    Parameters
+    ----------
+    input
+        A [num_points, num_classes] tensor of logits
```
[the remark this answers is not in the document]
bhancock8: No, they're paraphrases from PyTorch that remove some of the longer equations and add in shape details.

snorkel/classification/loss.py
```diff
@@ -0,0 +1,122 @@
+from typing import Callable, List, Mapping, Optional, Sequence, Tuple, Union
+
+import torch
+import torch.nn.functional as F
+
+Outputs = Mapping[str, List[torch.FloatTensor]]
+
+
+def cross_entropy_soft(
+    input: torch.FloatTensor,
```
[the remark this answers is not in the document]
bhancock8: Yeah, I was torn on this. Originally did `inputs` instead to avoid the keyword, but ended up sticking with `input` to mirror PyTorch's `F.cross_entropy()` signature.

snorkel/classification/loss.py
```diff
@@ -0,0 +1,122 @@
...
+        return cum_losses
+    elif reduction == "mean":
+        return cum_losses.mean()
+    elif reduction == "sum":
+        return cum_losses.sum()
+    else:
+        raise ValueError("Keyword 'reduction' must be one of ['none', 'mean', 'sum']")
+
+
+def ce_soft_loss_from_outputs(
```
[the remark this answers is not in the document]
bhancock8: Yeah, `cross_entropy_with_probs_from_outputs` is user-facing (sort of). The default loss func is `cross_entropy_from_outputs` and the user can pass in the `with_probs` version to do soft labels instead. You can see why I was tempted by `ce` instead of `cross_entropy` and `soft` instead of `with_probabilities`. But you know, a lot of loss functions have long names in deep learning libraries.

snorkel/classification/loss.py
```diff
@@ -0,0 +1,122 @@
...
+        The gold labels (integers) to calculate loss from
+    active
+        A mask of which data points to consider when calculating loss
+
+    Returns
+    -------
+    torch.Tensor
+        The calculated loss
+    """
+    return F.cross_entropy(outputs[op_name][0][active], (Y.view(-1))[active])
```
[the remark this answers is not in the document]
bhancock8: I've wanted to get rid of the 'active' argument as well. The issue is that doing so puts us in charge of any reduction that happens. e.g., we'll have to decide whether to average or sum the losses in a batch instead of leaving it up to the user, and some loss functions (e.g., ranking loss) are at the batch level rather than the individual example level anyway. Passing `active` is the most flexible (the user can use it or not; all we care about is that their function returns a number at the end of the day), but being more prescriptive (doing the filtering on the model side before the loss function and forcing that we always do the 'mean' reduction) is cleaner. I guess I'm leaning more toward the latter, since it covers the most common use case and makes it easier for the user to swap in their own function for ours (ones less arg to worry about in the signature). @vincentschen any thoughts here?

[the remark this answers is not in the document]
bhancock8: Ah, I remembered the other issue: we pass the function the whole outputs dictionary (so that it can use multiple outputs as in multiple-choice settings or an output from a layer other than the last one as in hierarchical MTL settings ); we don't know beforehand which output(s) to filter and filtering all outputs feels clumsy/wasteful. If we were to make the assumption that we will always calculate losses and predictions using only the output from the last operation, that would simplify a number of things. And come to think of it, I believe that the unusual cases mentioned above (multiple choice, hierarchical MTL) could be handled with a sort of trivial final operation that just identity maps from whatever layers are necessary to the fields the loss function needs. Let's do that in a separate PR, since it has more to do with model execution rather than the probabilistic cross entropy loss.

snorkel/classification/__init__.py
bhancock8: Yeah, the two standard losses (soft and hard) in _from_outputs format, and the raw loss function to be a general purpose corollary to F.cross_entropy().

snorkel/classification/loss.py
bhancock8: I wish. F.cross_entropy() has no axis arg; input has to be [num_points, num_classes] and target has to be [num_points]. It really is painful having a for loop in the loss calculation, but I think writing our own vectorized version not in cuda would still be slower.

bhancock8:
Comments addressed and passing tests.

snorkel/classification/loss.py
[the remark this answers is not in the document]
henryre: Those are too long to read. Assuming you're on top of the right call here.

PR #1367, opened 2019-08-08
Augmentation docs

brahmaneya:
**Test Plan**
Only documentation changes. 
tox and tox -e doctest pass.


henryre:
@brahmaneya you should be able to rerun travis and pass

PR #1369, opened 2019-08-08
Adding candidate_viewer.py

phosseini:
This is a class which contains methods for viewing the extracted candidates in snorkel. In current version, there're two methods:
- print_candidates for printing extracted candidates (keyword-based printing can also be done using this method)
- get_doc_text for getting information of a doc including its text string and corresponding sentences along with their starting indexes

### DIFF
--- a/snorkel/contrib/candidate_viewer.py
+++ b/snorkel/contrib/candidate_viewer.py
@@ -1,4 +1,3 @@
-#!/usr/bin/env python
 """
 Viewing/Printing information of extracted candidates in Snorkel
 """
--- a/snorkel/contrib/candidate_viewer.py
+++ b/snorkel/contrib/candidate_viewer.py
@@ -1,3 +1,5 @@
+#!/usr/bin/env python
+
 import sqlite3
 
 __author__ = "Pedram Hosseini"
@@ -38,7 +40,7 @@ class CandidateViewer:
             conn = sqlite3.connect(db_path)
             return conn
         except sqlite3.Error as e:
-            print("[log-CandidateViewer] error in creating connection to snorkel database. [Details]: " + str(e))
+            print("[log-CandidateViewer] error in creating connection to snorkel database. [Details]: {0}".format(str(e)))
             return None
 
     def _db_select(self, query):
@@ -48,7 +50,7 @@ class CandidateViewer:
             rows = cur.fetchall()
             return rows
         except Exception as e:
-            print("[log-CandidateViewer] failed to execute select query. [Details]: " + str(e))
+            print("[log-CandidateViewer] failed to execute select query. [Details]: {0}".format(str(e)))
             return None
 
     def _get_span_text(self, span_id):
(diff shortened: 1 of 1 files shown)


snorkel/contrib/candidate_viewer.py
```diff
@@ -0,0 +1,128 @@
...
+        """
+        getting text string of a document along with list of sentences and their starting indexes
+        :param doc_name: name of the document
+        :return: doc_text, doc_sents, doc_sents_start_idx
+        """
+        doc_sents = []
+        doc_sents_start_idx = []
+
+        tmp = self._db_select("SELECT id FROM document WHERE name = \"" + doc_name + "\"")
+        if tmp is not None:
```
vincentschen: nit: more pythonic to use `if tmp:` where applicable?

snorkel/contrib/candidate_viewer.py
```diff
@@ -0,0 +1,128 @@
...
+            doc_id = tmp[0][0]
+            doc_sent_query = "SELECT stable_id FROM context WHERE type = \"sentence\" and stable_id like \"" + doc_name + "::%\""
+            doc_sent_info = self._db_select(doc_sent_query)
+            doc_sent_query = "SELECT text FROM sentence WHERE document_id = " + str(doc_id) + ""
+            doc_sent_text = self._db_select(doc_sent_query)
+
+            if doc_sent_info is not None and doc_sent_text is not None:
```
vincentschen: nit: see above, `if doc_sent_info`...

snorkel/contrib/candidate_viewer.py
```diff
@@ -0,0 +1,128 @@
+import sqlite3
+
+
+class CandidateViewer:
```
vincentschen: can you add a doctoring with init parameters and example usage here?

phosseini: I added a docstring and example usage, but please let me know if the format is not correct.

bhancock8:
Thanks for the contribution, @phosseini!

snorkel/contrib/candidate_viewer.py
```diff
@@ -0,0 +1,163 @@
...
+                doc_text = doc_text + doc_sent_text[len(doc_sent_text)-1][0]
+
+                # adding information of the last sentence
+                id_info = doc_sent_info[len(doc_sent_info) - 1][0].split(":")
+                doc_sents.append(doc_sent_text[len(doc_sent_text)-1][0])
+                doc_sents_start_idx.append(int(id_info[3]))
+
+                return doc_text, doc_sents, doc_sents_start_idx
+            return "", [], []
+        return "", [], []
```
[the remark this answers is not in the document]
phosseini: If I remove it and in case user passes an arbitrary value as `doc_name`, calling the method will result in the following error: `TypeError: 'NoneType' object is not iterable` please correct me if I'm wrong

PR #1371, opened 2019-08-08
Make NumPy compatibility for LFApplier explicit

henryre:
Has been compatible all along, but test that and message correctly. Also add some doctests

### DIFF
--- a/test/labeling/apply/test_lf_applier.py
+++ b/test/labeling/apply/test_lf_applier.py
@@ -49,6 +49,16 @@ def h(x: DataPoint) -> int:
     return -1
 
 
+@labeling_function()
+def f_np(x: DataPoint) -> int:
+    return 0 if x[1] > 42 else -1
+
+
+@labeling_function(resources=dict(db=[3, 6, 9]))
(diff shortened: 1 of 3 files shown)


test/labeling/apply/test_lf_applier.py
```diff
@@ -93,9 +103,15 @@ def fp_memoized(x: DataPoint) -> int:
     def test_lf_applier_no_labels(self) -> None:
         data_points = [SimpleNamespace(num=num) for num in DATA]
         applier = LFApplier([h])
-        L = applier.apply(data_points)
+        L = applier.apply(data_points, progress_bar=False)
         np.testing.assert_equal(L, -1)
 
+    def test_lf_applier_numpy(self) -> None:
+        X = np.vstack((DATA, DATA)).T
```
paroma: is this to explicitly create 2D array to differentiate from doctest?

paroma: why not `X = DATA` and remove the indexing in the LF

henryre: because the point is to test that it works with ndarrays

PR #1375, opened 2019-08-09
Add details to logging message on shared modules/ops

bhancock8:
Make more informative logging statement when creating a network (for sanity checking how many modules/operations should be shared).

**Test plan:**
Constructed model locally and saw that logging statement looked correct.

### DIFF
--- a/snorkel/classification/snorkel_classifier.py
+++ b/snorkel/classification/snorkel_classifier.py
@@ -102,8 +102,8 @@ class SnorkelClassifier(nn.Module):
         logging.info(
             f"Created multi-task model {self.name} that contains "
             f"task(s) {self.task_names} from "
-            f"{len(unique_ops)} ({len(all_ops) - len(unique_ops)} shared) and "
-            f"{len(unique_mods)} ({len(all_mods) - len(unique_mods)} shared)."
+            f"{len(unique_ops)} operations ({len(all_ops) - len(unique_ops)} shared) and "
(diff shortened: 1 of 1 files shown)


snorkel/classification/snorkel_classifier.py
```diff
@@ -94,9 +94,16 @@ def __init__(
...
+        # Report total task count and duplicates
+        all_ops = [op.name for t in tasks for op in t.task_flow]
+        unique_ops = set(all_ops)
+        all_mods = [mod_name for t in tasks for mod_name in t.module_pool]
+        unique_mods = set(all_mods)
         logging.info(
             f"Created multi-task model {self.name} that contains "
-            f"task(s) {self.task_names}."
+            f"task(s) {self.task_names} from "
+            f"{len(unique_ops)} operations ({len(all_ops) - len(unique_ops)} shared) and "
```
brahmaneya: I am not very familiar with pytorch, so ignore if this seems incorrect, but I think
f"{len(all_ops)} total ops, {len(unique_ops) unique ops} and"
f"{len(all_mods)} total modules, {len(unique_mods) unique modules}"
is more understandable.

bhancock8: It's a good question, but in this case, I think it makes more sense to talk about shared modules. e.g., if I have a shared body and five unshared heads, the more important piece of info to me is that of my 6 modules, only 1 of them is shared between more than one task.

PR #1376, opened 2019-08-10
Add Spark version of NLPLabelingFunction

henryre:
Useful for e.g. Drybell example. Need to go in and make the preprocessor Spark-compatible. Looks big, but just abstracting most stuff to a parent class (which is almost the same as NLPLF now), then adding two extremely simple subclasses

**Test plan**
* Added unit tests
* Local `tox -e spark`

### DIFF
--- a/snorkel/labeling/lf/nlp.py
+++ b/snorkel/labeling/lf/nlp.py
@@ -27,22 +27,87 @@ class SpacyPreprocessorConfig(NamedTuple):
     parameters: SpacyPreprocessorParameters
 
 
-class NLPLabelingFunction(LabelingFunction):
-    r"""Special labeling function type for SpaCy-based LFs.
+class BaseNLPLabelingFunction(LabelingFunction):
+    """Base class for spaCy-based LFs."""
+
+    _nlp_config: SpacyPreprocessorConfig
+
+    @classmethod
+    def _create_preprocessor(
+        cls, parameters: SpacyPreprocessorParameters
+    ) -> SpacyPreprocessor:
+        raise NotImplementedError
+
+    @classmethod
+    def _create_or_check_preprocessor(
+        cls,
+        text_field: str,
+        doc_field: str,
+        language: str,
+        disable: Optional[List[str]],
+        pre: List[BasePreprocessor],
+        memoize: bool,
+    ) -> None:
+        # Create a SpacyPreprocessor if one has not yet been instantiated.
+        # Otherwise, check that configuration matches already instantiated one.
+        parameters = SpacyPreprocessorParameters(
+            text_field=text_field,
+            doc_field=doc_field,
+            language=language,
+            disable=disable,
+            pre=pre,
+            memoize=memoize,
+        )
+        if not hasattr(cls, "_nlp_config"):
+            nlp = cls._create_preprocessor(parameters)
+            cls._nlp_config = SpacyPreprocessorConfig(nlp=nlp, parameters=parameters)
+        elif parameters != cls._nlp_config.parameters:
+            raise ValueError(
+                "{cls.__name__} already configured with different parameters: "
+                f"{cls._nlp_config.parameters}"
+            )
+
+    def __init__(
+        self,
+        name: str,
+        f: Callable[..., int],
+        resources: Optional[Mapping[str, Any]] = None,
+        pre: Optional[List[BasePreprocessor]] = None,
+        fault_tolerant: bool = False,
+        text_field: str = "text",
+        doc_field: str = "doc",
+        language: str = EN_CORE_WEB_SM,
+        disable: Optional[List[str]] = None,
+        memoize: bool = True,
+    ) -> None:
+        self._create_or_check_preprocessor(
+            text_field, doc_field, language, disable, pre or [], memoize
+        )
+        super().__init__(
+            name,
+            f,
+            resources=resources,
+            pre=[self._nlp_config.nlp],
+            fault_tolerant=fault_tolerant,
+        )
+
+
+class NLPLabelingFunction(BaseNLPLabelingFunction):
+    r"""Special labeling function type for spaCy-based LFs.
 
     This class is a special version of ``LabelingFunction``. It
     has a ``SpacyPreprocessor`` integrated which shares a cache
     with all other ``NLPLabelingFunction`` instances. This makes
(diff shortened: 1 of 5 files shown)


henryre:
Note to self: drop make_spark return

snorkel/labeling/lf/nlp.py
```diff
@@ -243,3 +224,48 @@ def __call__(self, f: Callable[..., int]) -> NLPLabelingFunction:
...
+    doc_field
+        Name of data point field to output parsed document to
+    language
+        spaCy model to load
+        See https://spacy.io/usage/models#usage
+    disable
+        List of pipeline components to disable
+        See https://spacy.io/usage/processing-pipelines#disabling
+    memoize
+        Memoize preprocessor outputs?
```
brahmaneya: I think the statement "Whether to memoize preprocessor outputs" seems better than having a question.
Same for all other instances of memoize.

snorkel/labeling/lf/nlp.py
```diff
@@ -243,3 +224,48 @@ def __call__(self, f: Callable[..., int]) -> NLPLabelingFunction:
...
+    ... def has_person_mention(x):
+    ...     person_ents = [ent for ent in x.doc.ents if ent.label_ == "PERSON"]
+    ...     return 0 if len(person_ents) > 0 else -1
+    >>> has_person_mention
+    NLPLabelingFunction has_person_mention, Preprocessors: [SpacyPreprocessor...]
+
+    >>> from types import SimpleNamespace
+    >>> x = SimpleNamespace(text="The movie was good.")
+    >>> has_person_mention(x)
+    -1
```
brahmaneya: Do we not document inherited attributes?

snorkel/labeling/lf/nlp_spark.py
```diff
@@ -0,0 +1,106 @@
...
+    BaseNLPLabelingFunction,
+    SpacyPreprocessorParameters,
+    base_nlp_labeling_function,
+)
+
+
+class SparkNLPLabelingFunction(BaseNLPLabelingFunction):
+    r"""Special labeling function type for SpaCy-based LFs running on Spark.
+
+    This class is a Spark-compatabile version of ``NLPLabelingFunction``.
```
brahmaneya: compatible

snorkel/labeling/lf/nlp_spark.py
```diff
@@ -0,0 +1,106 @@
...
+    name
+        Name of the LF
+    f
+        Function that implements the core LF logic
+    resources
+        Labeling resources passed in to ``f`` via ``kwargs``
+    pre
+        Preprocessors to run before SpacyPreprocessor is executed
+    fault_tolerant
+        Output -1 if LF execution fails?
```
brahmaneya: If True, output -1 if ...

snorkel/labeling/lf/nlp_spark.py
```diff
@@ -0,0 +1,106 @@
...
+    fault_tolerant
+        See above
+    """
+
+    @classmethod
+    def _create_preprocessor(
+        cls, parameters: SpacyPreprocessorParameters
+    ) -> SpacyPreprocessor:
+        preprocessor = SpacyPreprocessor(**parameters._asdict())
+        make_spark_preprocessor(preprocessor)
```
brahmaneya: The name make_spark_preprocessor makes it sound like it returns a new preprocessor, rather than modifying the given one in place. In this case, it does both. But it still seems more readable to 
```
 preprocessor = write make_spark_preprocessor(preprocessor)
```
Or even better, to rename make_spark_preprocessor to something like make_spark_compatible

test/labeling/lf/test_nlp_spark.py
```diff
@@ -0,0 +1,47 @@
...
+
+
+def has_person_mention(x: DataPoint) -> int:
+    person_ents = [ent for ent in x.doc.ents if ent.label_ == "PERSON"]
+    return 0 if len(person_ents) > 0 else -1
+
+
+@pytest.mark.spark
+class TestNLPLabelingFunction(unittest.TestCase):
+    def _run_lf(self, lf: SparkNLPLabelingFunction) -> None:
```
brahmaneya: check_person_mention_lf?
_run_lf sounds like it runs an arbitrary lf, but it contains specific checks related to person mentions.

test/labeling/lf/test_nlp_spark.py
```diff
@@ -0,0 +1,47 @@
...
+        self.assertEqual(lf(x), 0)
+
+    def test_nlp_labeling_function(self) -> None:
+        lf = SparkNLPLabelingFunction(name="my_lf", f=has_person_mention)
+        self._run_lf(lf)
+
+    def test_nlp_labeling_function_decorator(self) -> None:
+        @spark_nlp_labeling_function()
+        def has_person_mention(x: DataPoint) -> int:
+            person_ents = [ent for ent in x.doc.ents if ent.label_ == "PERSON"]
```
brahmaneya: You could just call the global has_person_mention(x) within this function instead of repeating its code (rename this to avoid collison).

snorkel/labeling/lf/nlp.py
henryre: I've seen the question format for bools quite a bit. Let's leave for now.

snorkel/labeling/lf/nlp.py
henryre: Nope, we should! I'll update

snorkel/labeling/lf/nlp_spark.py
henryre: `make_spark_preprocessor` monkey patches then returns the preprocessor, so the syntax you wrote is valid. but mypy has trouble since `make_spark_preprocessor` has a return type of `Mapper`, so for this internal version, it makes sense to do it like this.

test/labeling/lf/test_nlp_spark.py
henryre: that's a bit harder to read than this, so i'll leave it as it is here

snorkel/labeling/lf/nlp.py
henryre: oh sorry, the attributes for these decorators aren't accessible, so we skip those in doc

PR #1378, opened 2019-08-11
[EZ] Fix typo

henryre:


bhancock8:
Looks like the travis failure was due to the stochastic label model test. Pushed the restart button.

PR #1380, opened 2019-08-11
Simplify model definitions

bhancock8:
Simplify the MultitaskClassifier definition! Biggest benefit: this significantly simplifies user-defined loss_func and output_func methods.
- Assume loss_func and output_func are passed only the output of the last operation
- Drop *_from_output versions of the cross_entropy and softmax, because they're no longer needed
- Remove the need to specify the trivial "0" for most operation input lists
- Perform active filtering in the MultitaskClassifier rather than the loss_func and output_func

**Test plan:**
`tox`
All these functionalities had corresponding tests which were broken by the change but have been updated to reflect the new expected signatures.
This will require updates in tutorials that use the MultitaskClassifier.


snorkel/classification/multitask_classifier.py
```diff
@@ -271,10 +264,16 @@ def calculate_loss(
...
+                # Filter out any inactive examples if inputs is a Tensor
+                if not active.all() and isinstance(inputs, torch.Tensor):
+                    inputs = inputs[active]
+                    Y = Y[active]
+
                 loss_dict[label_name] = self.loss_funcs[task_name](
-                    outputs=outputs,
-                    Y=move_to_device(Y, self.config.device),
-                    active=move_to_device(active, self.config.device),
+                    inputs, move_to_device(Y, self.config.device)
```
vincentschen: nice! cleaner

snorkel/classification/multitask_classifier.py
```diff
@@ -27,7 +27,7 @@
 
 from .task import Operation, Task
 
-OutputDict = Dict[str, Mapping[Union[str, int], Any]]
+OutputDict = Dict[str, Union[Any, Mapping[str, Any]]]
```
vincentschen: this feels like a lose constraint— not `Any` is not a Tensor or anything?

snorkel/slicing/modules/slice_combiner.py
```diff
@@ -54,18 +54,18 @@ def __init__(
         self.temperature = temperature
 
     def forward(  # type:ignore
-        self, flow_dict: Dict[str, List[torch.Tensor]]
+        self, output_dict: Dict[str, torch.Tensor]
```
vincentschen: independent from the MultitaskClassifier, it felt a little confusing to name this `output_dict`, given its the input to `forward`

okay for now, but we could consider changing names if something better comes to mind

bhancock8:
Good comments. Will address in subsequent PR. Merging this to unblock other things in motion.

PR #1391, opened 2019-08-13
🎉 redux

henryre:
🎉


henryre:
@vincentschen I disabled squash merges so I wouldn't blow it.

henryre:
^ dang you're right. @stephenbach quick PR?

PR #1394, opened 2019-08-14
[EASY] Rename to SlicingClassifier

vincentschen:
Renames `BinarySlicingClassifier` -> `SlicingClassifier`

**Test Plan**
* Tox passes


vincentschen:
> Only question is: does the SlicingClassifier support more than 2 classes? If not, do we clearly catch that case somewhere with NotImplementedErrors?

@bhancock8  It doesn't, but these will be caught in `SliceCombiner` down the line: https://github.com/snorkel-team/snorkel/blob/master/snorkel/slicing/modules/slice_combiner.py#L96

PR #1395, opened 2019-08-14
PyTorch 1.2 compatibility

henryre:


dataframing:
@henryre I took a look at this branch's Travis build logs, since it would be great to have Snorkel be compatible with PyTorch 1.2.0. It seems like the build fails when running the `mypy` type-checker under `tox -e type`, specifically referring to L102 of `snorkel/classification/multitask_classifier.py`, shown below:

https://github.com/snorkel-team/snorkel/blob/cd283f52f8e8fef4efb199a8e5f487b3f0b8d227/snorkel/classification/multitask_classifier.py#L102

Tracing through the type hints, `tasks` is of type `List[Task]` (which implements the iterator protocol), but `t.module_pool` is of type `nn.ModuleDict` which doesn't implement the `__next__` method. It does implement the `__iter__` method, but I don't think `mypy` understands iterables that don't define `__next__`. I was able to ignore this line and have the rest of the testing pipeline work as-intended (running `tox` to check all sub-commands):

```python
all_mods = [
    mod_name for t in tasks for mod_name in t.module_pool  # type: ignore
]
```

Obviously not ideal, but could this be an acceptable work-around for now? I suspect the proper solution is to make an upstream issue/PR into PyTorch which defines a trivial `__next__` on `nn.ModuleDict`.

henryre:
Hi @dataframing thanks for checking in here! We left this sitting as a staging branch since we're planning on making v0.10.0 compatible with PyTorch 1.2. We'll update it with the latest changes in the next couple weeks, so this will be super helpful! In general, mypy doesn't play super nicely with PyTorch, so we use the `ignore` approach in a few places.

bhancock8:
Let's jump straight to 1.3, the current stable branch of PyTorch.

PR #1399, opened 2019-08-14
Update contrib guidelines

henryre:
Only documentation

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -28,7 +28,7 @@ For example, you can simply activate this environment and start using Snorkel:
 ```bash
 source .env/bin/activate
 
-python3 -c "import snorkel; print(dir(snorkel))"
+python3 -c "import snorkel; print(dir(snorkel.labeling))"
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.md
```diff
@@ -124,15 +136,23 @@ action (e.g. `.collect()`).
 ### Submitting PRs
 
 When submitting a PR, make sure to use the preformatted template.
+Except in special cases, all PRs should be against `master`.
+Avoid using "staging branches" as much as possible.
+If you want to add complicated features, please
+[stack your PRs](https://graysonkoonce.com/stacked-pull-requests-keeping-github-diffs-small/)
```
brahmaneya: This seems good in principle, but very complicated. Have you actually seen anyone do this, (as opposed to just breaking into multiple PRs and sending them for review one after the other)?

henryre: It’s not so complicated! We did it quite a bit for core Snorkel development on v0.9!

PR #1420, opened 2019-08-19
Link to code of conduct

henryre:

Seems more common to add link to code of conduct rather than document itself. Making this change for tutorials as well.

## Test plan

N/A


CONTRIBUTING.md
```diff
@@ -28,7 +30,7 @@ For example, you can simply activate this environment and start using Snorkel:
 ```bash
 source .env/bin/activate
 
-python3 -c "import snorkel; print(dir(snorkel.labeling))"
+python3 -c "import snorkel.labeling; print(dir(snorkel.labeling))"
```
vincentschen: looks good, but why not just `import snorkel; print(dir(snorkel))`?

henryre: won't be informative

PR #1429, opened 2019-08-25
Add gh workflow to close stale issues and PRs

vincentschen:
Add github workflow to close stale prs/issues: https://github.com/actions/stale

## Test plan
Will check monitor workflow after it is deployed; does not affect core package

## Checklist

Need help on these? Just ask!

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -1,0 +1,21 @@
+name: Mark/close stale issues and pull requests
+
+on:
+  schedule:
+  - cron: "0 * * * *"
+
+jobs:
+  stale:
+
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/stale@v1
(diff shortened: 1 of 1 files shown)


vincentschen:
ping @henryre for visibility because you've configured other github workflows

henryre:
what's the deal with help wanted issues? also do we need to run this every hour? feels like a bad use of fossil fuels

vincentschen:
> what's the deal with help wanted issues? 

The [issue](https://github.com/actions/stale/issues/10) is picking up momentum :) Until then, we can trivially remove the stale label. I'd vote to ship this first to make sure we like the workflow.

> also do we need to run this every hour? feels like a bad use of fossil fuels



This was the default; happy to change to daily or even weekly. That said, there are probably higher-leverage ways to save the earth... 

❌ ⛏

Issue #1437, opened 2019-08-30
negative predictions for data point with positive labels


pengdumle:
Hi,
I am using snorkel for a search result quality prediction project.
I have quite a few cases where the predictions do not quite make sense to me.
There are false negative results where the predictions are 0 (negative), but if you look at the labels assigned by each labeling functions, they are either 1(positive) or -1(abstain). There are nearly no 0(negative) at all.
Could anyone help clarify this for me please?
Thanks,
Peng

paroma:
Can you share the L matrix and details about how you're training the LabelModel? You can also take a look at [this issue](https://github.com/snorkel-team/snorkel/issues/1422#issuecomment-523593367) for details on how ties are broken, which could be causing this behavior here

pengdumle:
Sure.
Here are the L matrix and predictions:
L_test[:20]
array([[-1, -1, -1, -1, -1, -1, -1,  0,  0,  1, -1,  1,  1,  1,  1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [-1, -1, -1, -1, -1, -1, -1,  0,  0,  1, -1,  1,  1,  1,  1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [-1, -1, -1, -1, -1, -1, -1,  0,  0,  1, -1,  1,  1,  1,  1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [-1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1, -1,  0, -1, -1,
        -1, -1,  0, -1, -1, -1, -1, -1, -1, -1],
       [ 1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 1,  1,  1,  1,  1,  1,  1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 1,  0,  0, -1,  1,  1,  1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
        -1, -1, -1, -1, -1, -1, -1, -1, -1, -1],
       [ 1,  1,  1, -1,  1,  1,  1, -1, -1, -1, -1, -1, -1, -1, -1, -1,
[...]

PR #1444, opened 2019-09-04
LabelModel parameter clamping fix (fixes issue 1422)

### LINKED ISSUE #1422: LabelModel produces equal probability for labeled data
## Issue description

I am using snorkel for creating binary text classification training examples with 9 labeling functions. However, I find that for some data that are trained with the label model, they receive a probabilistic label of equal probabilities (i.e. `[0.5 0.5]`), despite that they only receive label of one class from the labeling functions (e.g. `[-1  0 -1 -1  0 -1 -1 -1 -1]`, so only class `0` or `ABSTAIN`), why is that?

Besides, I find that setting `verbose = True` when defining the `LabelModel` does not print  the logging information.

Lastly, if producing the label `[0.5 0.5]` is a normal behavior, then when filtering out unlabeled data points, they should be removed as well, because they do not contribute to the training (of a classifier), and if the classifier that does not support probablistic labels  (e.g. FastText), using `argmax` will always lead to class `0` (which is undesired).

## Code example/repro steps

Below I show my code of defining `LabelModel`:
```
print('==================================================')
print('Training label model...')
label_model = LabelModel(cardinality=2, verbose=True)
label_model.fit(L_train=L_train, n_epochs=10000, lr=0.001, log_freq=100, seed=1)
print('Done!')
```

Below I show some of the logs I print:
```
Check results for data in the training set:  a_text_in_the_training_set_but_I_removed_here
        * Output of L_train for this data point is: [-1  0 -1 -1  0 -1 -1 -1 -1]
        * Output of probs_train for this data point is: [0.5 0.5]
        * Output of probs_train_filtered for this data point is: [0.5 0.5]
        * Output of preds_train_filtered for this data point is: 0
```

## Expected behavior

I expect that if a datapoint receives label of only a single class, then it should not get equal probabilities of both classes when the label model is trained.

## System info

* Build command you used (if compiling from source):
* OS: Windows
* Python version: 3.7
* Snorkel version: 0.9
* Versions of any other relevant libraries:

ajratner:
_High level:_ This change makes the clamping threshold of `LabelModel` parameters a hyperparameter in `train_config`, and sets to a much smaller default value to avoid issues in sparse label setting.

_In more detail:_ The `LabelModel` learns the conditional probabilities of the labeling functions outputting certain labels given certain true (but unobserved or latent) true labels, e.g. `P(\lambda = x | Y = y)`. We also clamp these estimates to some min / max values just to guard against pathological errors... but as one might guess, when `P(\lambda = x | Y = y)` is really small- for example in the common sparse label setting where LFs mostly abstain- then clamping at too high of a minimum value messes everything up- we basically save clamped parameters that say these sparse LFs are equally likely to be wrong versus right! Luckily, this is fixed by changing the clamping parameter, which is now a kwarg in `LabelModel.fit`, and defaults to a much smaller value.

## Related issue(s)
- **Fixes #1422**
- Possibly addresses #1437 (TBD)

## Test plan
- Added a test (marked complex) that successfully (a) replicates the original error in #1422 and (b) solves it.

## Checklist

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -46,8 +46,8 @@ class TrainConfig(Config):
         A random seed to initialize the random number generator with
     log_freq
         Report loss every this many epochs (steps)
-    mu_clamp_min
-        Restrict the learned conditional probabilities to [mu_clamp_min, 1-mu_clamp-min]
+    mu_clamp_eps
+        Restrict the learned conditional probabilities to [mu_clamp_eps, 1-mu_clamp_eps]
     """
 
     n_epochs: int = 100
@@ -60,7 +60,7 @@ class TrainConfig(Config):
     prec_init: float = 0.7
     seed: int = np.random.randint(1e6)
     log_freq: int = 10
-    mu_clamp_min: float = 0.0001
+    mu_clamp_eps: float = 0.0001
 
 
 class LabelModelConfig(Config):
@@ -804,10 +804,11 @@ class LabelModel(nn.Module):
             self._update_lr_scheduler(epoch)
 
         # Clamp learned parameters
-        # Note: If mu_clamp_min is set too high, e.g. in sparse settings where LFs
-        # mostly abstain, this will cause null results!
+        # Note: If mu_clamp_eps is set too high, e.g. in sparse settings where LFs
+        # mostly abstain, this will result in learning conditional probabilities all
+        # equal to mu_clamp_eps (and/or 1 - mu_clamp_eps)!
         self.mu.data = self.mu.clamp(  # type: ignore
-            self.train_config.mu_clamp_min, 1 - self.train_config.mu_clamp_min
+            self.train_config.mu_clamp_eps, 1 - self.train_config.mu_clamp_eps
         )
 
         # Return model to eval mode
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -60,7 +60,7 @@ class TrainConfig(Config):
     prec_init: float = 0.7
     seed: int = np.random.randint(1e6)
     log_freq: int = 10
-    mu_eps: float = -1.0
+    mu_eps: Optional[float] = None
 
 
 class LabelModelConfig(Config):
@@ -810,10 +810,10 @@ class LabelModel(nn.Module):
         # Note: Use user-provided value, else default to 1 / n', where n' is n rounded
         # to the closest power of ten; this rounding is done to make it more obvious
         # when the parameters have been clamped.
-        if self.train_config.mu_eps >= 0:
+        if self.train_config.mu_eps is not None:
             mu_eps = self.train_config.mu_eps
         else:
-            mu_eps = min(0.01, 1 / 10 ** np.round(np.log10(self.n)))
+            mu_eps = min(0.01, 1 / 10 ** np.ceil(np.log10(self.n)))
         self.mu.data = self.mu.clamp(mu_eps, 1 - mu_eps)  # type: ignore
 
         # Return model to eval mode
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -46,6 +46,8 @@ class TrainConfig(Config):
         A random seed to initialize the random number generator with
     log_freq
         Report loss every this many epochs (steps)
+    mu_clamp_min
+        Restrict the learned conditional probabilities to [mu_clamp_min, 1-mu_clamp-min]
     """
 
     n_epochs: int = 100
@@ -58,6 +60,7 @@ class TrainConfig(Config):
     prec_init: float = 0.7
     seed: int = np.random.randint(1e6)
     log_freq: int = 10
+    mu_clamp_min: float = 0.0001
 
 
 class LabelModelConfig(Config):
@@ -319,7 +322,6 @@ class LabelModel(nn.Module):
             # The 0th row (corresponding to abstains) is the difference between
             # the sums of the other rows and one, by law of total prob
             c_probs[i * (self.cardinality + 1), :] = 1 - mu_i.sum(axis=0)
-        c_probs = np.clip(c_probs, 0.01, 0.99)
 
         if source is not None:
             return c_probs[
@@ -377,7 +379,7 @@ class LabelModel(nn.Module):
         L_shift = L + 1  # convert to {0, 1, ..., k}
         self._set_constants(L_shift)
         L_aug = self._get_augmented_label_matrix(L_shift)
-        mu = np.clip(self.mu.detach().clone().numpy(), 0.01, 0.99)
+        mu = self.mu.detach().clone().numpy()
         jtm = np.ones(L_aug.shape[1])
 
         # Note: We omit abstains, effectively assuming uniform distribution here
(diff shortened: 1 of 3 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -46,6 +46,8 @@ class TrainConfig(Config):
         A random seed to initialize the random number generator with
     log_freq
         Report loss every this many epochs (steps)
+    mu_clamp_min
```
henryre: mu_clamp_eps

snorkel/labeling/model/label_model.py
```diff
@@ -46,6 +46,8 @@ class TrainConfig(Config):
...
+        Restrict the learned conditional probabilities to [mu_clamp_min, 1-mu_clamp-min]
```
henryre: nit: typo - -> _

snorkel/labeling/model/label_model.py
```diff
@@ -58,6 +60,7 @@ class TrainConfig(Config):
     prec_init: float = 0.7
     seed: int = np.random.randint(1e6)
     log_freq: int = 10
+    mu_clamp_min: float = 0.0001
```
henryre: do we expect anyone to touch this? if no, why 1e-4 and not e.g. 1e-8?

snorkel/labeling/model/label_model.py
```diff
@@ -801,6 +803,14 @@ def fit(
             # Update learning rate
             self._update_lr_scheduler(epoch)
 
+        # Clamp learned parameters
+        # Note: If mu_clamp_min is set too high, e.g. in sparse settings where LFs
+        # mostly abstain, this will cause null results!
```
henryre: what are "null results"?

snorkel/labeling/model/label_model.py
```diff
@@ -801,6 +803,14 @@ def fit(
...
+        self.mu.data = self.mu.clamp(  # type: ignore
```
henryre: Why do this here (and outside of the computation graph) instead of directly at inference time?

henryre:
@ajratner also please change the PR name to something more informative since it will be the commit name

snorkel/labeling/model/label_model.py
ajratner: Orthogonal questions... pass for now :)

snorkel/labeling/model/label_model.py
ajratner: Regarding doing it here- I view this as part of our current procedure for learning / setting `mu`, therefore would go here.  Also safer if mu used in multiple sub-functions.

As for outside the computation graph- `torch.clamp` returns a tensor not a parameter, that's why I did this

snorkel/labeling/model/label_model.py
ajratner: Actually... @henryre thoughts on setting it to `None`, and then defaulting to picking something prop. to `1/n`?  Basically that's setting it to allow params to still be sensitive to LFs that only label a few of the `n` data points.

ajratner: See what I just committed

ajratner:
@henryre ready for re-review

snorkel/labeling/model/label_model.py
```diff
@@ -801,6 +803,20 @@ def fit(
...
+        # Note: If mu_eps is set too high, e.g. in sparse settings where LFs
+        # mostly abstain, this will result in learning conditional probabilities all
+        # equal to mu_eps (and/or 1 - mu_eps)!
+        # Note: Use user-provided value, else default to 1 / n', where n' is n rounded
+        # to the closest power of ten; this rounding is done to make it more obvious
+        # when the parameters have been clamped.
+        if self.train_config.mu_eps >= 0:
+            mu_eps = self.train_config.mu_eps
+        else:
+            mu_eps = min(0.01, 1 / 10 ** np.round(np.log10(self.n)))
```
henryre: Does `ceil` make more sense here? as is now: n = 300 -> mu_eps = 0.01

test/labeling/model/test_label_model.py
```diff
@@ -355,14 +365,14 @@ def setUp(self) -> None:
...
-        """Test the LabelModel's estimate of P and Y."""
+    def test_label_model_basic(self) -> None:
+        """Test the LabelModel's estimate of P and Y on a simple synthetic dataset."""
         np.random.seed(123)
         P, Y, L = generate_simple_label_matrix(self.n, self.m, self.cardinality)
 
         # Train LabelModel
         label_model = LabelModel(cardinality=self.cardinality, verbose=False)
-        label_model.fit(L, n_epochs=200, lr=0.01, seed=123)
+        label_model.fit(L, n_epochs=1000, lr=0.01, seed=123)
```
henryre: why do we need more epochs now?

test/labeling/model/test_label_model.py
```diff
@@ -371,9 +381,39 @@ def test_label_model(self) -> None:
         np.testing.assert_array_almost_equal(P, P_lm, decimal=2)
 
         # Test predicted labels
-        Y_lm = label_model.predict_proba(L).argmax(axis=1)
-        err = np.where(Y != Y_lm, 1, 0).sum() / self.n
-        self.assertLess(err, 0.1)
+        score = label_model.score(L, Y)
+        self.assertGreaterEqual(score["accuracy"], 0.9)
```
henryre: nice!

test/labeling/model/test_label_model.py
```diff
@@ -371,9 +381,39 @@ def test_label_model(self) -> None:
...
+
+        # Test estimated LF conditional probabilities
+        P_lm = label_model._get_conditional_probs().reshape(
+            (self.m, self.cardinality + 1, -1)
+        )
+        np.testing.assert_array_almost_equal(P, P_lm, decimal=2)
+
+        # Test predicted labels *only on non-abstained data points*
+        Y_pred = label_model.predict(L, tie_break_policy="abstain")
+        idx, = np.where(Y_pred != -1)
```
henryre: nit: `idx, _` for readability

test/labeling/model/test_label_model.py
```diff
@@ -371,9 +381,39 @@ def test_label_model(self) -> None:
...
+        acc = np.where(Y_pred[idx] == Y[idx], 1, 0).sum() / len(idx)
+        self.assertGreaterEqual(acc, 0.65)
```
henryre: this feels like a pretty low bar. what is the accuracy without the clamping patch?

test/labeling/model/test_label_model.py
```diff
@@ -344,6 +344,16 @@ def test_warmup(self):
             lr_scheduler_config = {"warmup_steps": 1, "warmup_unit": "batches"}
             label_model.fit(L, lr_scheduler_config=lr_scheduler_config)
 
+    def test_set_mu_eps(self):
+        MU_EPS = 0.0123
```
henryre: nit: lowercase if not a module-level constant

snorkel/labeling/model/label_model.py
```diff
@@ -58,6 +60,7 @@ class TrainConfig(Config):
     prec_init: float = 0.7
     seed: int = np.random.randint(1e6)
     log_freq: int = 10
+    mu_eps: float = -1.0
```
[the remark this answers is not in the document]
ajratner: I get a `mypy` error in my VSCode linter when I have it as `None`- is this something we have fixed in the config?

test/labeling/model/test_label_model.py
ajratner: It's a hard problem when the LFs are this sparse; no way around that.  Without the clamping patch, the accuracy as measured here- not counting abstains- is roughly the same but coverage is far lower because of the error where LFs are counted as having null signal

henryre: Ok cool so `test_label_model_sparse` fails without this patch?

test/labeling/model/test_label_model.py
ajratner: This doesn't work, there's only one element

henryre: ah ok, sg

test/labeling/model/test_label_model.py
ajratner: Yes

PR #1445, opened 2019-09-04
Write complete json log after training

### LINKED ISSUE #1439: Logging functions do not get called
## Issue description

When logging is enabled, and the log_writer is json, nothing outputs to the log file.
Specifically, in the `snorkel/classification/training/loggers/log_writer.py` file, the `write_log` function never gets called anywhere in the code base.

I can directly call this function after training like so:
```
trainer.log_writer.write_log("output_log.json")
```
Which outputs the metric dict at the expected location.

But I feel that this should be done automatically after training (called in the `snorkel/classification/training/trainer.py` file) if logging is enabled.

## Code example/repro steps

Here is my trainer config:
```
trainer_config = {
    "progress_bar": True,              
    "n_epochs": 2, 
    "lr": 0.02, 
    "logging": True, 
    "log_writer": "json",
    "checkpointing": True,
}
```

## Expected behavior
It should output the training & validation metrics to a .json file in the "logs" folder.

bhancock8:
As part of closing the `LogManager` at the end of training, write the run log to file.
This appears to have been neglected before because when the log_writer is set to `tensorboard` (the default), the TensorBoard log is written incrementally over the course of training, whereas the json log is written all at once at the end.

## Related issue(s)
Fixes #1439 

## Test plan
New unit test confirms that log is written at the end of training.

## Checklist

Need help on these? Just ask!

### DIFF
--- a/snorkel/classification/training/loggers/log_writer.py
+++ b/snorkel/classification/training/loggers/log_writer.py
@@ -81,7 +81,7 @@ class LogWriter:
         Parameters
         ----------
         config
-            JSON-compatible config to write to TensorBoard
+            JSON-compatible config to write to file
         config_filename
             Name of file in logging directory to write to
         """
--- a/snorkel/classification/training/loggers/log_writer.py
+++ b/snorkel/classification/training/loggers/log_writer.py
@@ -131,4 +131,4 @@ class LogWriter:
 
     def close(self) -> None:
         """Close writer if necessary."""
-        pass
+        self.write_log("log.json")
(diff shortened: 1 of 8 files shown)


snorkel/classification/training/loggers/log_writer.py
```diff
@@ -131,4 +131,4 @@ def write_json(self, dict_to_write: Mapping[str, Any], filename: str) -> None:
 
     def close(self) -> None:
         """Close writer if necessary."""
-        pass
+        self.write_log("log.json")
```
vincentschen: naming-wise, this (`.close`) feels like the wrong place to do the actual dump to disk, it feels like some cleanup operation should be happening here instead.

If it's a special case / hard to fix class names, let's make it clear in the comments for now

bhancock8: Good point. Refactored close() to cleanup() and added some clarifying docstrings.

Peter-UCL:
Much thanks!

PR #1447, opened 2019-09-05
Fix lf summary no abstain 1446

### LINKED ISSUE #1446: Confusion matrix in LFAnalysis.lf_summary supposes you have at least one ABSTAIN label
## Issue description

I think that the *correct* and *incorrect* columns from the `LFAnalysis().lf_summary(Y)` are wrong when you don't have an `ABSTAIN`, label (i.e. -1) in the input data.

The confusion matrix does not have the same size if the unique labels are [0, 1, 2] or [-1, 0, 1, 2]. I understand that the `[1:, 1:]`slicing avoids the label -1 (which the first row/colomn) here
https://github.com/snorkel-team/snorkel/blob/ad5731af21a0e628a1c186c3c0b9f9963bae4b0c/snorkel/labeling/analysis.py#L357

## Code example/repro steps

```python
from typing import NamedTuple

import numpy as np

from sklearn.metrics import confusion_matrix

from snorkel.labeling import LFAnalysis


class MockLF(NamedTuple):
    name: str


Y = np.array([0, 1, 0, 2, 1, 0, 1, 2])

lfs = [MockLF("one"), MockLF("two"), MockLF("three")]

# suppose you have three labeling functions
L = np.array([[0, 1, 0],
              [0, 0, 1],
              [1, 0, 2],
              [2, 2, 2],
              [0, 1, 1],
              [0, 0, 1],
              [0, 1, 0],
              [0, 2, 1]])

L_w_abstain = np.array([[-1, 1, 0],
                        [0, 0, 1],
                        [1, -1, 2],
                        [2, 2, 2],
                        [0, 1, -1],
                        [0, 0, 1],
                        [0, -1, 0],
                        [0, 2, 1]])

analysis = LFAnalysis(L, lfs=lfs)
print(analysis.lf_summary(Y))

# well_a = LFAnalysis(L_w_abstain, lfs=lfs)
# print(well_a.lf_summary(Y))
```

## Expected behavior

The first labeling function named "one" should have 3 corrects and 5 incorrects instead of

```
       j   Polarity  Coverage  Overlaps  Conflicts  Correct  Incorrect  Emp. Acc.
one    0  [0, 1, 2]       1.0       1.0      0.875        1          0      0.375
two    1  [0, 1, 2]       1.0       1.0      0.875        4          0      0.750
three  2  [0, 1, 2]       1.0       1.0      0.875        3          1      0.500
```

## System info

* OS: GNU/Linux Debian unstable
* Python version: 3.7
* Snorkel version: 0.9
* Versions of any other relevant libraries:

Thanks!
Damien G.

garaud:

Fix the computation of correct/incorrect values from the confusion matrix when the L input data didn't have ABSTAIN label. We only slice the confusion matrix on [1:, 1:] when there is some ABSTAIN label, i.e. -1 values. The full confusion matrix should be taken into account when you don't have -1 values.

## Related issue(s)

Fixes #1446 

## Checklist

Need help on these? Just ask!

* [X] I have added tests to cover my changes.

I've some trouble with pyspark for the test part. But I think it's not related with this fix.

### DIFF
--- a/snorkel/labeling/analysis.py
+++ b/snorkel/labeling/analysis.py
@@ -354,7 +354,12 @@ class LFAnalysis:
         d["Conflicts"] = Series(data=self.lf_conflicts(), index=lf_names)
 
         if Y is not None:
-            confusions = [confusion_matrix(Y, self.L[:, i])[1:, 1:] for i in range(m)]
+            labels = np.unique(
+                np.concatenate((Y.flatten(), self.L.flatten(), np.array([-1])))
+            )
+            confusions = [
+                confusion_matrix(Y, self.L[:, i], labels)[1:, 1:] for i in range(m)
+            ]
             corrects = [np.diagonal(conf).sum() for conf in confusions]
             incorrects = [
                 conf.sum() - correct for conf, correct in zip(confusions, corrects)
--- a/test/labeling/test_analysis.py
+++ b/test/labeling/test_analysis.py
@@ -16,6 +16,15 @@ L = [
 
 Y = [0, 1, 2, 0, 1, 2]
 
+L_wo_abstain = [
+    [3, 3, 3, 5, 4, 3],
+    [3, 4, 5, 5, 4, 3],
+    [5, 3, 4, 4, 5, 3],
+    [4, 4, 5, 4, 3, 3],
+    [3, 4, 3, 5, 4, 3],
+    [5, 3, 3, 4, 4, 3],
+]
+
 
 def f(x):
     return -1
@@ -24,6 +33,7 @@ def f(x):
 class TestAnalysis(unittest.TestCase):
     def setUp(self) -> None:
         self.lfa = LFAnalysis(np.array(L))
+        self.lfa_wo_abstain = LFAnalysis(np.array(L_wo_abstain))
         self.Y = np.array(Y)
 
     def test_label_coverage(self) -> None:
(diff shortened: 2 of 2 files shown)


test/labeling/test_analysis.py
```diff
@@ -24,6 +33,7 @@ def f(x):
 class TestAnalysis(unittest.TestCase):
     def setUp(self) -> None:
         self.lfa = LFAnalysis(np.array(L))
+        self.lfa_bis = LFAnalysis(np.array(L_2))
```
henryre: What does `bis` refer to here?

snorkel/labeling/analysis.py
```diff
@@ -354,7 +354,8 @@ def lf_summary(
         d["Conflicts"] = Series(data=self.lf_conflicts(), index=lf_names)
 
         if Y is not None:
-            confusions = [confusion_matrix(Y, self.L[:, i])[1:, 1:] for i in range(m)]
+            slice_abstain = [slice(int(x), None) for x in ((self.L == -1).sum(axis=0) > 0)]
```
bhancock8: We may be able to simplify this by using the `labels` kwarg in scikit-learn's `confusion_matrix()` method. It let's you specify what the total label space is, so we can just make sure that it always includes -1 before we lop off the abstains for our calculations. Something like this:
```
labels = np.unique(np.concatenate((Y.flatten(), self.L.flatten(), np.array([-1]))))
confusions = [confusion_matrix(Y, self.L[:, i], labels)[1:, 1:] for i in range(m)]
```

garaud: Good point, I'll do that. Thank you for the tips.

test/labeling/test_analysis.py
garaud: Agreed. It's a quite bad name. Sometimes, it means "2nd" in french... I replace this suffix by `_wo_abstain` for "without abstain".

test/labeling/test_analysis.py
```diff
@@ -130,3 +140,18 @@ def test_lf_summary(self) -> None:
     def test_wrong_number_of_lfs(self) -> None:
         with self.assertRaisesRegex(ValueError, "Number of LFs"):
             LFAnalysis(np.array(L), [LabelingFunction(s, f) for s in "ab"])
+
+    def test_lf_summary_without_abstain(self) -> None:
+        df = self.lfa_wo_abstain.lf_summary(self.Y, est_weights=None)
```
henryre: Last nit, I promise: can we add a 3 to L_wo_abstain and a 4 to the Y used here? This will make sure that any future changes account for non-identical label sets between L and Y, which we handle correctly in this change.

garaud: Good idea :)
I updated the test part.

henryre:
@garaud we usually leave it to authors to merge, but we're pushing a patch today so I'm going to go ahead and merge for you. Thanks again, this is great!!

garaud:
> we usually leave it to authors to merge, but we're pushing a patch today so I'm going to go ahead and merge for you.

Don't worry. Thank you for the review.

The weak supervision is quite new for me and snorkel is a great project! We use the weak supervision to label massive 3D point clouds (LIDAR scans in natural environment such as rocks, cliff, vegetation, etc.) and it's really helpful.

PR #1449, opened 2019-09-05
Catch class balance errors and test L matrix edge cases

paroma:
- Add tests for checking LabelModel behavior with limited conflict and overlap in L matrix
- Catch errors in class balance format in LabelModel

## Related issue(s)
None

## Test plan
- Add tests for class balance changes
- Add tests for L matrices with limited overlap or conflict
- Fix existing test to change 0 to -1 for abstains
 
## Checklist

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -552,7 +552,7 @@ class LabelModel(nn.Module):
             )
         if len(self.p) != self.cardinality:
             raise ValueError(
-                f"Y_dev has {len(self.p)} classes. Does not match LabelModel cardinality {self.cardinality}."
+                f"Y_dev has {len(self.p)} class(es). Does not match LabelModel cardinality {self.cardinality}."
             )
 
         self.P = torch.diag(torch.from_numpy(self.p)).float()
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -546,10 +546,14 @@ class LabelModel(nn.Module):
         else:
             self.p = (1 / self.cardinality) * np.ones(self.cardinality)
 
-        if np.any(self.p) == 0:
+        if np.any(self.p == 0):
             raise ValueError(
-                f"No examples for class(es) {np.where(self.p)[0]}."
+                f"Class balance prior is 0 for class(es) {np.where(self.p)[0]}."
             )
+        if len(sorted_counts) != self.cardinality:
+                raise ValueError(
+                    f"Y_dev has {len(sorted_counts)} classes. Does not match LabelModel cardinality {self.cardinality}."
+                )
 
         self.P = torch.diag(torch.from_numpy(self.p)).float()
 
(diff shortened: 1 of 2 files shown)


test/labeling/model/test_label_model.py
```diff
@@ -29,26 +29,50 @@ def _set_up_model(self, L: np.ndarray, class_balance: List[float] = [0.5, 0.5]):
...
+        L = np.array([[-1, 0, 1], [-1, 0, 2], [0, -1, 2], [-1, 0, -1]])
         with self.assertRaisesRegex(ValueError, "L_train has cardinality"):
             label_model.fit(L, n_epochs=1)
 
-        L = np.array([[0], [1], [-1]])
+        L = np.array([[0, 1], [1, 1], [0, 1]])
         with self.assertRaisesRegex(ValueError, "L_train should have at least 3"):
             label_model.fit(L, n_epochs=1)
 
+    def test_mv_default(self):
```
henryre: what change is this testing? do we actually add mv as a default anywhere here?

test/labeling/model/test_label_model.py
```diff
@@ -29,26 +29,50 @@ def _set_up_model(self, L: np.ndarray, class_balance: List[float] = [0.5, 0.5]):
...
         Y_dev = np.array([0, 0, 1, 1, 0, 0, 0, 0, 1, 1])
         label_model._set_class_balance(class_balance=None, Y_dev=Y_dev)
         np.testing.assert_array_almost_equal(label_model.p, np.array([0.6, 0.4]))
 
+        class_balance = np.array([0.0, 1.0])
+        with self.assertRaisesRegex(ValueError, "Class balance prior is 0"):
+            label_model._set_class_balance(class_balance=class_balance, Y_dev=Y_dev)
+
+        Y_dev_one_class = np.array([0, 0, 0])
+        with self.assertRaisesRegex(ValueError, "Y_dev has"):
```
henryre: nit: change substring to `"Does not match LabelModel cardinality"`

test/labeling/model/test_label_model.py
paroma: it's checking for degenerate L matrices (aka low overlap and no conflicts) that the label model should internally default to the predictions MV would assign. it doesn't explicitly call majority vote anywhere

snorkel/labeling/model/label_model.py
```diff
@@ -545,6 +545,16 @@ def _set_class_balance(
...
         else:
             self.p = (1 / self.cardinality) * np.ones(self.cardinality)
+
+        if np.any(self.p == 0):
+            raise ValueError(
+                f"Class balance prior is 0 for class(es) {np.where(self.p)[0]}."
+            )
+        if len(self.p) != self.cardinality:
+            raise ValueError(
+                f"Y_dev has {len(self.p)} class(es). Does not match LabelModel cardinality {self.cardinality}."
```
henryre: this could also be because `class_balance` is the wrong size

henryre: this error is also confusing for the Y_dev case. what's happening is that we're saying some of the class priors are 0 (by virtue of not being in Y_dev) which is the same as the above ValueError. rec: we raise a separate message under each of the if/else block (under `if class_balance is not None:` if it's the wrong size, then something like`Class balance prior is 0` in the other two cases.

PR #1450, opened 2019-09-05
Ignore abstains in Scorer, change LabelModel default tie break policy

paroma:
- Change Scorer default to ignore abstains in preds
- Change LabelModel tie break policy default to abstain (instead of random)
- Log warning when calling LabelModel `score()` function

## Test plan
- Add tests in LabelModel for `predict()` and `score()` functions related to abstain default
- Add test for Scorer to check abstains ignored in preds by default

## Checklist

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -477,8 +477,10 @@ class LabelModel(nn.Module):
         >>> label_model.score(L, Y=np.array([1, 1, 1]), metrics=["f1"])
         {'f1': 0.8}
         """
-        if tie_break_policy == 'abstain':
-            logging.warning(f"Metrics calculated over datapoints with non-abstain labels only")
+        if tie_break_policy == "abstain":
+            logging.warning(
+                f"Metrics calculated over datapoints with non-abstain labels only"
+            )
 
         Y_pred, Y_prob = self.predict(
             L, return_probs=True, tie_break_policy=tie_break_policy
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -455,7 +455,7 @@ class LabelModel(nn.Module):
         L
             An [n,m] matrix with values in {-1,0,1,...,k-1}
         Y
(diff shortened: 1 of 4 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -472,6 +472,11 @@ def score(
         >>> label_model.score(L, Y=np.array([1, 1, 1]), metrics=["f1"])
         {'f1': 0.8}
         """
+        if tie_break_policy == "abstain":  # pragma: no cover
+            logging.warning(
+                "Metrics calculated over datapoints with non-abstain labels only"
```
bhancock8: Nit: we've been using data points (2 words)

test/analysis/test_scorer.py
```diff
@@ -68,12 +68,19 @@ def test_abstain_labels(self) -> None:
...
+        # Test abstain=-1 for gold
         scorer = Scorer(metrics=["accuracy"], abstain_label=-1)
         results = scorer.score(golds, preds, probs)
         results_expected = dict(accuracy=0.75)
         self.assertEqual(results, results_expected)
 
+        # Test abstain=-1 for preds and gold
+        abstain_preds = np.array([-1, -1, 1, 1, 0])
+        abstain_probs = np.array([0.5, 0.5, 0.9, 0.7, 0.4])
+        results = scorer.score(golds, abstain_preds, abstain_probs)
```
bhancock8: No need to pass in probs here. They're optional, and you're only calculating accuracy in this scorer, which just requires golds and preds.

test/labeling/model/test_label_model.py
```diff
@@ -209,6 +209,13 @@ def test_predict_proba(self):
         np.testing.assert_array_almost_equal(probs, true_probs)
 
     def test_predict(self):
+        L = np.array([[-1, 1, 0], [0, -1, 1], [1, 0, -1]])
```
bhancock8: Can we add a simple comment here that this test is confirming that with 3 LFs that counteract one another results in tie votes and therefore abstains on all points?

PR #1457, opened 2019-09-11
Ray-based Tuner for LabelModel

paroma:

- Add LabelModelTuner that uses [Ray Tune](https://ray.readthedocs.io/en/latest/tune.html) to search over specified LabelModel numerical hyper parameters
     - Add ray to requirements

## Test plan
- Add tests for random search and grid search
- Add complex test with Ray Hyperband scheduler with random search

## Checklist

Need help on these? Just ask!


ajratner:
@paroma from quick skim: LGTM!!  But I wonder: would it be worth thinking about some heuristics we might consider for when `Y_dev` is not available?  Maybe something for a separate commit, and just get this one in first, but can briefly discuss tomorrow?

snorkel/labeling/__init__.py
```diff
@@ -10,4 +10,5 @@
     RandomVoter,
 )
 from .model.label_model import LabelModel  # noqa: F401
+from .model.tuner import LabelModelTuner  # noqa: F401
```
henryre: need to remove to keep `ray` off the main import path. can revisit later, but doesn't look essential right now.

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        Gold labels for dev set for tuning hyperparameters
+    metric
+        Metric to maximize while hyperparameter tuning
+    fit_config
+        Arguments for changing LabelModel fit() config defaults
+    """
+
+    def __init__(
+        self,
+        label_model,
```
henryre: type plz!

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        L_train: np.ndarray,
+        L_dev: np.ndarray,
+        Y_dev: np.ndarray,
+        metric: str = "accuracy",
+        fit_config: Optional[Dict[str, Any]] = {},
```
henryre: No mutable defaults, just switch to `None`

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+    ) -> None:
+
+        self.label_model = label_model
+        self.L_train = L_train
+        self.L_dev = L_dev
+        self.Y_dev = Y_dev
+        self.metric = metric
+        self.fit_config = fit_config
+
+    def _trainable(self):  # type: ignore
```
henryre: `def _trainable(self) -> "LabelModelTrainable":` or `def _trainable(self) -> Trainable:`

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        """Return a Trainable class definition for Ray tuner with some user-defined parameters from LabelModelTuner class."""
+
+        class LabelModelTrainable(Trainable):
+            """Trainable class for LabelModel for Ray tuner."""
+
+            def _setup(self, config: Optional[Dict[str, Any]]):
```
henryre: return types for all class methods

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+                self.tuner = LabelModelTrainable.tuner  # type: ignore
```
henryre: I'm guessing this is some ray pattern, but i don't see why this is necessary

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+
+            def _save(self, checkpoint_dir: str):
+                path = os.path.join(checkpoint_dir, "checkpoint")
+                with open(path, "w") as f:
+                    f.write(json.dumps({"timestep": self.timestep}))
+                return path
+
+            def _restore(self, checkpoint_path: str):
+                with open(checkpoint_path) as f:
+                    self.timestep = json.loads(f.read())["timestep"]
```
henryre: `json.load(f)`

henryre: what about the config?

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        return LabelModelTrainable
+
+    def run_tuner(
+        self,
+        config: Dict[str, Any],
+        scheduler: Optional[Any] = None,
+        name: Optional[str] = "label_model_tuner",
+        num_samples: Optional[int] = 5,
+        verbose: Optional[bool] = True,
+        experiment_configs: Optional[Dict[str, Any]] = {
```
henryre: no mutable defaults

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+            "stop": {"training_iteration": 100}
+        },
+        run_configs: Optional[Dict[str, Any]] = {
```
henryre: no mutable defaults

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+
+            def _restore(self, checkpoint_path: str):
+                with open(checkpoint_path) as f:
+                    self.timestep = json.loads(f.read())["timestep"]
+
+        # returning a LabelModelTrainable class definition with tuner as user-defined LabelModelTuner instance
+        LabelModelTrainable.tuner = self
+        return LabelModelTrainable
+
+    def run_tuner(
```
henryre: `tune`?

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        self,
+        config: Dict[str, Any],
+        scheduler: Optional[Any] = None,
+        name: Optional[str] = "label_model_tuner",
+        num_samples: Optional[int] = 5,
+        verbose: Optional[bool] = True,
```
henryre: not used

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        self.L_train = L_train
+        self.L_dev = L_dev
+        self.Y_dev = Y_dev
+        self.metric = metric
+        self.fit_config = fit_config
+
+    def _trainable(self):  # type: ignore
+        """Return a Trainable class definition for Ray tuner with some user-defined parameters from LabelModelTuner class."""
+
+        class LabelModelTrainable(Trainable):
```
henryre: why do we need this to be defined here?

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+            config=config,
+            **experiment_configs,
+        )
+
+        analysis = ray.tune.run(exp, scheduler=scheduler, **run_configs)
+        self.best_config = analysis.get_best_config(metric=self.metric)
+        self.label_model.fit(
+            self.L_train,
+            **self.fit_config,
+            **analysis.get_best_config(metric=self.metric),
```
henryre: `**self.best_config`

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+    """A hyperparameter tuner for the LabelModel using Ray Tune (https://ray.readthedocs.io/en/latest/tune.html).
+
+    Example
+    -------
+    >>> L_train = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1]])
+    >>> L_dev = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1]])
+    >>> Y_dev = np.array([1, 1, 0, 1])
+    >>> label_model = LabelModel(cardinality=2)
+    >>> label_model_tuner = LabelModelTuner(label_model, L_train, L_dev, Y_dev)
+    >>> label_model_tuner = LabelModelTuner(label_model, L_train, L_dev, Y_dev, fit_config={"class_balance" : np.array([0.5, 0.5])})
```
henryre: nit: line is too long

snorkel/labeling/model/tuner.py
```diff
@@ -0,0 +1,157 @@
...
+        Parameters
+        ----------
+        config
+            Dictionary of hyperparameters for LabelModel fit() to search over
+        scheduler
+            Special scheduler for hyperparameter search trials
+        name
+            Name of experiment
+        num_samples
+            Number of trials to run. Set to 1 for gridsearch
```
henryre: From my reading of this, 1 would mean pick one random set of hyperparams and just run that (i.e. not a search). is it a flag value?

snorkel/labeling/model/tuner.py
paroma: Ray requires a Trainable with specific forms for the _train(), _setup() methods, but doesn't have a clean way of taking in additional args for those methods. This links to the overall LabelModelTuner class when Trainable is defined so it has access to the args that are constant across all runs and required across fit/score etc.

snorkel/labeling/model/tuner.py
paroma: This is from the hyberband example [here](https://github.com/ray-project/ray/blob/9ce6dd9b883594cec4ce42bd55962d53088afabe/python/ray/tune/examples/hyperband_example.py#L38) 

On second read, the _setup function there only has timestep while I have others, will fix here

snorkel/labeling/model/tuner.py
paroma: https://github.com/snorkel-team/snorkel/pull/1457#discussion_r323348959

it refers to the LabelModelTuner class that it's defined in. Ray requires a class object when the tuner is called, and this returns a custom class object every time

snorkel/labeling/model/tuner.py
paroma: There's no flag value iirc, but with grid search, one "sample" goes through all the possible combinations by default.

I can change default for num_smaples to be 1 and then set it explicitly for the rest

paroma:
@ajratner agreed - we can have something without Y_dev that depends on looking at the learned weights or patterns in assigned labels. should be easy to edit the existing LabelModelTuner class to check if Y_dev exists!

paroma:
Closing in favor of #1461, discussion related to hyper parameter search without Y_dev still applies

PR #1461, opened 2019-09-12
LabelModel wrapper for scikit-learn hyper parameter search

paroma:
Create LabelModelWrapper to match scikit-learn signature. Currently functionality is to allow hyper parameter search functions from scikit-learn

## Related issue(s)
None

## Test plan
- Add test for wrapper via scikit-learn GridSearch

## Checklist


snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+
+import numpy as np
+from sklearn.base import BaseEstimator, ClassifierMixin  # type: ignore
+
+from snorkel.labeling import LabelModel
+from snorkel.utils.lr_schedulers import LRSchedulerConfig
+from snorkel.utils.optimizers import OptimizerConfig
+
+
+class SKlearnLabelModel(BaseEstimator, ClassifierMixin):
```
henryre: SklearnLabelModel

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
+from typing import Any, List, Optional
+
+import numpy as np
+from sklearn.base import BaseEstimator, ClassifierMixin  # type: ignore
```
henryre: what's going on here?

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        self.verbose = verbose
+        self.device = device
+        self.metric = metric
+        self.tie_break_policy = tie_break_policy
+        self.n_epochs = n_epochs
+        self.lr = lr
+        self.l2 = l2
+        self.optimizer = optimizer
+        self.optimizer_config = (
+            OptimizerConfig()  # type: ignore
```
henryre: nit: `optimizer_config or OptimizerConfig()`

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+
+    Examples
+    --------
+    >>> L_train = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1], [0, 0, 0, 1]])
+    >>> L_dev = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1]])
+    >>> Y_dev = np.array([1, 1, 0, 1])
+
+    >>> from sklearn.model_selection import GridSearchCV
+    >>> param_grid = [{"lr": [0.01, 0.0000001], "l2": [0.001, 0.5], "metric": ["accuracy"]}]
+    >>> clf = GridSearchCV(SKlearnLabelModel(L_train), param_grid, cv=[(slice(None), slice(None))])
```
henryre: `(slice(None), slice(None))` and passing L_train to init is a code smell, but grrrrr all of sklearn's model selection utils are all tied to CV, not generic validation set fitting. did you look into `PredefinedSplit`? we could provide a utility that glues L_train and L_dev together, makes an appropriate Y, then produces the corresponding `PredefinedSplit` object for people to use with their *CV calls.

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        self.log_freq = log_freq
+        self.mu_eps = mu_eps
+        self.class_balance = class_balance
+
+        self.label_model = LabelModel(
+            cardinality=self.cardinality, verbose=self.verbose, device=self.device
+        )
+
+    def fit(
+        self, L_dev: np.ndarray, Y_dev: Optional[np.ndarray] = None
```
henryre: wait but L_dev isn't used at all...

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+        Y_dev
+             Gold labels associated with data points in L_dev
+
+        Returns
+        -------
+        SKlearnLabelModel
+        """
+        self.label_model.fit(
+            L_train=self.L_train,
+            Y_dev=Y_dev,
```
henryre: do we need to provide Y_dev here to use CV? i think no, just for print outs

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+            cardinality=self.cardinality, verbose=self.verbose, device=self.device
+        )
+
+    def fit(
+        self, L_dev: np.ndarray, Y_dev: Optional[np.ndarray] = None
+    ) -> "SKlearnLabelModel":
+        """
+        Train label model.
+
+        **Note that one of the inputs is L_dev but it is not used to train the LabelModel. This is to make the wrapper compatible with scikit-learn hyperparameter search functionality.
```
henryre: ah

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,191 @@
...
+            lr_scheduler_config=self.lr_scheduler_config,
+            prec_init=self.prec_init,
+            seed=self.seed,
+            log_freq=self.log_freq,
+            mu_eps=self.mu_eps,
+        )
+
+        return self
+
+    def score(self, L_dev: np.ndarray, Y_dev: np.ndarray = None) -> float:
```
henryre: why default to None...?

snorkel/labeling/model/wrapper.py
paroma: We need methods like get and set params  from these scikit learn base estimators

snorkel/labeling/model/wrapper.py
paroma: Yeah this is tied to the cv splits. Gridsearch expects X,Y to be the same for train and score, and does CV internally to figure out what to train on and what to score on. Might be fixed with predefined split

snorkel/labeling/model/wrapper.py
henryre: I meant why is mypy upset?

snorkel/labeling/model/wrapper.py
paroma: We need Y_dev because that’s what gridsearch expects, without it, it thinks it’s a unsupervised learning alg

snorkel/labeling/model/wrapper.py
paroma: Note to self: https://scikit-learn.org/stable/modules/generated/sklearn.model_selection.PredefinedSplit.html

Setting all of L_train’s indices to -1 in test split should fix the issue. Gridsearch complains when number of splits is less than 1, but will test wrapping in PredefinedSplit

snorkel/labeling/model/wrapper.py
paroma: `snorkel/labeling/model/wrapper.py:4: error: No library stub file for module 'sklearn.base'`

henryre: ah ok needs to be added to setup.cfg a la https://github.com/snorkel-team/snorkel/blob/master/setup.cfg#L92

snorkel/labeling/utils.py
```diff
@@ -27,3 +28,45 @@ def filter_unlabeled_dataframe(
     """
     mask = (L != -1).any(axis=1)
     return X.iloc[mask], y[mask]
+
+
+def create_cv_data(
```
henryre: would recommend keeping in wrapper so that there's a single, canonical import statement (i.e. everyone thinks to always import this when importing the wrapper

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,178 @@
...
+        metric: str = "accuracy",
+        tie_break_policy: str = "abstain",
+        n_epochs: int = 100,
+        lr: float = 0.01,
+        l2: float = 0.0,
+        optimizer: str = "sgd",
+        optimizer_config: Optional[OptimizerConfig] = None,
+        lr_scheduler: str = "constant",
+        lr_scheduler_config: Optional[LRSchedulerConfig] = None,
+        prec_init: float = 0.7,
```
henryre: seems random

snorkel/labeling/utils.py
```diff
@@ -27,3 +28,45 @@ def filter_unlabeled_dataframe(
...
+    cv_split
+        PredefinedSplit object with two splits: L_train in train and *_dev in test
+    """
+
+    if np.shape(L_dev)[0] != np.shape(Y_dev)[0]:
+        raise ValueError("Num. datapoints in Y_dev and L_dev do not match.")
+
+    # combine train and dev L and Y
+    L_all = np.append(L_train, L_dev, axis=0)
+    Y_all = np.append(-1 * np.ones(np.shape(L_train)[0]), Y_dev)
```
henryre: guessing `np.ones` creates a float array

snorkel/labeling/utils.py
```diff
@@ -27,3 +28,45 @@ def filter_unlabeled_dataframe(
...
+
+    # create cv split array with two splits
+    test_fold = np.append(
+        -1 * np.ones(np.shape(L_train)[0]), np.ones(np.shape(L_dev)[0])
```
henryre: let's do `n_train = np.shape(L_train)[0]` and `n_dev = np.shape(L_dev)[0]` at the beginning since we use so much

snorkel/labeling/utils.py
henryre: a little torn on what to name this...we aren't using CV but we are sticking it in a CV routine...maybe just clarify in doc string.

test/labeling/test_utils.py
```diff
@@ -18,3 +18,16 @@ def test_filter_unlabeled_dataframe(self) -> None:
         np.testing.assert_array_almost_equal(
             y_filtered, np.array([[0.25, 0.25, 0.25, 0.25], [0.2, 0.3, 0.5, 0.0]])
         )
+
+    def test_create_cv_data(self) -> None:
+        L_train = np.array([[1, 1, 0], [-1, -1, 1], [0, 0, 1], [1, 1, 0]])
+        L_dev = np.array([[-1, 1, 1], [0, -1, 0]])
+        Y_dev = np.array([-1, 1])
```
henryre: is -1 a valid value here?

test/labeling/test_utils.py
```diff
@@ -18,3 +18,16 @@ def test_filter_unlabeled_dataframe(self) -> None:
...
+
+        L, Y, cv_split = create_cv_data(L_train, L_dev, Y_dev)
+        self.assertEqual(np.shape(L)[0], np.shape(L_train)[0] + np.shape(L_dev)[0])
```
henryre: we know exactly what L and Y should be. should test that, not shapes

test/labeling/model/test_wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+    def test_grid_search(self):
+        L, Y, cv_split = self._create_splits()
+        param_grid = [
+            {"lr": [0.01, 0.0000001], "l2": [0.001, 0.5], "metric": ["accuracy"]}
+        ]
+
+        clf = GridSearchCV(SklearnLabelModel(), param_grid, cv=cv_split)
+        clf.fit(L, Y)
+
+        self.assertGreaterEqual(clf.best_score_, 0.75)
```
henryre: this isn't what we're try to test here. we should test the properties of the `clf` object (e.g. the `best_estimator_`, the `best_param_set_` or whatever.

test/labeling/model/test_wrapper.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def test_random_search(self):
```
henryre: this probably isn't necessary, it'll just test sklearn param grid creation utilities

snorkel/labeling/model/wrapper.py
paroma: default from [LabelModel](https://github.com/snorkel-team/snorkel/blob/9af1c77cb68a31258a476e234a67d3f5e9587f53/snorkel/labeling/model/label_model.py#L61)

test/labeling/test_utils.py
paroma: and this doesn't error because -1s are ignored in gold labels with the Scorer() object!

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,226 @@
...
+        Returns
+        -------
+        float
+            Score for metric speficied in self.metric
+        """
+
+        results = self.label_model.score(L, Y, [self.metric], self.tie_break_policy)
+        return results[self.metric]
+
+    def create_param_search_data(
```
henryre: don't think this should be an instance method...doesn't use `self`. maybe a class method? but probably better as a top-level function

test/labeling/model/test_wrapper.py
```diff
@@ -0,0 +1,54 @@
...
+        wrapper = SklearnLabelModel()
+        L, Y, cv_split = wrapper.create_param_search_data(L_train, L_dev, Y_dev)
+
+        param_grid = [{"n_epochs": [5, 100], "lr": [1e-10, 1e-2], "metric": ["f1"]}]
+
+        clf = GridSearchCV(wrapper, param_grid, cv=cv_split)
+        clf.fit(L, Y)
+
+        self.assertEqual(
+            clf.best_params_, {"lr": 0.01, "metric": "f1", "n_epochs": 100}
```
henryre: cool so we expect this since all other settings should be drastically underfit?

henryre: might want to set n_epochs to [1, 10] just to be safer? e.g. if there are changes to label model behavior?

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,226 @@
...
+    Examples
+    --------
+    >>> L_train = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1], [0, 0, 0, 1]])
+    >>> L_dev = np.array([[1, 1, 1, 0], [0, 1, 1, 0], [1, 0, 0, 1], [0, 1, 0, 1]])
+    >>> Y_dev = np.array([1, 1, 0, 1])
+
+    >>> from sklearn.model_selection import GridSearchCV
+    >>> param_grid = [{"lr": [0.01, 0.0000001],
+    ...    "l2": [0.001, 0.5], "metric": ["accuracy"]}]
+    >>> wrapper = SklearnLabelModel()
```
henryre: nit: would call this `label_model`

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,226 @@
...
+    >>> clf = GridSearchCV(wrapper, param_grid, cv=cv_split)
+    >>> clf.fit(L, Y)  # doctest: +SKIP
+    >>> clf.best_score_  # doctest: +SKIP
+
+    >>> from sklearn.model_selection import RandomizedSearchCV
+    >>> param_dist = {"lr": np.random.uniform(low=1e-7, high=1e-2, size=(50,)),
+    ...     "l2": np.random.uniform(low=1e-2, high=0.5, size=(50,))}
+    >>> random_search = RandomizedSearchCV(wrapper, param_distributions=param_dist, n_iter=4, cv=cv_split, iid=False)
+    >>> random_search.fit(L, Y)  # doctest: +SKIP
+    >>> random_search.best_score_  # doctest: +SKIP
```
henryre: let's print the output of this right?

snorkel/labeling/model/wrapper.py
```diff
@@ -0,0 +1,226 @@
...
+    >>> Y_dev = np.array([1, 1, 0, 1])
+
+    >>> from sklearn.model_selection import GridSearchCV
+    >>> param_grid = [{"lr": [0.01, 0.0000001],
+    ...    "l2": [0.001, 0.5], "metric": ["accuracy"]}]
+    >>> wrapper = SklearnLabelModel()
+    >>> L, Y, cv_split = wrapper.create_param_search_data(L_train, L_dev, Y_dev)
+    >>> clf = GridSearchCV(wrapper, param_grid, cv=cv_split)
+    >>> clf.fit(L, Y)  # doctest: +SKIP
+    >>> clf.best_score_  # doctest: +SKIP
```
henryre: also print output

test/labeling/model/test_wrapper.py
paroma: rest do underfit. and its grid search, so it'll either pick 5 epochs or 100 epochs (and the former would be wrong)

snorkel/labeling/model/wrapper.py
paroma: using `np.around` since this is random search and best score can be 0.75 or 0.1

test/labeling/model/test_wrapper.py
paroma: it's grid searching over just the 4 options, so it should underfit at 5 and work at 100

snorkel/labeling/model/wrapper.py
paroma: made it static method - didn't need cls. mainly to avoid another import statement and also to package the wrapper and this method together. can revisit if it's too weird

PR #1470, opened 2019-09-25
Dependency Aware Label Model

brahmaneya:
Add Dependency Aware Label Model to contrib. Made some internal changes in main label model that don't change any external behavior but make the subclassing for the dependency aware case easier.
The main change is the way we compute augmented L (now includes one-hot terms for cliques of dependent LFs) and O (which is still L^{T}L). We can't compute O_inverse explicitly when LFs are unipolar, because O becomes singular. So we use a loss function on (the augmented) O that forces out factorization to approximately invert it.

## Test plan
Added new tests. Modified some old tests to be compatible with changes.

## Checklist

Need help on these? Just ask!


snorkel/contrib/deps/labeling/model/label_model.py
```diff
@@ -0,0 +1,180 @@
...
+        Examples
+        --------
+        >>> L = np.array([[0, 0, -1], [-1, 0, 1], [1, -1, 0]])
+        >>> Y_dev = [0, 1, 0]
+        >>> label_model = DependencyAwareLabelModel(verbose=False)
+        >>> label_model.fit_with_deps(L, deps=[(0, 2)])  # doctest: +SKIP
+        >>> label_model.fit_with_deps(L, deps=[(0, 2)], Y_dev=Y_dev)  # doctest: +SKIP
+        >>> label_model.fit_with_deps(L, deps=[(0, 2)], class_balance=[0.7, 0.3])  # doctest: +SKIP
+        """
+        # Set random seed
```
paroma: if `deps` passed in is empty, should it default to calling the original LabelModel fit function?

snorkel/contrib/deps/labeling/model/label_model.py
```diff
@@ -0,0 +1,180 @@
...
+
+        L_shift = L_train + 1  # convert to {0, 1, ..., k}
+        if L_shift.max() > self.cardinality:
+            raise ValueError(
+                f"L_train has cardinality {L_shift.max()}, cardinality={self.cardinality} passed in."
+            )
+
+        self._set_constants(L_shift)
+        self._set_class_balance(class_balance, Y_dev)
+        self._set_dependencies(deps or [])
```
paroma: can we default `deps = [ ]` in the function itself instead of `None`?

snorkel/contrib/deps/labeling/model/label_model.py
```diff
@@ -0,0 +1,180 @@
...
+            logging.info("Computing O...")
+        self._generate_O(L_shift)
+        self._init_params()
+
+        # Set model to train mode
+        self.train()
+
+        # Move model to GPU
+        if self.config.verbose and self.config.device != "cpu":  # pragma: no cover
+            logging.info("Using GPU...")
```
paroma: It's worth checking on GPU - there were some parameters that had to be cast explicitly for the independent LabelModel

snorkel/labeling/model/label_model.py
```diff
@@ -881,7 +913,7 @@ def fit(
 
         self._set_constants(L_shift)
         self._set_class_balance(class_balance, Y_dev)
-        self._create_tree()
+        self._set_dependencies([])
```
paroma: can we default `deps = [ ]` in the `_set_dependencies` function? or pass in `(`deps = [ ])`
and nit: can we call it `_set_structure()` or something that's more general than just dependencies?

test/labeling/model/test_label_model.py
```diff
@@ -117,7 +117,8 @@ def test_generate_O(self):
 
         # Higher order returns same matrix (num source = num cliques)
         # Need to test c_tree form
-        label_model._generate_O(L + 1, higher_order=True)
+        label_model.higher_order = True
```
paroma: we shouldn't have `higher_order = True` tests in the independent label model version, just to maintain the difference

test/labeling/model/test_label_model.py
```diff
@@ -131,8 +132,9 @@ def test_augmented_L_construction(self):
         L_shift = L + 1
         lm = LabelModel(cardinality=k, verbose=False)
         lm._set_constants(L_shift)
-        lm._create_tree()
-        L_aug = lm._get_augmented_label_matrix(L_shift, higher_order=True)
+        lm._set_dependencies([])
+        lm.higher_order = True
```
paroma: same as above

snorkel/contrib/deps/test/labeling/model/test_label_model.py
```diff
@@ -0,0 +1,242 @@
...
+                [0, 0, 1, 1, 0, 0, 0, 0, 0, 0],
+                [0, 0, 1, 1, 0, 0, 0, 0, 0, 0],
+            ]
+        )
+
+        mask = label_model.mask.numpy()
+        np.testing.assert_array_equal(mask, true_mask)
+
+
+@pytest.mark.complex
```
paroma: can we add:
* a test that explicitly checks the conversion from low rank to mu for the dependent case
* a test for multi-class (can be complex)
* a test comparing the `weights` explicitly to the recovered weights (like [here](https://github.com/snorkel-team/snorkel/blob/77e36c644fb2a93467514230c72fa776d9229ab5/test/labeling/model/test_label_model.py#L168))

test/labeling/model/test_label_model.py
paroma: also there should be a warning/error when `higher order=True` but the independent LabelModel is being called

snorkel/contrib/deps/labeling/model/label_model.py
```diff
@@ -0,0 +1,180 @@
...
+import scipy as sp
+import torch
+
+from snorkel.labeling.analysis import LFAnalysis
+from snorkel.labeling.model.label_model import LabelModel, TrainConfig
+from snorkel.utils.config_utils import merge_config
+
+
+class DependencyAwareLabelModel(LabelModel):
+    """A LabelModel that handles dependencies and learn associated weights to assign training labels."""
```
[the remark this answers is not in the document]
brahmaneya: Can you send me the links (or paste them here?)

test/labeling/model/test_label_model.py
brahmaneya: The label model already has some higher order code (it had some even before this PR, but now it has more). e.g. The function to create l_aug. We should have some tests to improve coverage for that right?

snorkel/contrib/deps/labeling/model/label_model.py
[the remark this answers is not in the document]
paroma: [Learning Dependency Structures for Weak Supervision Models](https://arxiv.org/pdf/1903.05844.pdf) and [AAAI'19](https://arxiv.org/pdf/1810.02840.pdf)

test/labeling/model/test_label_model.py
paroma: got it, this can stay as is then

snorkel/contrib/deps/test/labeling/model/test_label_model.py
paroma: discussion IRL about checking conditional probabilities instead of weights

snorkel/contrib/deps/labeling/model/label_model.py
brahmaneya: We want this to eventually replace LabelModel right? So it'd be good if the current code can just be copy pasted in place of LabelModel.fit. Adding a call to LabelModel.fit will make it harder.

snorkel/contrib/deps/labeling/model/label_model.py
brahmaneya: You mean testing? How do we do that?

snorkel/contrib/deps/test/labeling/model/test_label_model.py
brahmaneya: Added a multiclass test.

snorkel/contrib/deps/labeling/model/label_model.py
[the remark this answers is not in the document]
brahmaneya: Added

snorkel/contrib/deps/labeling/model/label_model.py
paroma: I manually changed device to “gpu” in the tests and ran it on a machine with GPUs

PR #1473, opened 2019-09-27
Combine Dependency Learner and LabelModel 

paroma:
Call DependencyLearner from DependencyAwareLabelModel
Add scikit-learn wrapper around DependencyAwareLabelModel for hyperparameter search

## Test plan
Add integration tests for Learner and LabelModel
Add hyperparameter search tests for checking expected performance

## Checklist

### DIFF
--- a/snorkel/contrib/deps/labeling/model/label_model.py
+++ b/snorkel/contrib/deps/labeling/model/label_model.py
@@ -8,6 +8,7 @@ import numpy as np
 import scipy as sp
 import torch
 
+from snorkel.contrib.deps.labeling.model.dep_learner import DependencyLearner
 from snorkel.labeling.analysis import LFAnalysis
 from snorkel.labeling.model.label_model import LabelModel, TrainConfig
 from snorkel.utils.config_utils import merge_config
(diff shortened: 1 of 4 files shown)


snorkel/contrib/deps/labeling/model/label_model.py
```diff
@@ -21,7 +22,9 @@ def _loss_inv_mu(self, l2: float = 0) -> torch.Tensor:
         loss_2 = torch.norm(torch.sum(self.mu @ self.P, 1) - torch.diag(self.O)) ** 2
         return loss_1 + loss_2 + self.loss_l2(l2=l2)
 
-    def _robust_pca_Q(self, L: np.ndarray) -> np.ndarray:
+    def _robust_pca_Q(
+        self, L: np.ndarray, gamma: Optional[float] = 1e-8, lam: Optional[float] = 0.1
```
brahmaneya: Is 1/sqrt(M) a better default for this? If so, you can default if to None here and set it to 1/sqrt(M) if None below.

paroma: In synthetics, actually 0.1 worked better. 1/sqrt(M) breaks when M is very small

I could do min (0.1, 1/sqrt(M)) if that works with the None option as suggested

PR #1480, opened 2019-10-02
Add fault tolerance to appliers

henryre:

Adds fault tolerance to appliers. Allows LFs/SFs to fail (abstaining if they do) and records the number of failures for each LF/SF.

**Note:** we don't track failure metadata for Dask or PySpark in this PR, since merging across processes can be tricky. We can add later if requested.

## Test plan

Added new unit tests for core, Pandas, Dask, and PySpark appliers. Ran `tox -e spark` and `tox -e complex`

### DIFF
--- a/snorkel/labeling/apply/core.py
+++ b/snorkel/labeling/apply/core.py
@@ -1,5 +1,5 @@
 from itertools import chain
-from typing import List, Tuple, Union
+from typing import DefaultDict, Dict, List, NamedTuple, Tuple, Union
 
 import numpy as np
 from tqdm import tqdm
@@ -9,6 +9,28 @@ from snorkel.types import DataPoint, DataPoints
 from snorkel.utils.data_operators import check_unique_names
 
 RowData = List[Tuple[int, int, int]]
+
+
+class ApplierMetadata(NamedTuple):
+    """Metadata about Applier call."""
+
+    # Map from LF name to number of faults in apply call
+    faults: Dict[str, int]
+
+
+class _FunctionCaller:
+    def __init__(self, fault_tolerant: bool):
+        self.fault_tolerant = fault_tolerant
+        self.fault_counts: DefaultDict[str, int] = DefaultDict(int)
+
+    def __call__(self, f: LabelingFunction, x: DataPoint) -> int:
+        if not self.fault_tolerant:
+            return f(x)
+        try:
+            return f(x)
+        except Exception:
+            self.fault_counts[f.name] += 1
+            return -1
 
 
 class BaseLFApplier:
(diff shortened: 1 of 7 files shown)


snorkel/labeling/apply/core.py
```diff
@@ -11,6 +11,27 @@
...
+
+    faults: Dict[str, int]
+
+
+class _FunctionCaller:
+    def __init__(self, fault_tolerant: bool):
+        self.fault_tolerant = fault_tolerant
+        self.fault_counts: DefaultDict[str, int] = DefaultDict(int)
+
+    def __call__(self, f: LabelingFunction, x: DataPoint) -> int:
```
vincentschen: do we want support for `f(*args, **kwargs)`, in case folks want to pass preprocessors/resources?

snorkel/labeling/apply/core.py
```diff
@@ -11,6 +11,27 @@
 RowData = List[Tuple[int, int, int]]
 
 
+class ApplierMetadata(NamedTuple):
+    """Metadata about Applier call."""
+
+    faults: Dict[str, int]
```
vincentschen: comment here — what is `str` -> `int`

snorkel/labeling/apply/dask.py
```diff
@@ -20,7 +20,9 @@ class DaskLFApplier(BaseLFApplier):
     For more information, see https://docs.dask.org/en/stable/dataframe.html
     """
 
-    def apply(self, df: dd, scheduler: Scheduler = "processes") -> np.ndarray:
+    def apply(
+        self, df: dd, scheduler: Scheduler = "processes", fault_tolerant: bool = False
```
vincentschen: nit: why are we still naming the args `df`?

test/labeling/apply/test_spark.py
```diff
@@ -53,6 +59,19 @@ def test_lf_applier_spark(self) -> None:
...
 
+    @pytest.mark.complex
+    @pytest.mark.spark
+    def test_lf_applier_spark_fault(self) -> None:
+        sc = SparkContext.getOrCreate()
+        sql = SQLContext(sc)
+        df = pd.DataFrame(dict(num=DATA))
+        rdd = sql.createDataFrame(df).rdd
+        applier = SparkLFApplier([f, f_bad])
+        with self.assertRaises(Exception):
```
vincentschen: can we be more specific about the exception?

snorkel/labeling/apply/core.py
henryre: can abstract later if needed, but let's leave for now

test/labeling/apply/test_spark.py
henryre: tried but there's a race condition for which exception gets thrown first

snorkel/labeling/apply/core.py
vincentschen: this seems like a pretty general case that we'll need to address soon, agree that it can be a different PR

PR #1488, opened 2019-10-10
Capped-iteration LabelModel permutation sampling.

### LINKED ISSUE #1486: Linear increase in labels leads to exponential increase in LabelModel fit time.
**First things first:** thank you all for all the hard work going into this library! It's fantastic and incredibly accessible, especially with the numerous tutorials. Onto the issue:

---

## Issue description

When attempting to use `snorkel.labeling.LabelModel` for more than ~8 classes, the call to `_break_col_permutation_symmetry` (more specifically, `permutations(range(k))` where `k` is the number of classes) increases runtime at an exponential rate.

## Code example/repro steps

I've isolated a minimal-ish example below (forgive the improper use of globals):

```python
from typing import Dict, List, Optional, Tuple
import time

from sklearn.datasets import make_classification
from snorkel.labeling import LabelingFunction, LabelModel, PandasLFApplier
import numpy as np
import pandas as pd


# Constants.
N_CLASSES: Optional[int] = None
ABSTAIN: int = -1


def generate_data() -> Tuple[pd.DataFrame, np.ndarray]:

    # Generate dummy dataset.
    X, y = make_classification(
        n_samples=50_000, 
        n_features=N_CLASSES * 2,
        n_classes=N_CLASSES,
        n_informative=N_CLASSES,
        n_clusters_per_class=1,
        random_state=2019,
    )

    # Assert we properly created our dataset.
    assert X.shape[0] == y.shape[0]
    assert len(set(y)) == N_CLASSES

    # Cast our array to DataFrame.
    X_df: pd.DataFrame = pd.DataFrame(X)
    return X_df, y


def greater_than_lf(x: pd.Series, cutoff: float, label: int) -> int:
    return label if x[label] > cutoff else ABSTAIN


def make_ge_lf(cutoff: float, label: int) -> LabelingFunction:
    return LabelingFunction(
        name=f"ge-{cutoff:g}_label-{label}",
        f=greater_than_lf,
        resources=dict(cutoff=cutoff, label=label)
    )


def run():
    
    # Generate data.
    X_df, y = generate_data()
        
    # Generate very bad, not good labeling functions.
    lfs: List[LabelingFunction] = [
        make_ge_lf(cutoff=0.5, label=label) for label in range(N_CLASSES)
    ]
    
    # Initialize and apply our applier onto our training data.
    applier = PandasLFApplier(lfs=lfs)
    L_train: np.ndarray = applier.apply(df=X_df)
    
    # Initialize and fit our label model.
    label_model = LabelModel(
        cardinality=N_CLASSES,
        verbose=True,
        device="cpu"
    )
    
    tic: float = time.time()
    label_model.fit(
        L_train=L_train,
        n_epochs=10,
        lr=1e-3,
        log_freq=1,
        seed=2019
    )
    toc: float = time.time()
    print(f"Fitting label model with {N_CLASSES} classes took {toc - tic:.3f} seconds.")
```

```python
for i in range(3, 12):
    N_CLASSES: int = i
    run()
```

This gives print logs like (excluding benign PyTorch UserWarnings re:`torch.bool`):

```
Fitting label model with 3 classes took 0.051 seconds.
Fitting label model with 4 classes took 0.037 seconds.
Fitting label model with 5 classes took 0.070 seconds.
Fitting label model with 6 classes took 0.257 seconds.
Fitting label model with 7 classes took 1.543 seconds.
Fitting label model with 8 classes took 13.348 seconds.
Fitting label model with 9 classes took 130.37 seconds.
Fitting label model with 10 classes took 1403.96 seconds.
```

## Expected behavior

I suppose I expected the core model training loop (forward, loss, backprop, etc.) to be the significant consumer of runtime, especially with 10+ classes. I didn't expect the model to (attempt to) undergo a computational loop for `NPermutations(k)` iterations, which can be (very) costly.

## Screenshots

N/A

## System info

* How you installed Snorkel (conda, pip, source): `pip` (via `pipenv`)
* Build command you used (if compiling from source): N/A
* OS: Ubuntu 16.04.6 LTS
* Python version: 3.7.4
* Snorkel version: `==0.9.1`
* Versions of any other relevant libraries: `torch==1.1.0.post2`, `numpy==1.17.2`

## Additional context

I'm trying to perform a single-label, multi-class classification problem with number of classes `k ≈ 17`. I wrote `k` such labeling functions (with great ease! Thank you for the incredible interface), applied them onto a validation set, applied them onto the training set, then attempted to fit a `LabelModel`. I noticed the call wasn't returning, and thought it was the training loop. After debugging, I noticed that the call to `_break_col_permutation_symmetry` was the culprit, and upon further inspection I realized that doing `17!` iterations would be 355,687,428,096,000 (355 trillion) iterations, which even if each loop took 1 nanosecond (1e-9 seconds), would take ~247 days.

I noticed that the number of labels in the tutorials seem to limit the number of non-abstain classes to ~3 or 4, which is probably reasonable. I wonder if there's any consideration for how to have snorkel handle larger numbers of classes, especially for cases with _really_ large numbers of classes (e.g., `ImageNet`'s 1000 classes). I had the following ideas, but not sure if they're sensible given Snorkel's need to view all labeling functions and adjust the `mu` parameter accordingly:

- Build `k` one-versus-rest `LabelModel`s and aggregate their predictions accordingly. This seems more appropriate if we were doing a multi-label, binary classification approach. But I can imagine a case where we treat the problem as such, then squash and apply some softmax to the predictions. A bit hacky, but it would work?
- Use `MajorityLabelVoter`.
- Try distributing the inner loop computation via multiprocessing? Numba?
- Settle for a random search over the permutations instead of the full grid search?
- ??

I'd appreciate any advice here! I completely believe I could be missing something very obvious, so any feedback is welcome. Thank you again for your time on this project!

dataframing:

This PR addresses #1486, where we found that a one unit increase in target cardinality (`k`) results in an exponential (~8-10x) increase in runtime, due to attempting to evaluate all `k!` permutations of our labeling functions while searching for an ideal `μ` in `LabelModel._break_col_permutation_symmetry`. This makes using `LabelModel` a non-starter for multi-class classification tasks with a cardinality greater than ~8.

To address this issue, we instead implement a randomized search policy over all `k!` permutations – parameterized only by `n_iter` – which is the maximum number of distinct permutations we'd like to consider before early-stopping.

### Considerations

In general there are a few questions we should be aware of:
- Would we rather have a function that maps `cardinality -> number of iterations`? Since `cardinality -> # permutations` is exponential, we'd need to linearize by applying the logarithm. But I didn't bother to try and craft a function, since it seemed more complicated than it was worth.
- The sampling policy in place attempts to sample a unique permutation from the population of `k!` permutations. It may re-generate a previous permutation, but the final sample of permutations is unique.
- I tried sticking with `itertools.permutations` to generate our samples, but found it a bit difficult to reconcile with randomly sampling along the entire space. We could chain the returned generator with `itertools.islice`, but I'm not 100% sure what the implication would be of selecting a contiguous sequence of `n_iter` permutations and claiming it's "random." Thoughts welcome here!
- It may be worth having a generator that yields a unique sample from `np.random.permutation(k)` and memoizes prior samples' hash in a set, instead. That way, we're not generating a set of `n_iter` tuples of length 17.
- This PR sets the default `n_iter` in the `LabelModelConfig` to 1 million, which gives us an amortized performance that's between cardinalities 9 and 10. This can be set by the user at model initialization time, but worth considering if that's a proper default or if we should be more conservative.

## Related issue(s)

Fixes #1486.

## Test plan

I've added a test `test_label_model_large_multiclass` under `TestLabelModelAdvanced`, which is marked as a complex test suite. It attempts to run a `LabelModel` with `cardinality = 20`, which would take a _long_ time without a capped number of iterations. Instead, I've capped the number of iterations to 100, so it finishes pretty quickly (a few seconds, modulo local computing resources).

To test, run:

```bash
tox            # General test suite.
tox -e complex # Complex test suite.
```

Anecdotally, I've also run the example from #1486 with `10 <= N_CLASSES <= 13` and the runtime is constant with respect to the provided `n_iter`. At `n_iter: int = 500_000`, here are examples of runs on my local machine:

```
Sampling from all possible permutations...: 100%|██████████| 500000/500000 [00:43<00:00, 11510.61it/s]
Searching for ideal μ amongst candidates...: 100%|██████████| 500000/500000 [03:13<00:00, 2586.52it/s]
Fitting label model with 10 classes took 238.014 seconds.
-------------------
Sampling from all possible permutations...: 100%|██████████| 500000/500000 [00:40<00:00, 12370.89it/s]
Searching for ideal μ amongst candidates...: 100%|██████████| 500000/500000 [03:22<00:00, 2473.01it/s]
Fitting label model with 11 classes took 243.921 seconds.
-------------------
Sampling from all possible permutations...: 100%|██████████| 500000/500000 [00:40<00:00, 12325.28it/s]
Searching for ideal μ amongst candidates...: 100%|██████████| 500000/500000 [03:39<00:00, 2276.67it/s]
Fitting label model with 12 classes took 261.595 seconds.
-------------------
Sampling from all possible permutations...: 100%|██████████| 500000/500000 [00:39<00:00, 12528.03it/s]
Searching for ideal μ amongst candidates...: 100%|██████████| 500000/500000 [04:00<00:00, 2075.09it/s]
Fitting label model with 13 classes took 282.288 seconds.
-------------------
```

**Note:** the runtime still increases _a bit_ with every increase in cardinality, but that's because each sampled permutation is one item larger, which adds up over `n_iter` samples. Anecdotally, as `k` increases, the time it takes to sample `n_iter` unique permutations decreases (fewer and fewer opportunities for duplicate sampling), but the core search for an ideal `μ` increases in runtime (larger matrices being generated and multiplied).

## Checklist

Need help on these? Just ask!

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -2,7 +2,7 @@ import logging
 import pickle
 import random
 from collections import Counter
-from itertools import chain, permutations
+from itertools import chain
 from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Union
 
 import numpy as np
@@ -74,10 +74,18 @@ class LabelModelConfig(Config):
         Whether to include print statements
     device
         What device to place the model on ('cpu' or 'cuda:0', for example)
+    n_iter
+        Maximum number of iterations where LabelModel will search for a valid
+        μ value (default: 1,000,000). Note: the maximum number of iterations
+        is `k!`, which is exponential with respect to the model's cardinality.
+        A default of 1 million iterations places a cap between models with
+        cardinality 9 (9! = 362,880) and 10 (10! = 3,628,800). `n_iter` trades-off
+        runtime vs optimality of the discovered μ value.
     """
 
     verbose: bool = True
     device: str = "cpu"
+    n_iter: int = 1_000_000
 
 
 class _CliqueData(NamedTuple):
@@ -110,6 +118,7 @@ class LabelModel(nn.Module):
     >>> label_model = LabelModel(cardinality=3)
     >>> label_model = LabelModel(cardinality=3, device='cpu')
     >>> label_model = LabelModel(cardinality=3)
+    >>> label_model = LabelModel(cardinality=19, n_iter=2500000)
 
     Parameters
     ----------
@@ -804,15 +813,36 @@ class LabelModel(nn.Module):
         assumption that we could use, and in practice this may require further
         iteration here.
         """
+        import tqdm
+
         mu = self.mu.cpu().detach().numpy()
         P = self.P.cpu().detach().numpy()
         d, k = mu.shape
 
-        # Iterate through the possible perumation matrices and track heuristic scores
-        Zs = []
-        scores = []
-        for idxs in permutations(range(k)):
-            Z = np.eye(k)[:, idxs]
+        # Limit number of iterations to avoid exponentially-increasing runtime.
+        # See: https://github.com/snorkel-team/snorkel/issues/1486
+        max_n_iter: int = np.math.factorial(k)
+        n_iter: int = self.config.n_iter if self.config.n_iter and 0 < self.config.n_iter < max_n_iter else max_n_iter
+
+        # Otherwise, sample `n_iter` unique permutation indices from the full set of `max_n_iter` permutations.
+        sampled_permutations: Set[Tuple[int, ...]] = set()
+        pbar = tqdm.tqdm(desc="Sampling from all possible permutations", total=n_iter)
+        while len(sampled_permutations) < n_iter:
+            permutation: Tuple[int, ...] = tuple(np.random.permutation(range(k)))
+            if permutation not in sampled_permutations:
+                sampled_permutations.add(permutation)
+                pbar.update(n=1)
+        pbar.close()
+
+        # Iterate through the possible permutation matrices and track heuristic scores.
+        Zs: List[np.ndarray] = []
+        scores: List[int] = []
+        for permutation_indices in tqdm.tqdm(
+            iterable=sampled_permutations,
+            desc="Searching for ideal μ amongst candidates",
+            total=n_iter,
+        ):
+            Z = np.eye(k)[:, permutation_indices]
             Zs.append(Z)
 
             # If Z and P commute, get heuristic score, else skip
@@ -820,6 +850,18 @@ class LabelModel(nn.Module):
                 scores.append(self._count_accurate_lfs(mu @ Z))
             else:
                 scores.append(-1)
+
+        if max(scores) == -1 and self.config.verbose:
+            if n_iter == max_n_iter:
+                logging.warning("Could not heuristically find a valid μ value.")
+            else:
+                logging.warning(
+                    (
+                        "Could not heuristically find an ideal μ value. "
+                        "Try passing a larger `n_iter` to the LabelModel initialization "
+                        "(currently set to {n_iter:,}; maximum is {max_n_iter:,})."
+                    ).format(n_iter=n_iter, max_n_iter=max_n_iter)
+                )
 
         # Set mu according to highest-scoring permutation
         self.mu = nn.Parameter(
--- a/test/labeling/model/test_label_model.py
+++ b/test/labeling/model/test_label_model.py
@@ -506,6 +506,21 @@ class TestLabelModelAdvanced(unittest.TestCase):
         score = label_model.score(L, Y)
         self.assertGreaterEqual(score["accuracy"], 0.9)
 
+    def test_label_model_large_multiclass(self) -> None:
+        """Test the LabelModel's estimate of P and Y on a dataset with a high-cardinality target."""
+        np.random.seed(123)
+
+        # Generate label matrix with high cardinality.
+        cardinality: int = 20
+        P, Y, L = generate_simple_label_matrix(self.n, self.m, cardinality=cardinality)
+
+        # Train LabelModel and only allow for 100 permutation iterations. At
+        # cardinality = 20, the theoretical maximum number of permuted iterations
+        # is `20!`, which is ~2.43e18. This test should never finish unless we've
+        # properly bound the number of iterations from above.
+        label_model = LabelModel(cardinality=cardinality, verbose=False, n_iter=100)
+        label_model.fit(L, n_epochs=200, lr=0.01, seed=123)
+
     def test_label_model_sparse(self) -> None:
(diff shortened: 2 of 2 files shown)


dataframing:
@henryre I can't seem to tag a reviewer for this PR, but let me know if you have any questions/comments/concerns. Also attempting to tag @plison

ajratner:
@dataframing to start: thank you so much for pointing out this issue in such a detailed and thorough way, and for this PR!  We owe contributors like you big time for this kind of help and support :)

Also, apologies for some lack of coordination on our end.  We have a planned update that will (at least partially) address this issue in a more principled way than random sampling, i.e. by relying on some additional default assumptions applied to the underlying mathematical model that yields valid permutations of the learned weights.  We will be trying to get to this soon, and so I'd probably hold off on further development here for now.  I'll then circle back and see if there's a way to merge our solution with some of the work done here. Thanks!

dataframing:
@ajratner not a problem! It was a fun exercise. I'm looking forward to the update, and I'll be on the lookout for other ways to contribute. Thanks to the team for their hard work on this! 👍

henryre:
@dataframing @ajratner miscommunication was on me. Thanks for the effort on this!

plison:
I've been experimenting with @dataframing 's code, as I am working on a classification problem with 18 distinct classes (so enumerating all possible permutations is obviously not feasible).  Just wanted to say that the code runs, but I noticed a bug at the end of the (proposed) ` _break_col_permutation_symmetry` method, namely that the `mu` parameter is modified  (line 866-868 of `label_model.py`) even when a valid μ value isn't found. This is a bad idea, as in that case, the `mu` parameter will transformed into a random permutation that is much worse than the initial `mu` value.  I think a `return` should be added on line 865 to avoid changing the `mu` parameter when no solution is found.

(just wanted to mention this in case someome else tries that code and stumbles upon the same problem).

dataframing:
@plison I agree, but this also seems to be what's done on `master`, no? There's no guard on whether or not to update the `mu` instance attribute.

https://github.com/snorkel-team/snorkel/blob/b8e47c23af012e0396d8d55aaf9a74d5c2376bf7/snorkel/labeling/model/label_model.py#L824

So perhaps this is a separate, but related issue?

dataframing:
I made a comment a moment ago, but I think it's wrong. Yes, I think if we're only sampling a fraction of the space of all permutations, we should make sure to not update the `mu` instance unless we have a valid (non -1) entry in `scores`.

eggie5:
@ajratner I'm on the lookout for this also, as I just moved from 4 classes to 12 and I noticed the LabelModel does not seem to finish training! What is a workaround? CUDA or use the Majority Voter?

PR #1502, opened 2019-10-30
Faster symmetry breaking

brahmaneya:
Use the Munkres algorithm instead of brute force search to find the optimal permutation for breaking column symmetry. Our objective for symmetry breaking is to maximize the total probability of LFs being accurate (summed over LFs). We optimize this over combinations of permutations of columns corresponding to labels with the same prior probability. This is equivalent to optimizing the trace of the product of the summed probabilities matrix with the permutation matrix, which can be done in O(n^3) time.

Fixes # (issue)
#1486 

## Test plan
Modified existing symmetry breaking test.

## Checklist

Need help on these? Just ask!

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -772,19 +772,16 @@ class LabelModel(nn.Module):
         LFs being accurate when not abstaining.
 
             \sum_lf \sum_{y=1}^{cardinality} P(\lf = y, Y = y)
-
-        This is the standard assumption we have made in algorithmic and theoretical
-        work to date. Note however that this is not the only possible heuristic /
-        assumption that we could use, and in practice this may require further
-        iteration here.
         """
         mu = self.mu.cpu().detach().numpy()
         P = self.P.cpu().detach().numpy()
         d, k = mu.shape
+        # We want to maximize the sum of diagonals of matrices for each LF. So
+        # we start by computing the sum of conditional probabilities here.
         probs_sum = sum([mu[i : i + k] for i in range(0, self.m * k, k)]) @ P
 
-        m = Munkres()
-        Z = np.zeros([self.cardinality, self.cardinality])
+        munkres_solver = Munkres()
+        Z = np.zeros([k, k])
 
         # Compute groups of indicess with equal prior in P.
         groups: DefaultDict[float, List[int]] = defaultdict(list)
@@ -796,10 +793,10 @@ class LabelModel(nn.Module):
                 continue
             # Compute submatrix corresponding to the group.
             probs_proj = probs_sum[[[g] for g in group], group]
-            # Use the Munkres algorithm to find optimal permutation.
+            # Use the Munkres algorithm to find the optimal permutation.
             # We use minus because we want to maximize diagonal sum, not minimize,
             # and transpose because we want to permute columns, not rows.
-            permutation_pairs = m.compute(-probs_proj.T)
+            permutation_pairs = munkres_solver.compute(-probs_proj.T)
             for i, j in permutation_pairs:
                 Z[group[i], group[j]] = 1.0
 
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -1,14 +1,15 @@
 import logging
 import pickle
 import random
-from collections import Counter
-from itertools import chain, permutations
-from typing import Any, Dict, List, NamedTuple, Optional, Set, Tuple, Union
+from collections import Counter, defaultdict
+from itertools import chain
+from typing import Any, DefaultDict, Dict, List, NamedTuple, Optional, Set, Tuple, Union
 
 import numpy as np
 import torch
 import torch.nn as nn
 import torch.optim as optim
+from munkres import Munkres  # type: ignore
 
 from snorkel.analysis import Scorer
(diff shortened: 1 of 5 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -794,10 +768,10 @@ def _break_col_permutation_symmetry(self) -> None:
...
         satisfy these objectives, and thus is an equally valid (symmetric) solution.
-        Therefore, we select the solution where the most LFs are estimated to be more
-        accurate than not when not abstaining, i.e., where for the majority of LFs,
+        Therefore, we select the solution that maximizes the summed probability of the
+        LFs being accurate when not abstaining.
 
-            P(\lf = Y) > P(\lf != Y, \lf != -1).
+            \sum_lf \sum_{y=1}^{cardinality} P(\lf = y, Y = y)
 
         This is the standard assumption we have made in algorithmic and theoretical
```
bhancock8: Update comment? This is no longer the standard assumption we have made to date.

snorkel/labeling/model/label_model.py
```diff
@@ -807,25 +781,31 @@ def _break_col_permutation_symmetry(self) -> None:
...
-            # If Z and P commute, get heuristic score, else skip
-            if np.allclose(Z @ P, P @ Z):
-                scores.append(self._count_accurate_lfs(mu @ Z))
-            else:
-                scores.append(-1)
-
-        # Set mu according to highest-scoring permutation
+        probs_sum = sum([mu[i : i + k] for i in range(0, self.m * k, k)]) @ P
+
+        m = Munkres()
```
bhancock8: No need for such a short var name here. It's used far away from this definition and just the once.

snorkel/labeling/model/label_model.py
```diff
@@ -807,25 +781,31 @@ def _break_col_permutation_symmetry(self) -> None:
...
+        for i, f in enumerate(P.diagonal()):
+            groups[np.around(f, 3)].append(i)
+        for group in groups.values():
+            if len(group) == 1:
+                Z[group[0], group[0]] = 1.0  # Identity permutation
+                continue
+            # Compute submatrix corresponding to the group.
+            probs_proj = probs_sum[[[g] for g in group], group]
+            # Use the Munkres algorithm to find optimal permutation.
+            # We use minus because we want to maximize diagonal sum, not minimize,
```
bhancock8: Nice comments

snorkel/labeling/model/label_model.py
ajratner: We could just delete this comment.  I guess we've often made a slightly different assumption, namely that more LFs are > 50% than not (phrased in the binary, single LF param setting), rather than that the sum of their accuracies is maximized... but not sure worth confusing people here?  Just delete this bit?

test/labeling/model/test_label_model.py
```diff
@@ -449,34 +449,39 @@ def test_symmetry_breaking(self):
         mu = np.array(
             [
                 # LF 0
-                [0.75, 0.25],
-                [0.25, 0.75],
+                [0.75, 0.15, 0.1],
+                [0.20, 0.75, 0.3],
+                [0.05, 0.10, 0.6],
```
ajratner: Is there a reason not to also keep the previous k=2 test?

ajratner: I would just keep them both?

snorkel/labeling/model/label_model.py
```diff
@@ -807,25 +781,31 @@ def _break_col_permutation_symmetry(self) -> None:
...
-            Zs.append(Z)
-
-            # If Z and P commute, get heuristic score, else skip
-            if np.allclose(Z @ P, P @ Z):
-                scores.append(self._count_accurate_lfs(mu @ Z))
-            else:
-                scores.append(-1)
-
-        # Set mu according to highest-scoring permutation
+        probs_sum = sum([mu[i : i + k] for i in range(0, self.m * k, k)]) @ P
```
ajratner: Can add a one line comment here- we want to maximize a sum of (the diagonals of) matrices, therefore we can just sum first here

snorkel/labeling/model/label_model.py
```diff
@@ -807,25 +781,31 @@ def _break_col_permutation_symmetry(self) -> None:
...
+
+        m = Munkres()
+        Z = np.zeros([self.cardinality, self.cardinality])
```
ajratner: Should be consistent about using `self.cardinality` vs. `k` for clarity?

snorkel/labeling/model/label_model.py
```diff
@@ -807,25 +781,31 @@ def _break_col_permutation_symmetry(self) -> None:
...
+                Z[group[0], group[0]] = 1.0  # Identity permutation
+                continue
+            # Compute submatrix corresponding to the group.
+            probs_proj = probs_sum[[[g] for g in group], group]
+            # Use the Munkres algorithm to find optimal permutation.
+            # We use minus because we want to maximize diagonal sum, not minimize,
+            # and transpose because we want to permute columns, not rows.
+            permutation_pairs = m.compute(-probs_proj.T)
+            for i, j in permutation_pairs:
+                Z[group[i], group[j]] = 1.0
```
ajratner: Beautiful!!

PR #1540, opened 2020-01-27
Fix squeeze bug for length 1 array, and inconsistency on requirements and setup.py 

brahmaneya:
#Description of proposed changes
Squeezing a length 1 array for used to create a scalar instead of a 1D array as intended. Fixed by adding an extra check.
Also made munkres version in setup.py consistent with requirements.

## Test plan
Update unit test for changed function.

## Checklist

Need help on these? Just ask!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -35,7 +35,7 @@ setup(
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        "munkres==1.1.2",
+        "munkres>=1.0.6",
         "numpy>=1.16.0,<2.0.0",
         "scipy>=1.2.0,<2.0.0",
         "pandas>=0.25.0,<0.26.0",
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -35,7 +35,7 @@
     packages=find_packages(),
     include_package_data=True,
     install_requires=[
-        "munkres==1.1.2",
+        "munkres>=1.0.6",
```
vincentschen: are these version commits related to this patch?

test/utils/test_core.py
```diff
@@ -21,6 +21,10 @@ def test_to_int_label_array(self):
         Y = to_int_label_array(X, flatten_vector=True)
         np.testing.assert_array_equal(Y, Y_expected)
 
+        Y = to_int_label_array(np.array([[1]]), flatten_vector=True)
+        Y_expected = np.array([1])
+        np.testing.assert_array_equal(Y, Y_expected)
+
```
vincentschen: nice!

setup.py
brahmaneya: They're not, but `tox -e check` was failing without it.

brahmaneya: I'll try reverting it and see if travis ci passes.

brahmaneya: CI fails if I revert.

PR #1565, opened 2020-04-05
Upgrade pytorch requirement to >=1.2.0

vincentschen:
* Update dev + setup.py requirements for `torch`
* Use native `torch.utils.tensorboard` and remove `tensorboardX` dep
* Moved `lr_scheduler.step()` after `optimizer.step()`
* Replace deprecated uses of `.byte()` with `.bool()`

## Related issue(s)
Fixes https://github.com/snorkel-team/snorkel/issues/1558

## Test plan
* All unit tests pass with `torch==1.2.0, 1.3.0, 1.4.0`
* Updated Slicing convergence test for longer fit.
* `tox -e {complex,spark}` passes with `torch==1.2.0, 1.3.0, 1.4.0`

## Checklist
Need help on these? Just ask!


snorkel/classification/training/trainer.py
```diff
@@ -240,6 +237,9 @@ def fit(
                 # Update the parameters
                 self.optimizer.step()
 
+                # Update lr using lr scheduler
+                self._update_lr_scheduler(total_batch_num)
```
bhancock8, line 241: Why is this code block being moved? If it doesn't have to do with an issue introduced by the pytorch upgrade, then would strongly recommend it occur in separate PR.

requirements.txt
```diff
@@ -18,15 +18,14 @@ tqdm>=4.33.0,<5.0.0
 
 # Internal models
 scikit-learn>=0.20.2,<0.22.0
-torch>=1.1.0,<1.2.0
+torch>=1.4.0,<2.0.0
```
henryre, line 21: Is this incompatible with 1.2 and 1.3?

snorkel/classification/training/trainer.py
vincentschen: yep, it's related to the upgrade. otherwise, pytorch throws this warning:
```
UserWarning: Detected call of `lr_scheduler.step()` before `optimizer.step()`. In PyTorch 1.1.0 and later, you should call them in the opposite order: `optimizer.step()` before `lr_scheduler.step()`
```

requirements.txt
vincentschen: yes, all tests pass with the older versions 1.2.0 and 1.3.0.
relaxing the requirement to `torch>=1.2.0,<2.0.0`

setup.py
```diff
@@ -41,9 +41,9 @@
         "pandas>=0.25.0,<0.26.0",
         "tqdm>=4.33.0,<5.0.0",
         "scikit-learn>=0.20.2,<0.22.0",
-        "torch>=1.1.0,<1.2.0",
+        "torch>=1.2.0,<2.0.0",
+        "tensorboard>=1.14.0,<2.0.0",
```
henryre, line 45: how did we decide on this range?

henryre:
@vincentschen can we update the Test Plan with details from the multiple version tests?

setup.py
vincentschen: I got an import error that looked like this:
```
ImportError: TensorBoard logging requires TensorBoard with Python summary writer installed. This should be available in 1.14 or above
```

vincentschen:
> @vincentschen can we update the Test Plan with details from the multiple version tests?

Yep! Naively, pinned the version to each minor release and ran all unit/integration tests.

PR #1568, opened 2020-04-05
[BUMP]: v0.9.4+dev

vincentschen:
Part of the 0.9.4 release

### DIFF
--- a/snorkel/version.py
+++ b/snorkel/version.py
@@ -1,6 +1,6 @@
 _MAJOR = "0"
 _MINOR = "9"
-_REVISION = "4"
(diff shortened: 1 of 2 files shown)


snorkel/version.py
```diff
@@ -1,6 +1,6 @@
 _MAJOR = "0"
 _MINOR = "9"
-_REVISION = "4"
+_REVISION = "4+dev"
```
henryre, line 3: 5+dev!

vincentschen: whoops thanks!

PR #1572, opened 2020-04-08
Feature/memoize key

Wirg:
### Feature

Currently, there is no way to decide the key to be memoized when using preprocessor(memoize=True).
This leads to 2 issues :
- memoization can not be done for unhashable classes (typically a group of pandas rows). We need to wrap or subclass it.
- memoization key can not be specific to a preprocessing.
Example : We are trying to evaluate the reliability of a paragraph in a blog post.
We could evaluate the reliability of the paragraph and of the website.
The preprocessing corresponding to those 2 tasks will share the same key for memoize, which is not ideal : a website can have a few thousand paragraphs so we will evaluate website reliability a lot more than necessary.

### Result

```py
@preprocessor(memoize=True, memoize_key=lambda p: p.base_website_url)
def add_website_reliability(paragraph):
    paragraph.website_reliability = evaluate_reliability(paragraph.base_website_url)
    return paragraph
```

### Implementation

Add a `memoize_key : Optional[HashingFunction]` to the `BaseMapper`, if provided and not `None`, it will be used instead of `get_hashable` to define the hash of the input.

`memoize_key` has been made accessible to the different functions providing `memoize` api.

## Related issue(s)

https://github.com/snorkel-team/snorkel/issues/1561

## Checklist

Need help on these? Just ask!

@henryre , I will be waiting for your review.
I ran `tox -e doc`, but it did not produce any change and I have a bunch of `WARNING: toctree contains reference to nonexisting document`, is it normal ?
By the way, how would you like to discuss my use case further ?


Wirg:
@henryre I just noticed it won't work in the expected way.

The snorkel pattern is to return the x_mapped so the cache will change the data point.

```py
    def test_decorator_mapper_memoized_use_memoize_key(self) -> None:
        square_hit_tracker = SquareHitTracker()

        @lambda_mapper(memoize=True, memoize_key=lambda x: x.num)
        def square(x: DataPoint) -> DataPoint:
            x.num_squared = square_hit_tracker(x.num)
            return x

        x8 = self._get_x()
        x8_mapped = square(x8)
        assert x8_mapped is not None
        self.assertEqual(x8_mapped.num_squared, 64)
        self.assertEqual(square_hit_tracker.n_hits, 1)
        x8_with_another_text = self._get_x(text="Henry is still having fun")
        x8_with_another_text_mapped = square(x8_with_another_text)
        assert x8_with_another_text_mapped is not None
        self.assertEqual(x8_with_another_text_mapped.num_squared, 64)
        self.assertEqual(square_hit_tracker.n_hits, 1)
        # This should fail :/
        self.assertEqual(x8_with_another_text_mapped, x8_mapped)
```

henryre:
Hi @Wirg, thanks for putting this up! Based on the example you put up, the expected behavior would be `square(x8_with_another_text) == x8_mapped` since the hashing function was (intentionally) "poorly chosen" in the test. Are you saying `self.assertEqual(x8_with_another_text_mapped, x8_mapped)` will trigger an AssertionError in the current implementation?

Wirg:
Hi @henryre ,

I hope you're going well.

Small bump on this PR.

Wirg:
Hi @henryre ,

Another bump for this pr.

What do you want me to do ?

Do you have some change ? Do you want to give up on this feature ?

henryre:
Hi @Wirg, sorry for the delay here and thanks for the reminder! Taking a look today!

test/map/test_core.py
```diff
@@ -263,6 +263,25 @@ def square(x: DataPoint) -> DataPoint:
...
 
+    def test_decorator_mapper_memoized_use_memoize_key(self) -> None:
+        square_hit_tracker = SquareHitTracker()
+
+        @lambda_mapper(memoize=True, memoize_key=lambda x: x.num)
+        def square(x: DataPoint) -> DataPoint:
+            x.num_squared = square_hit_tracker(x.num)
+            return x
+
+        x8 = self._get_x()
```
henryre, line 274: It might be preferable to have an example of canonical usage in this unit test. For example, using a custom unique ID:

```
x1 = SimpleNamespace(uid="id1", num=8, unhashable=some_unhashable_pandas_object)
x2 = SimpleNamespace(uid="id1", num=8, unhashable=some_unhashable_pandas_object)
...
```

and then we use `memoize_key = lambda x: x.uid`

snorkel/map/core.py
```diff
@@ -106,9 +108,18 @@ class BaseMapper:
         Memoize mapper outputs?
     """
 
-    def __init__(self, name: str, pre: List["BaseMapper"], memoize: bool) -> None:
+    def __init__(
+        self,
+        name: str,
+        pre: List["BaseMapper"],
+        memoize: bool,
+        memoize_key: Optional[HashingFunction] = None,
```
henryre, line 116: Any reason to prefer `None` as the default instead of just using `get_hashable` directly?

test/map/test_core.py
Wirg: Great idea.

snorkel/map/core.py
Wirg: I see 2 reasons :
- function are mutable and the "good practice" is to avoid mutable in default parameters. I don't really see a situation where we would mutate this function tho.
- we will have to import get_hashable in all the subclasses and all the wrappers of BaseMapper

henryre: Great point about mutating functions in modules, hadn't thought of that route. Sounds good!

test/map/test_core.py
henryre: @Wirg let me know when you can make this change, then happy to approve the PR! Thanks again for your patience!

Wirg: So I did the change. I am not fully satisfied. If tomorrow someone change :
- `get_hashable` to support pandas dataframe
- `memoize_key` not to be used

The test won't fail.

Wirg:
@henryre 

I finally changed the test. I am not fully satisfied. If tomorrow someone changes :
- `get_hashable` to support pandas dataframe
- `memoize_key` not to be used

The test won't fail.

henryre:
@Wirg good thinking! You could add an additional field in the test called `not_used` and have different values for the two data points

Wirg:
@henryre so I added a `not_used` int.

I am still encountering new test failures.
I fixed F541 (f-string used with no parameters).
I am facing a typing failure due to `torch.nn.Linear` usage in `snorkel/slicing/utils`.
I am not sure what are my steps on this ?
Is this already fixed and I should rebase ?

Wirg:
@henryre small bump

I don't know what I should be doing regarding codecov.
Am I expected to add more tests. If yes, where ?

henryre:
@Wirg looks like codecov was being a bit temperamental, will go ahead and merge in. Thanks for your hard work here!

PR #1578, opened 2020-04-17
Update releasing notes for builds on RTD

vincentschen:
Update the release plan so we remember to trigger a build on `latest`.

## Related issue(s)
https://github.com/snorkel-team/snorkel/issues/1575


vincentschen:
> nice! what do we do if it isn't?

if the build isn't triggered, it will just be stale, right?
also, do we need both this and `master` ? seems like the same thing

PR #1629, opened 2021-02-12
Support Training With Sparse Matrices

### LINKED ISSUE #1625: LabelModel should support loading sparse matrices
## Problem I Want To solve
I've found it easy to generate millions of labels with label functions, but loading them into Snorkel is hard. 
The problem is the conversion to [augmented format](https://github.com/snorkel-team/snorkel/blob/ed7771812a0484ad593485dc9e7c67091d483e37/snorkel/labeling/model/label_model.py#L147)  and (for training) the [calculation of the O matrix](https://github.com/snorkel-team/snorkel/blob/ed7771812a0484ad593485dc9e7c67091d483e37/snorkel/labeling/model/label_model.py#L243). 

## Describe the solution you'd like
In addition to letting the user load the full label matrix (n_docs,n_funcs), we can let the user load the indicator matrix (n_docs,n_funcs*n_labels) in sparse format. 
e.g. user would input a list of tuples (doc_id,func_id*num_labels+label_id) and populate a sparse matrix. 
This makes the L.T@L calculation cheap, and saves lots of time and memory building indicator matrix.

Torch supports Sparse matrices, so we could even do training and inference without the memory hassle of the dense L matrix. 

### Example:
I calculate and store the label functions in SQL, so it's easy to generate that list of tuples. 
image

### Caveat
This would make modelling dependencies between LFs harder, but since [_create_tree is degenerate](https://github.com/snorkel-team/snorkel/blob/ed7771812a0484ad593485dc9e7c67091d483e37/snorkel/labeling/model/label_model.py#L589) that doesn't seem to be an issue in practice. 

## Describe alternatives you've considered
The other alternative is some "big-data" solution, but that's a lot of friction for something I can do so simply. 

## Additional context
I'm implementing this anyway for my own fun, happy to contribute it back if theirs interest

talolard:
# Disclaimer

This PR isn't done. It does what it's supposed to do and has tests, but style and code cleanliness might not be there. 

I'm not totally confidant this implementation is a fit. I'd appreciate if someone could take a look and let me know if I'm on track before I polish this.   Maybe @bhancock8 who replied to the original issue ? 

Adds support for training and inference with sparse matrices. 

This PR adds a few convenience functions to help the user work with sparse matrices representations of L_ind / or the objective matrix (do either have a formal name ? ).

I presume most users will call 'train_model_from_sparse_event_cooccurence', which takes a list of tuples representing L_ind indices and value (which is always 1), populate a sparse matrix and runs training. 

train_model_from_sparse_event_cooccurence calls 'train_model_from_known_objective' which gets a dense numpy representation of O  and trains. When I use Snorkel I call this function and calculate O elsewhere, it's faster. 

Internally, there is some refactoring in LabelModel to support  `train_model_from_known_objective`, constants are set differently and the tree and clique data calculations are moved a little. 

## Related issue(s) 
Fixes #1625 

## Test plan

I wrote tests in test_sparse_data_helpers. 
Basically the tests create an L matrix in standard format, and then compare the output of normal Snorkel to Sparse Snorkel. 

## Checklist

Need help on these? Just ask!


bhancock8 (commented):
Thanks for submitting this draft! As you mentioned, there would be a separate pass on style, linting, etc. before this could be merged, so I mostly just took a look at the high level structure (except for where I couldn't help myself to comment on individual lines). 

I can cc a couple of the original authors of the label model to double check correctness of the new methods once the structure is in a good place. From a structural standpoint, you're right that this doesn't quite match the style of the methods around it. The changes to label_model.py are fairly general, mostly breaking large blocks of code into smaller methods, which isn't a bad idea anyway, especially given the size of that file. But the helper methods aren't object-oriented like the other primary classes in `labeling/model/`. 

There are two ways that I could see this moving forward:
1. Move these helper methods to a sparse_label_model project under `snorkel/contrib` (see the README there for details), where they can be used by others as helpful methods, but are less integrated and not guaranteed the same level of long-term support. 
2. Create a `SparseLabelModel` class that inherits from `LabelModel` and adds these new functionalities as part of specialized `fit` methods. This keeps the new complexity away from the simple case where L is dense in `label_model.py`, while maintaining the object-centric design used in that module.

snorkel/labeling/model/label_model.py
```diff
@@ -876,15 +906,41 @@ def fit(
             )
 
         self._set_constants(L_shift)
-        self._set_class_balance(class_balance, Y_dev)
-        self._create_tree()
+        self._common_training_preamble(
```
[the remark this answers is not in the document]
talolard: What do you think of setting the random seed in the class constructor ?

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,223 @@
+# -*- coding: utf-8 -*-
+"""Sparse Data Helpers
+
+Indexing throughout this module is 0 based, with the assumption that "abstains" are ommited.
+
+When working with larger datasets, it can be convenient to load the data in sparse format. This module
+provides utilities to do so. We provide functions for a number of cases.
+
+The user has the AugmentedMatrix (L_ind) in tuple form. AugmentedMatrix is of shape (num_examples,numfuncs*num_classes)
```
bhancock8, line 9: Nit: Let's specify each of these as "Case 1", "Case 2", etc. so it's both (a) more obvious that these are mutually exclusive, and (b) easier to reference.

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,223 @@
...
+    """
+
+    def __init__(self, known_dimensions: Optional[KnownDimensions] = None):
+        super().__init__()
+        if known_dimensions:
+            self._set_constants(known_dimensions=known_dimensions)
+
+    def get_weights(self) -> np.ndarray:
+        r""" SparseLabelModel doesn't support this method.
+        We need to calucalte coverage from a sparse format which is not implemented.
```
bhancock8, line 50: calucalte

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,223 @@
...
+
+    def predict(
+        self,
+        cliquesets: CliqueSetList,
+        return_probs: Optional[bool] = False,
+        tie_break_policy: str = "abstain",
+    ) -> Union[CliqueSetProbs, CliqueSetProbsAndPreds]:
+        r"""Runs prediction on a ```CliqueSetList```
+
+        A ```LabelModels``` output is determined by the "Events" that cooccured, which we call a CliqueSet.
```
bhancock8, line 68: Nit: ```LabelModel```'s

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,223 @@
...
+            A 2-tuple whose first element is a list of CliqueSets and whose second element is an [len(cliquesets),k] array
+            such that ar[i] are the probabilities for cliqueset i
+
+        CliqueSetProbsAndPreds
+            A 3-tuple whose first element is a list of CliqueSets and whose second element is an [len(cliquesets),k] array,
+            and third element as [len(cliquestes),1] array such that ar_1[i] are the probabilities for cliqueset i
+            and ar_2[i] is the predicted class for that cliqueset.
+
+        """
+        # the users cliqueset might be an unordered iterable (set) so we take the ordered list
```
bhancock8, line 95: Nit: Comments start with capital letters in this repo

snorkel/labeling/model/sparse_label_model/sparse_label_model_helpers.py
```diff
@@ -0,0 +1,69 @@
...
+     This could be a dict, but we want to preserve the whole array to leverage the pre-existing code for calculating
+     a predicted class
+"""
+CliqueSetProbsAndPreds = Tuple[CliqueSetList, np.ndarray, np.ndarray]
+"""
+    CliqueSetProbsAndPreds extends CliqueSetProbs with a third tuple element, an array of predicted class_ids
+"""
+
+
+class UnnormalizedObjective(Exception):
```
bhancock8, line 30: Nit: UnnormalizedObjectiveException (make it more clear this is an exception and not an object when it's used elsewhere in the code)

snorkel/labeling/model/label_model.py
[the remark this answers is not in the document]
bhancock8: That still leaves the possibility of a user expecting deterministic behavior but not getting it if they're, for example, not initializing the class and training immediately back-to-back. If there's a strong case for putting it there, I'm open to hearing it, but all other things being equal, keeping it as fool-proof as possible (as close to where the randomness is relied on as possible) has preference.

snorkel/labeling/model/label_model.py
```diff
@@ -876,15 +904,41 @@ def fit(
...
         lf_analysis = LFAnalysis(L_train)
         self.coverage = lf_analysis.lf_coverages()
 
         # Compute O and initialize params
         if self.config.verbose:  # pragma: no cover
             logging.info("Computing O...")
         self._generate_O(L_shift)
+        self._common_training_loop()
+
+    def _common_training_preamble(
```
bhancock8, line 919: Nit: we can drop the `_common` prefix here and for the training loop method below, since the base class doesn't necessarily need to be aware that it has children that will be using these methods.

bhancock8 (commented):
Nice! This is moving in the right direction—the class abstraction and subdirectory feel pretty good stylistically—no added complexity in the base case, but similar usage patterns.

I added a few hyper-local suggestions just from quick browsing. There are still some linting issues, of course, and I'll do a deeper pass once you feel like most of the big structural movement is done. Explicitly re-request review once tests are passing. I'm also tagging @fredsala for help reviewing the label model logic.

talolard:
I think the coverage tool isn't picking up on some of the tests. 
Static methods in 

`sparse_example_eventlist_label_model.py` and  `sparse_event_pair_label_model.py` get tested explicitly in the two tests I marked with `@pytest.mark.complex`

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,227 @@
+# -*- coding: utf-8 -*-
+"""Sparse Data Helpers.
+
+Indexing throughout this module is 0 based, with the assumption that "abstains" are ommited.
+
+When working with larger datasets, it can be convenient to load the data in sparse format. This module
```
bhancock8, line 6: Nit: Let's move this summary of this file's purpose above the indexing disclaimer.

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,227 @@
...
+
+Case 1:
+    The user has the AugmentedMatrix (L_ind) in tuple form. AugmentedMatrix is of shape
+    (num_examples,numfuncs*num_classes)
+    and the user has a list of tuples (i,j) that indicate that event j occoured for example i.
+
+Case 2:
+    The user has a list of 3-tuples(i,j,k) such that for document i, labeling function j predicted class k.
+
+The Case 3:
```
bhancock8, line 17: Drop "The" for consistency

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,227 @@
...
+When working with larger datasets, it can be convenient to load the data in sparse format. This module
+provides utilities to do so. We provide functions for a number of cases.
+
+Case 1:
+    The user has the AugmentedMatrix (L_ind) in tuple form. AugmentedMatrix is of shape
+    (num_examples,numfuncs*num_classes)
+    and the user has a list of tuples (i,j) that indicate that event j occoured for example i.
+
+Case 2:
+    The user has a list of 3-tuples(i,j,k) such that for document i, labeling function j predicted class k.
```
bhancock8, line 15: Nit: space before (i,j,k)

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,227 @@
...
+
+The Case 3:
+    user has a list of 3-tuples (i,j,c) where i and j range over [0,num_funcs*num_classes] such that
+    the events  i and j were observed to have co-occur c times.
```
bhancock8, line 19: Nit: extra space here and in case 5

snorkel/labeling/model/sparse_label_model/base_sparse_label_model.py
```diff
@@ -0,0 +1,227 @@
...
+        Returns
+        -------
+        CliqueSetProbs
+            A 2-tuple whose first element is a list of CliqueSets and whose second element is an [len(cliquesets),k] array
+            such that ar[i] are the probabilities for cliqueset i
+        """
+        rows = []
+        cols = []
+        data = []
+        cliquesets_list = (
```
bhancock8, line 135: Remove the parentheses, since it's just a list, not a tuple

snorkel/labeling/model/label_model.py
```diff
@@ -400,9 +409,14 @@
                [0., 1.],
                [0., 1.]])
         """
-        L_shift = L + 1  # convert to {0, 1, ..., k}
-        self._set_constants(L_shift)
-        L_aug = self._get_augmented_label_matrix(L_shift)
+        if not is_augmented:
+            # This is the usual mode
+            L_shift = L + 1  # convert to {0, 1, ..., k}
+            self._set_constants(L_shift)  # TODO - Why do we need this here ?
```
bhancock8, line 415: self._get_augmented_label_matrix uses at least self.cardinality, which is set in this method. Remove the TODO?

bhancock8 (commented):
Looking really close here! There are a handful of .idea files we need to pull out of the commit, and then some small formatting tweaks. I agree—codecov doesn't seem to be giving credit for @pytest.mark.complex methods but I can see that they're actually covered. If you can address the above comments and do one more light pass for typos/spacing/formatting, I think this should be ready to land!

PR #1637, opened 2021-03-10
Modify test to permit higher cardinality cases for label model

### LINKED ISSUE #1631: TestLabelModelAdvanced fails when changing cardinality
## Issue description
I run the tests in `TestLabelModelAdvanced`.
The setUp is 
```python
    def setUp(self) -> None:
        """Set constants for the tests."""
        self.m = 10  # Number of LFs
        self.n = 10000  # Number of data points
        self.cardinality = 2  # Number of classes

```
That test passes. 

If I change cardinaltiy to 3,4 ... the tests fail. 

## Code example/repro steps

See above
## Expected behavior
The tests should pass regardless of cardinality 

## System info

* How you installed Snorkel (conda, pip, source): source
* Build command you used (if compiling from source): tox 
* OS: Ubuntu
* Python version: 3.6.9
* Snorkel version:'0.9.6+dev'

## Additional context
Relates to #1625  and #1629 .

When I run this test on the classes I've added in #1629  they fail, and I don't know if it's just because of my code or also because the test is "specific"

fredsala:
The test_label_model_basic and test_label_model_sparse tests are specific to binary class; increasing the cardinality of the problem beyond 2 (which increases the number of parameters) prevents the tests from passing.

This change modifies the tests to use average error (which is invariant to the number of parameters) instead of maximum error (which makes higher-cardinality tests increasingly difficult).

## Related issue(s)
Fixes #1631

## Test plan
I ran the scenario in #1631 with cardinality 3,4,...,10 and the tests pass.

## Checklist

Need help on these? Just ask!

* [X] I have run `tox -e complex` and/or `tox -e spark` if appropriate.
* [X] All new and existing tests passed.

### DIFF
--- a/test/labeling/model/test_label_model.py
+++ b/test/labeling/model/test_label_model.py
@@ -559,7 +559,10 @@ class TestLabelModelAdvanced(unittest.TestCase):
 
         # Test estimated LF conditional probabilities
         P_lm = label_model.get_conditional_probs()
-        np.testing.assert_array_almost_equal(P, P_lm, decimal=2)
+        conditional_probs_err = (
+            np.linalg.norm(P.flatten() - P_lm.flatten(), ord=1) / P.size
+        )
+        self.assertLessEqual(conditional_probs_err, 0.01)
 
         # Test predicted labels
         score = label_model.score(L, Y)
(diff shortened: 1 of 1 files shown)


bhancock8 (approved):
Since this is a non-trivial computation, let's either update the name of the variable to something more explicit or add a comment on what exactly we're calculating. Then lgtm!

fredsala:
Ready for merge.

PR #1651, opened 2021-06-14
Update README with link

minhtuev:
Update README with Snorkel AI recruiting message


bhancock8 (approved):

rjurney:
What about some kind of statement that this project will continue to be maintained and isn't abandoned? That eventually you'll do more work here? Is that realistic?

PR #1652, opened 2021-07-03
Fixing log issue for Label Model

### LINKED ISSUE #1638: LabelModel.fit() Not Showing Output in Jupyter Notebook/Jupyterlab Cell
## Issue description

When calling LabelModel.fit() (with a verbose LabelModel), there is no output to the cell on the loss of the model.
Is this expected because of errors in Jupyter? Are there any workarounds? It's tougher to gauge how well the model is doing and to tune it accordingly. Would it be possible to instead return the loss from fit() like some frameworks do? 

## Code example/repro steps

`label_model = LabelModel(cardinality=2, verbose=False)`
`label_model.fit(L_train=L_train, n_epochs=100, log_freq=10, seed=123, lr=3e-5)`
## Expected behavior
Loss printed out 10 times (100 / 10) 

## System info

* Build command you used (if compiling from source): 
* OS: macOS Big Sur
* Python version: 3.8.6
* Snorkel version: 0.9.7
* Versions of any other relevant libraries: JupyterLab 2.2.0

anerirana:
Changing log level to `INFO` when verbose is set to `True` in `LabelModel.fit()`

## Related issue(s)

Fixes #1638 

## Test plan
No new test cases are required

## Checklist

Need help on these? Just ask!

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -883,7 +883,7 @@ class LabelModel(nn.Module, BaseLabeler):
         np.random.seed(self.train_config.seed)
         torch.manual_seed(self.train_config.seed)
 
-        # Set Logger
+        # Set Logger 
         self._set_logger()
 
         L_shift = L_train + 1  # convert to {0, 1, ..., k}
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -883,7 +883,7 @@ class LabelModel(nn.Module, BaseLabeler):
         np.random.seed(self.train_config.seed)
         torch.manual_seed(self.train_config.seed)
 
-        # Set Logger 
+        # Set Logger
         self._set_logger()
 
         L_shift = L_train + 1  # convert to {0, 1, ..., k}
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -622,6 +622,8 @@ class LabelModel(nn.Module, BaseLabeler):
 
     def _set_logger(self) -> None:
         self.logger = Logger(self.train_config.log_freq)
+        if self.config.verbose:
+            logging.basicConfig(level=logging.INFO)
 
     def _set_optimizer(self) -> None:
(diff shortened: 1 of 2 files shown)


anerirana:
@bhancock8, I have made the discussed changes. But, the pipeline is failing at the build stage for python3.6, as the tox command does not complete in 15 minutes Travis wait time. Can you please suggest a way forward for this? Thank you.

snorkel/labeling/model/label_model.py
```diff
@@ -622,6 +622,8 @@ def _execute_logging(self, loss: torch.Tensor) -> Metrics:
 
     def _set_logger(self) -> None:
         self.logger = Logger(self.train_config.log_freq)
+        if self.config.verbose:
+            logging.basicConfig(level=logging.INFO)
```
bhancock8, line 626: This appears to be changing the default logging the logging module, not for this specific class's logger. Can you modify just the logging level of self.logger instead?

anerirana: I did consider this. But, the logging has been implemented throughout the code at the root level. The logging inside `fit()` of `LabelModel` and `log()` of `Logger` are using module level logging function `logging.info()`. Therefore all the logs are generated by root level logger and the only way to set log level for `LabelModel` specifically is to change the `CustomeLogger` significantly. However, I don't think that is necessary. Even if I change the root logger level to `INFO`, it would not interfere with the logging flow. Because, before creating an `INFO` level log, there will always be a check for the verbose parameter.

bhancock8: I see. Alright, this looks good to me then!

bhancock8 (approved):

bhancock8:
Thanks for the PR! We'll take a look into why the Python 3.6 unit tests are getting stuck, as it appears to be unrelated to this PR.

anerirana:
Thanks @bhancock8, please merge the PR as well, because I don't have write access.

bhancock8:
Yep! I'll make sure this gets merged once the unit test build issue is resolved. Thanks @anerirana!

PR #1660, opened 2021-08-20
clean up flake8 ignores

asottile:
- W605 should not be ignored, it will become a syntax error in the future
    - I fixed the offenders automatically with [pyupgrade]
- `extend-ignore` is better than `ignore` as it preserves the default ignore
  list

[pyupgrade]: https://github.com/asottile/pyupgrade

## Test plan

- `python -Wonce -c 'import snorkel.labeling.model.label_model'`
- `python -Wonce -c 'import test.labeling.apply.lf_applier_spark_test_script'`

## Checklist

Need help on these? Just ask!

* [ ] I have updated the documentation accordingly. **N/A**
* [ ] I have run `tox -e complex` and/or `tox -e spark` if appropriate. **N/A**

### DIFF
--- a/setup.cfg
+++ b/setup.cfg
@@ -10,16 +10,12 @@ doctest_optionflags =
     FLOAT_CMP
 
 [flake8]
-ignore =
+extend-ignore =
     E203,
     E265, # Throws errors for '#%%' delimiter in VSCode jupyter notebook syntax
-    E266,
     E501,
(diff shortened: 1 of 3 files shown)


bhancock8:
Thank you for these updates, @asottile! All looks good to me. (The failing CI test is due to a separate known issue that we're investigating).

PR #1661, opened 2021-08-20
fixed lambda LR saving bug

humzaiqbal:
This PR fixes a probelm with the LambaLR scheduler in which label models trained with it can't be serialized properly because it uses a lambda function which pickle doesn't save. I fixed it by changing the scheduler to be class based using this as reference https://stackoverflow.com/questions/52758051/how-to-save-lambdalr-scheduler-in-pytorch-with-lambda-function
## Related issue(s)
https://github.com/snorkel-team/snorkel/issues/1658
Fixes # (issue)

## Test plan
I added a UT where I set a label model using this scheduler and ensured that now it can be saved
## Checklist

Need help on these? Just ask!

*  ✅ I have read the **CONTRIBUTING** document.
*  ✅ I have updated the documentation accordingly.
*  ✅  I have added tests to cover my changes.
*  ✅  I have run `tox -e complex` and/or `tox -e spark` if appropriate.
*  ✅  All new and existing tests passed.

### DIFF
--- a/snorkel/labeling/model/label_model.py
+++ b/snorkel/labeling/model/label_model.py
@@ -17,7 +17,7 @@ from snorkel.labeling.model.graph_utils import get_clique_tree
 from snorkel.labeling.model.logger import Logger
 from snorkel.types import Config
 from snorkel.utils.config_utils import merge_config
-from snorkel.utils.lr_schedulers import LRSchedulerConfig
+from snorkel.utils.lr_schedulers import LRPolicy, LRSchedulerConfig
 from snorkel.utils.optimizers import OptimizerConfig
 
 Metrics = Dict[str, float]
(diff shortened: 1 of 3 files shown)


snorkel/labeling/model/label_model.py
```diff
@@ -675,7 +675,7 @@ def _set_lr_scheduler(self) -> None:
                 total_steps - self.warmup_steps
             )
             lr_scheduler = optim.lr_scheduler.LambdaLR(  # type: ignore
-                self.optimizer, linear_decay_func
+                self.optimizer, lr_lambda=LRPolicy(self.warmup_steps, total_steps)
```
bhancock8, line 678: Is linear_decay_func still being used elsewhere or should it be dropped form that file altogether?

test/labeling/model/test_label_model.py
```diff
@@ -548,6 +549,23 @@ def test_symmetry_breaking_multiclass(self):
...
+        label_model.fit(
+            L_train=L_train,
+            n_epochs=500,
+            lr=0.001,
+            lr_scheduler="linear",
+            optimizer="adam",
+            log_freq=100,
+            seed=123,
+        )
+        label_model.save("./test_save.pickle")
```
bhancock8, line 566: Not a bad idea to include a comment why this test exists (i.e., it's sort of strange to save a model then delete it without reloading or doing anything else with it). Can mention that this model was not previously pickleable due to a lambda function in the schedule (leave breadcrumbs for future readers of the code)

Issue #1654, opened 2021-07-13
test_convergence is flaky when all seed-setting code is removed


melonwater211:
### Introduction
Several tests, including `test_convergence` in `test/classification/test_classifier_convergence.py`, seem to be flaky when all seed-setting code (e.g. `np.random.seed(0)` or `tf.random.set_seed(0)`) is commented out. 

For instance, in commit 1131d9b93798e145b2bc385fa65ff50fbc3c6d0a, `test_convergence` failed ~39% of the time (out of 500 runs) compared to 0% of the time (out of 500 runs) when no seed-setting code is removed.   

*`test_convergence` tests the convergence of the multitask classifier.*

### Motivation
Some tests can be flaky with high failure rates, but are not discovered when the seeds are set. We are trying to stabilize such tests.

### Environment
The tests were run using `pytest 6.2.4` in a `conda` environment with `Python 3.6.13`. The OS used was `Ubuntu 16.04`. 

### Possible Solutions
One possible solution to reduce flakiness is to change the parameters used for prediction. We tried changing the following parameters.

Increasing `lr` from 0.001 to 0.005 and 0.01 reduced the flakiness to ~8% and ~12% respectively.

Increasing `lr` from 0.001 to 0.002 reduced the flakiness to ~2%. 

*Increasing `lr` did not increase runtimes significantly.*

Increasing `n_epochs` from 10 to 15 reduced the flakiness to ~5%.

*This change increased runtimes by a factor of ~1.3.*

Increasing `n_epochs` from 10 to 20 reduced the flakiness to ~0%.

*This change increased runtimes by a factor of ~1.8.*

Another possible solution is to change the values used in the assertions.

Increasing the values used in both assertions (i.e. the assertions on lines 68 and 75) from 0.05 to 0.06 reduced flakiness to ~4%.

*Increasing the values used in the assertions did not increase runtimes significantly.*

Please let us know if these solutions are feasible or if there are any other solutions that should be incorporated. If you are interested, we can send the details of other tests demonstrating similar behavior. We will be happy to raise a Pull Request to fix the tests and incorporate any feedback that you may have.

bhancock8:
[...]

PR #1666, opened 2021-09-05
Fixing test_convergence

melonwater211:
This PR is a followup to Issue #1654 (@bhancock8). We have changed `n_epochs` from 10 to 20. Please let us know if there is anything else that should be done with regards to `test_convergence`. Additionally, there are several other tests that demonstrate similar behavior. Please let us know if we should send the details about those tests. Thank you.

### DIFF
--- a/test/classification/test_classifier_convergence.py
+++ b/test/classification/test_classifier_convergence.py
@@ -26,9 +26,9 @@ class ClassifierConvergenceTest(unittest.TestCase):
     @classmethod
     def setUpClass(cls):
         # Ensure deterministic runs
-        random.seed(123)
-        np.random.seed(123)
-        torch.manual_seed(123)
+        #random.seed(123)
+        #np.random.seed(123)
+        #torch.manual_seed(123)
 
     @pytest.mark.complex
     def test_convergence(self):
@@ -51,7 +51,7 @@ class ClassifierConvergenceTest(unittest.TestCase):
         model = MultitaskClassifier(tasks=[task1, task2])
 
         # Train
-        trainer = Trainer(lr=0.001, n_epochs=10, progress_bar=False)
+        trainer = Trainer(lr=0.001, n_epochs=20, progress_bar=False)
         trainer.fit(model, dataloaders)
         scores = model.score(dataloaders)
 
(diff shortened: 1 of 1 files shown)


rsmith49 (changes requested):
Hi @melonwater211! Thanks for taking the initiative to make the tests more robust.

We actually want to keep tests deterministic for better reproducibility, so please uncomment the lines where the seed is set.

That being said, increasing the number of epochs to 20 sounds great, and we greatly appreciate you taking the time to run those experiments to see which approach reduced non-deterministic flakiness the most. Feel free to increase epochs to reduce flakiness in other tests as well.

melonwater211:
Hi @rsmith49. 

After additional testing, we have determined that the best course of action may be to keep `n_epochs` at 10, increase `lr` to 0022549816312896175, and increase the values of the assertions on lines 68 and 75 to 0.06 and 0.07, respectively. After these changes, `test_convergence` failed 0 times (out of 2000 runs). Of course, if you would prefer to keep the tests deterministic, we will push the stabilization adjustments without removing the seed-setting code. Please let us know if there is anything else that should be done to fix this test.

We would be happy to raise PRs for other tests with similar issues. Thank you.

melonwater211:
Hi @rsmith49.

After running additional tests, we have determined that increasing `n_epochs` to 50 and `lr` to 0.002383698505238875 would better reduce flakiness. Additionally, we have kept the seed-setting code, as requested. Please let us know if anything else should be done to fix this test. We would be happy to raise PRs about other flaky tests. Thank you.

melonwater211:
Hi @rsmith49. I would also like to mention that, with these new changes in place, the probability of failure when the seed-setting code is removed is ~0.08%. Please let us know if anything else should be done or if the PR is ready to merge. Thank you.

rsmith49:
Hi @melonwater211 ! Sorry for the late reply. We recently went through some CI upgrades for the checks to become more streamlined. However, because of this, existing PRs such as this one have the old CI checks still attached. 

Would you mind rebasing to the new master branch and pushing this PR from there? You can accomplish that by

    git pull
    git rebase master
    git push --force

rsmith49:
Closing this as #1715 has addressed the same changes

PR #1671, opened 2021-10-02
Fixing flaky test

crawlingcub:
The test `test_labeling_convergence` in `test/labeling/test_convergence.py` sometimes fails when the seed settings are removed. It failed 52 out of 500 runs times for me on my machine. Updating the assertion threshold to `0.06` ensures that the test always passes (even without seeds). For obtaining this solution, I analyzed the underlying tail distribution of the `err` variable and computed the 99.99th percentile. This provided an estimate of `~0.0555`. I rounded up the value in the assertion to fix the test.

I think the seed settings on lines 61-63 can also be removed since this class only has one test. I can remove them if you agree with this change.

Please let me know if this change looks reasonable. I will be happy to incorporate any other suggestions that you may have.

Thanks!


fpoms:
@crawlingcub thanks for the contribution--and sorry for the delay here. I've confirmed this works as you described am ok to merge if that's ok with you!

fpoms (approved):

fpoms:
@crawlingcub if you could rebase onto main, and the tests pass, then I can merge this.

crawlingcub:
hi i merged master. can you merge?

fpoms:
Merged! Thanks @crawlingcub !

PR #1676, opened 2021-10-14
Change assert_equal to assert_almost_equal

rsmith49:

Missed a coverage test on #1674 , changing some of the array checks to `assert_almost_equal` instead of `assert_equal`.


rsmith49:
This was addressed in https://github.com/snorkel-team/snorkel/pull/1694

PR #1698, opened 2022-04-12
Support m1 mac

zexuan-zhou:
Loosen versions for `numpy`, `blis` to support M1 Mac

## Related issue(s)
https://github.com/snorkel-team/snorkel/issues/1695

Fixes # (issue)
https://github.com/snorkel-team/snorkel/issues/1695

## Test plan
Need help on this

## Checklist

Need help on these? Just ask!
Need help to test py36 and py37 as they are hard to install on M1 Mac.
Need help on `tox -e spark`
Need help on CICD
Need help on what document to update
Need help on whether there should be new test cases

* [X] I have run `tox -e complex` and/or `tox -e spark` if appropriate.
** [X] `tox -e complex` runs and passes all test
** [ ] `tox -e spark` is currently failing with `Exception: Java gateway process exited before sending its port number`
** [X] `tox -e py38` passes
** [X] `tox -e complex` passes
** [X] `tox -e doctest` passes
** [X] `tox -e check` passes
** [X] `tox -e type` passes (all code changes are regarding this)
** [ ] `tox -e py36  # Run unit tests pytest in Python 3.6`
** [ ] `tox -e py37  # Run unit tests pytest in Python 3.7`
** [ ] `tox -e spark  # Run Spark-based tests (marked with @pytest.mark.spark)`
** [ ] `tox -e doc  # Build documentation with Sphinx`

### DIFF
--- a/snorkel/utils/core.py
+++ b/snorkel/utils/core.py
@@ -172,6 +172,7 @@ def filter_labels(
     for label_name, filter_values in filter_dict.items():
         label_array: Optional[np.ndarray] = label_dict.get(label_name)
         if label_array is not None:
+            # _get_mask requires not-null input
             masks.append(_get_mask(label_array, filter_values))
     mask = (np.multiply(*masks) if len(masks) > 1 else masks[0]).squeeze()
 
--- a/snorkel/utils/core.py
+++ b/snorkel/utils/core.py
@@ -170,8 +170,9 @@ def filter_labels(
     """
     masks = []
     for label_name, filter_values in filter_dict.items():
-        if label_dict[label_name] is not None:
-            masks.append(_get_mask(label_dict[label_name], filter_values))
+        label_array: Optional[np.ndarray] = label_dict.get(label_name)
+        if label_array is not None:
+            masks.append(_get_mask(label_array, filter_values))
     mask = (np.multiply(*masks) if len(masks) > 1 else masks[0]).squeeze()
 
     filtered = {}
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@
 #### ESSENTIAL LIBRARIES
 
 # General scientific computing
-numpy>=1.16.5,<1.20.0
+numpy>=1.16.5,<1.22.0
 scipy>=1.2.0,<2.0.0
 
 # Data storage and function application
--- a/requirements.txt
+++ b/requirements.txt
@@ -9,7 +9,7 @@
 #### ESSENTIAL LIBRARIES
 
 # General scientific computing
-numpy>=1.16.5,<1.22.0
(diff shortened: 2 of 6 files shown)


zexuan-zhou:
@rsmith49 @bhancock8 Please take a look. I need some help on the tests for py36, py37, and spark, and CICD. Thank you

snorkel/utils/core.py
```diff
@@ -170,8 +170,10 @@ def filter_labels(
     """
     masks = []
     for label_name, filter_values in filter_dict.items():
-        if label_dict[label_name] is not None:
-            masks.append(_get_mask(label_dict[label_name], filter_values))
+        label_array: Optional[np.ndarray] = label_dict.get(label_name)
+        if label_array is not None:
+            # _get_mask requires not-null input
+            masks.append(_get_mask(label_array, filter_values))
```
zexuan-zhou, line 176: Mypy doesn't seem to be smartly enough to tell that `label_dict[label_name]` is not-null after the condition check `if label_dict[label_name] is not None`. It still treats `label_dict[label_name]` as `Optional[np.ndarray]` and pass it to `_get_mask` which has requirement on not-null input `def _get_mask(label_array: np.ndarray, filter_values: List[int])`. This causes mypy to error out.
By extracting `label_dict[label_name]` into another variable `label_array` the issue will go away.

rsmith49: Does
```
assert label_dict[label_name] is not None
```
fix the mypy check? If not I think this should be fine (cc: @bhancock8 )

rsmith49:
@zexuan-zhou thanks for getting going here! Can you re-create this PR to merge into `snorkel-team:update-numpy-version` instead of `snorkel-team:master`? That way we can run the checks in CI after merging in these changes before merging into master

snorkel/utils/core.py
zexuan-zhou: @rsmith49 Tried that. It doesn't fix mypy error. Also not sure if it's a good idea to add run time check here to prevent the program from running if `label_dict[label_name]` is None

zexuan-zhou:
> @zexuan-zhou thanks for getting going here! Can you re-create this PR to merge into `snorkel-team:update-numpy-version` instead of `snorkel-team:master`? That way we can run the checks in CI after merging in these changes before merging into master

@rsmith49 Done. Changed the base to `snorkel-team:update-numpy-version`

snorkel/utils/core.py
rsmith49: Agreed, I meant after the if statement (sorry that wasn't clear). If mypy is having issues with dict values though, this change should be fine

rsmith49 (approved):

PR #1699, opened 2022-04-12
Ignore mypy error when overwriting `label_dict` with new type

zexuan-zhou:
See code


zexuan-zhou:
@rsmith49 I think this is a better solution to deal with the mypy error on overwriting the type of invariant variable

PR #1710, opened 2022-07-29
Update release date for 0.9.9 and release instructions to be more consistent

humzaiqbal:
This PR updates the release instructions to only add the release version when we are about to cut a release not right after the previous one was cut
## Related issue(s)

Fixes # (issue)

## Checklist

Need help on these? Just ask!

* [ ] I have read the **CONTRIBUTING** document.

### DIFF
--- a/RELEASING.md
+++ b/RELEASING.md
@@ -28,6 +28,7 @@ Then run `chmod 600 ./.pypirc` so only you can read/write.
        $ git pull origin main
 
 1. Add a new changelog entry for the release.
+
        ## [0.9.0]
        ### [Breaking Changes]
        ### [Added]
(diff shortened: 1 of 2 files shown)


RELEASING.md
```diff
@@ -76,16 +76,6 @@ Then run `chmod 600 ./.pypirc` so only you can read/write.
 
 1. Update version to, e.g. 0.9.1+dev in `snorkel/version.py`.
 
-1. Add a new changelog entry for the unreleased version:
```
bhancock8, line 79: We do actually still want to create these sections under `[Unreleased]` so that there's a place to being storing these different types of changes as they're merged into the tip of main. We just don't want to have both `[Unreleased]` _and_ the name of the next version.

humzaiqbal: Ok added back

bhancock8 (approved):

